PR #7, opened 2019-04-17
CredSSP+NTLM finalization

vaffeine:
Major changes:
- Changed the way we retrieve user name. We get it from NTLM and not from cookie now.
- Added package capture for RDP
- Added Windows RDP Client support
- Fixed server error not being sent
- Added integration tests for CredSSP+NTLM


rdp-proto/src/ntlm/messages/server/authenticate.rs
```diff
@@ -150,15 +138,12 @@ where
     Ok(mic)
 }
 
-fn get_authenticate_message_fields(
+fn process_message_fiels(
```
ekse: typo: rename to `process_message_fields`

rdp-proto/src/tpdu.rs
```diff
@@ -23,6 +24,39 @@ pub enum X224TPDUType {
...
+            io::ErrorKind::UnexpectedEof,
+            "The buffer length is less then the real length",
+        ));
+    }
+
+    let (_, code) = parse_tdpu_header(&mut stream)?;
+
+    let mut tpdu = input.split_to(len as usize);
+    match code {
+        X224TPDUType::Data => tpdu.advance(TPDU_DATA_LENGTH),
```
ekse: `advance` panics if there is not enough data remaining. I didn't see a safe advance function in the library, I think we need to manually check the remaining length before calling advance.

https://docs.rs/bytes/0.4.12/bytes/struct.BytesMut.html#method.advance

rdp-proto/src/ntlm/messages/server/authenticate.rs
vaffeine: Thanks. Done

rdp-proto/src/tpdu.rs
vaffeine: Good catch, fixed it

vaffeine:
> The Remote Desktop client supports connecting without specifying a username, this is not supported by the proxy; it prints "RDP failed: invalid or unsuppored cookie message" and the connection is closed.
Remote Desktop client doesn't send cookie in the negotiation message if user doesn't specify username. I fixed it now. Thanks

> When enabling capture to pcap, only client side packets are captured.
Sorry, I was testing pcap files only with freerdp client and only briefly looked into the case with Windows client. For some reason Windows server responds with 4 packages that contain only zeroes right after the last CredSSP message. And Inceptor was not ready for that. So we kept draining 0 bytes from the package buffer. Now we drain the amount of bytes we read. Do you or @awakecoding know the possible reason of such behavior? It doesn't look like the end of the previous CredSSP message and only happens with Windows client but not with freerdp.

> Also, the --help message for the -f options says "Only WaykNow protocol can be saved." but it seems to be functional.
Fixed.

awakecoding:
@vaffeine about CredSSP, do you mean that the server sends a single 4-byte message filled with zeroes after the last CredSSP message, or at the end of the last CredSSP message? I suspect it is either a new field or padding that was added in recent versions of CredSSP to prevent a padding oracle from older versions of the protocol. I'll try to find out what it is exactly

awakecoding:
@vaffeine I just checked it is possible that those 4 bytes are from the "clientNonce" optional field added in version 5 of CredSSP: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-cssp/6aac4dea-08ef-47a6-8747-22ea7f6d8685

Otherwise, it could also be the optional errorCode field used in versions 3, 4 and 6 of CredSSP. Since "clientNonce" is said to be a nonce (random number), a value of zero wouldn't make much sense. However, a value of zero can make sense when interpreted as an errorCode, where zero means success.

Both fields are marked as OPTIONAL in ASN.1, so their presence is encoded somewhere in the message, I just don't recall exactly how you can check for presence (my ASN.1 is rusty).

I suspect mstsc and FreeRDP must be using different versions of CredSSP by default, hence the difference in the response from the server.

vaffeine:
@awakecoding Server sends single package that consists of 4 zeroes right after we send user credentials (step 9 in https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-cssp/94846575-5a58-44de-b07b-48b90af328fb). According to the specification there should not be any response from the server, but it indeed looks like an error code. It is not encoded in ASN.1, but Microsoft might have added it to display better error message on the client.

PR #10, opened 2019-04-30
wayk-1539 - Add http server

fdubois1:


src/http/controllers/health.rs
```diff
@@ -0,0 +1,36 @@
...
+        self.dispatch.dispatch(req, res);
+    }
+
+    fn base_path(&self) -> &str {
+        "/health"
+    }
+}
+
+fn health(_: &ControllerData, _req: &SyncRequest, res: &mut SyncResponse) {
+    res.status(StatusCode::OK).body("I'm here and I'm alive, that's enough");
```
richerarc: Maybe a check on what's going on inside the server could be insightful here (e.g. Check database connection)

fdubois1: Yeah, I looked what I could do, but we don't have a database and the server is only a listener on 8080 to create a proxy between 2 wayk.  If the program is running, it means that we are still listening on 8080 and the forward future are polled... So I didn't see anything that could be useful to check. I added that route just to have it in future if we would like to check a database connection for example, as you said.

richerarc: Yeah, but the fact that the program is still running doesn't ensure that your still listening on 8080

fdubois1: If something happen, I guess an error will be raised and futures will finish.  Or is there a way to know that we are still listening and processing request ?

PR #14, opened 2019-07-10
RDP Connection Sequece (Part 1)

vaffeine:
- Implemented RDP connection sequence up to capability sets exchange
- Ran cargo fix, cargo clippy and cargo fmt on code in Jet so Travis pipeline passes
- Added rustdoc for the previous iteration of RDP


ekse:
The file rdp-proto/src/rdp/tests.rs is not referenced anymore, the tests fail to compile if I add them.

vaffeine:
Oops, I divided the old rdp.rs file into several and did the same with the test. But apparently I forgot to remove this file after that. I will remove it in the next PR.

PR #15, opened 2019-07-10
rdp-proto is a library, remove the Cargo.lock

ekse:


vaffeine:
I read different opinions about the Cargo.lock file inside the library sub-crate, which is inside the binry crate. The general idea having it checked-in is that sub-crates of the binary crate should also have fixed dependencies because the crate is, overall, binary.
But I have seen different big project doing otherwise too, so I think it doesn't really matter.

ekse:
That's an interesting point. Isn't the project Cargo.lock keeping track of the same information? As long as rdp-proto is used only inside this project it probably doesn't really matter.

vaffeine:
Yes, the project's Cargo.lock keeps track of it. But if tests run separately for the crate and sub-crate it in theory can violate project's Cargo.lock. But this shouldn't happen in a good crates.
So it really really doesn't. I think we can remove it because we don't have a Cargo.lock in jet-proto.

PR #19, opened 2019-07-15
Fixed devolutions-jet so that it runs on macOS

LucFauvel:
- Added a password to the certificate so that macOS's keychain can import it natively (didn't work without a password because of keychain)
- Set jet to use openssl for TLS on macOS


LucFauvel:
> Can you make it unconditional instead of adding macOS? This should be compiled for Windows as well, we can handle the OpenSSL dependency

Done

vaffeine:
Hi

`cargo test` fails for me with:
```
2019-07-18T13:28:35Zthread ' <unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Normal(ErrorStack([Error { code: 587686001, library: "PKCS12 routines", function: "PKCS12_parse", reason: "mac verify failure", file: "../crypto/pkcs12/p12_kiss.c", line: 66 }]))', src/libcore/result.rsINFO: 997:5received negotiation response from server (protocol: HYBRID, flags: DYNVC_GFX_PROTOCOL_SUPPORTED | RDP_NEG_RSP_RESERVED | RESTRICTED_ADMIN_MODE_SUPPORTED | REDIRECTED_AUTHENTICATION_MODE_SUPPORTED)
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
, client: 127.0.0.1:47164
2019-test rdp_with_nla_ntlm ... FAILED
```

Can you please take a look and fix them?

P.S. It would be great to add `cargo test` to azure pipelines

PR #23, opened 2019-07-29
Added websocket support

LucFauvel:


src/websocket_client.rs
```diff
@@ -0,0 +1,171 @@
...
+        .get(header::CONNECTION)
+        .map_or(false, |v| connection_has(v, "upgrade"));
+    let is_websocket_upgrade = req.headers()
+        .get(header::UPGRADE)
+        .and_then(|v| v.to_str().ok())
+        .map_or(false, |v| v.eq_ignore_ascii_case("websocket"));
+    let is_websocket_version_13 = req.headers()
+        .get(header::SEC_WEBSOCKET_VERSION)
+        .and_then(|v| v.to_str().ok())
+        .map_or(false, |v| v == "13");
```
awakecoding: checking for websocket version 13 explicitly is probably too restrictive. maybe we should just check for the websocket version header presence instead, unless we have a meaningful check to perform with regards to the version number.

src/websocket_client.rs
```diff
@@ -0,0 +1,171 @@
...
+        .header(header::UPGRADE, "websocket")
+        .header(header::CONNECTION, "upgrade")
+        .header(header::SEC_WEBSOCKET_ACCEPT, key.as_str())
+        .header(header::SEC_WEBSOCKET_PROTOCOL, "binary")
+        .body(Body::empty())
+        .unwrap()
+}
+
+
+fn uuid_from_path(path: &str) -> Option<Uuid> {
```
awakecoding: the way the URL is parsed and then deconstructed appears a bit flaky to me, maybe use regular expressions to catch specific patterns for /jet/create/<uuid>, /jet/connect/<uuid>, /jet/accept/uuid? this can be cleaned up later though.

src/transport/ws.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+}
+
+impl Stream for WsJetStream {
+    type Item = Vec<u8>;
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Result<Async<Option<<Self as Stream>::Item>>, <Self as Stream>::Error> {
+        if let Ok(ref mut stream) = self.stream.try_lock() {
+            let mut buffer = [0u8; 65535];
```
awakecoding: is this the best way to handle receiving bytes? it creates a 65535-byte buffer, sets it to zero, then a bunch of bytes are read into it, and the result is converted to a vector.

ekse: I would recommend at least using a Vec as static arrays are allocated on the stack.

LucFauvel: This was copy pasted from the src/transport/tcp.rs file, if this is to be changed, it should be changed there as well.

awakecoding: Oops :) I'll ask François about it then

src/websocket_client.rs
awakecoding: For sample usage of regex to parse URLs, you can take a look here: https://github.com/Devolutions/qui-vive-rs/blob/master/src/service.rs#L85

src/websocket_client.rs
LucFauvel: This version check can be removed entirely if this is a cause for concern

richerarc: V13 was standardized in December 2011 and is supported from IE 10 and above and all other chromium base / Mozilla browser support it, the check could instead be replaced by a check for a version greater or equal to v13.

src/transport/ws.rs
richerarc: The match on stream.poll_read(&mut buffer) could be done in a loop where you would read small amount until nothing is left to read / or until max tcp frame is reached (e.g. 65535). The fact that each time poll is called a 65KB array / vec is allocated and then truncated to a potentially way-shorter one is also something that would be fixed by a loop.

src/websocket_client.rs
awakecoding: I agree with this suggestion. Let's parse and check for version >= 13.

src/transport/tcp.rs
```diff
@@ -199,40 +203,52 @@ impl Stream for TcpJetStream {
...
-                Ok(Async::Ready(len)) => {
-                    let mut v = buffer.to_vec();
-                    v.truncate(len);
-                    self.nb_bytes_read += len as u64;
-                    debug!("{} bytes read on {}", len, stream.peer_addr().unwrap());
-
-                    if let Some(interceptor) = self.packet_interceptor.as_mut() {
-                        interceptor.on_new_packet(stream.peer_addr().ok(), &v);
+            let mut result = Vec::new();
+            while result.len() <= TCP_READ_LEN {
```
fdubois1: I can imagine a few edge cases :

- What happen if we read exactly 8192 bytes ? We will read 8192 bytes and we will try to read again. The next call will probably return NotReady I guess... and we will lose those 8192 bytes.
- Is it possible to read data and get Ready(0) the call after ? I guess it could be possible if the connection is closed but we haven't read everything yet. We will read first then we will get the Ready(0) and lose data.
- If we reach the max (57343) and we send back the message but more data is still ready to be read, are we sure that we will be called to read again or we will be called only when another event will occur on the stream ?

richerarc: > If we reach the max (57343) and we send back the message but more data is still ready to be read, are we sure that we will be called to read again or we will be called only when another event will occur on the stream ?
Yes, the rule with streams is simple:
Assuming a Struct A (Stream) that poll on another struct B (Stream)
1. - if B is still ready and A return Ready -> A::poll will be called again
2. - if B is not ready and A return Ready -> A::poll will be called again and then will return not ready
3. - if B is not ready and A return not Ready -> A::poll will be called whenever B is ready
4. - if B is ready and A return not Ready -> A::poll will never be called again unless the task is woken up

You are talking about case 1.

PR #56, opened 2019-11-20
Logs optimization

Fotosmile:
The first batch of changes for Jet optimization:
1. Proxy moved out to the separate module;
2. Added LevelFilter to the AsyncDrain, so now Jet does not rain "channel overflow" messages to logs. The performance was improved by up to 25%;
3. Replaced debug logs in RDP connection sequence and in the "hot" places (such as TcpJetStream::poll, TcpJetSink::start_send) with trace logs, removed trace logs from release builds. Thus, a release build will not spend time to canceled trace logs, which can be viewed at any time in debug builds :) The performance was improved by up to 15%. @fdubois1 are you agree with these changes?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ serde_json = "1.0"
 slog = { version = "2.5", features = ["max_level_trace","release_max_level_trace"] }
 slog-term = "2.4"
 slog-async = "2.3"
-slog-scope = "4.1"
+slog-scope = "4.3"
 slog-scope-futures = { git = "https://gitlab.com/Fotosmile/slog-scope-futures.git" }
 chrono = { version = "0.4", features = ["serde"] }
 failure = "0.1"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -29,10 +29,10 @@ saphir = { version = "0.9.5", features = ["https", "request_handler"] }
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
-slog = { version = "2.5", features = ["max_level_trace","release_max_level_trace"] }
+slog = { version = "2.5", features = ["max_level_trace","release_max_level_debug"] }
```
fdubois1: Does it change something on the performance ? Or on the executable size ?  I don't think we have already logged with trace level in production, usually, debug should be enough, but what do we win to change it to debug ? Thanks

fdubois1: Oh, I had not seen your message in the PR.  Ok, if performance is better, I'm OK with that.

PR #60, opened 2019-12-04
Jet optimization

Fotosmile:
1. Replaced Arc Mutex with tokio::split for JetTransport;
2. Integrated BipBuffer for JetStream/JetSink;
3. Did other small optimization.


src/jet/candidate.rs
```diff
@@ -17,32 +25,36 @@ pub struct CandidateResponse {
...
+            association_id: c.association_id.clone(),
+            transport_type: c.transport_type.clone(),
             bytes_sent,
             bytes_recv,
         }
     }
 }
-#[derive(Clone)]
+
 pub struct Candidate {
```
Fotosmile: Because a candidate is a server, and there can be many of them until a client selects a specific server. After the client will do it, the transport is no longer needed (only nb_bytes_read / nb_bytes_written).
Therefore, I left here only the transport field, in which the server transport will be. After the client selects a candidate, the transport will be taken from the candidate, and corresponding copies of the client will be set in client_nb_bytes_read / written.

src/transport/mod.rs
```diff
@@ -1,30 +1,45 @@
...
+const PART_LEN: usize = 16 * 1024;
+
 pub trait Transport {
     fn connect(addr: &Url) -> JetFuture<Self>
     where
         Self: Sized;
-    fn message_sink(&self) -> JetSinkType<Vec<u8>>;
-    fn message_stream(&self) -> JetStreamType<Vec<u8>>;
+    fn peer_addr(&self) -> Option<SocketAddr>;
+    fn split_transport(
```
Fotosmile: Because now there is no Arc Mutex, we will do tokio::split for stream and sink. After that, in theory, the transport is no longer needed - therefore, we accept self.

src/jet_client.rs
```diff
@@ -144,224 +151,331 @@ impl Future for JetMsgReader {
...
+                    (StatusCode::OK, uuid)
+                }
+                2 => {
+                    let mut status_code = StatusCode::BAD_REQUEST;
+
+                    if let Some(association) = jet_associations.get_mut(&request.association) {
+                        if association.version() == JET_VERSION_V2 {
+                            if let Some(candidate) = association.get_candidate_mut(request.candidate) {
+                                if candidate.transport_type() == TransportType::Tcp {
+                                    candidate.set_state(CandidateState::Accepted);
```
Fotosmile: Here we do not set transport, because using the transport, we still need to respond to the JetAcceptResponse server. We will set it when we send the response to the server.

src/jet_client.rs
```diff
@@ -144,224 +151,331 @@ impl Future for JetMsgReader {
...
-            } else {
-                return Ok(Async::NotReady);
+                _ => {
+                    // No jet message exist with version different than 1 or 2
+                    unreachable!()
+                }
+            };
+
+            if request.version == 1 && status_code == StatusCode::OK {
+                self.remove_association_future = Some(Box::new(create_remove_association_future(
```
Fotosmile: We also create RemoveAssociationFuture in advance, while the association is available.

src/jet_client.rs
```diff
@@ -144,224 +151,331 @@ impl Future for JetMsgReader {
...
+            // Build response
+            let response_msg = JetMessage::JetAcceptRsp(JetAcceptRsp {
+                status_code,
+                version: request.version,
+                association,
+                instance: self.config.jet_instance().clone(),
+                timeout: ACCEPT_REQUEST_TIMEOUT_SEC,
+            });
+            let mut response_msg_buffer = Vec::with_capacity(512);
+            response_msg.write_to(&mut response_msg_buffer)?;
```
Fotosmile: Only once we serialize the message

src/jet_client.rs
```diff
@@ -144,224 +151,331 @@ impl Future for JetMsgReader {
...
-        let mut v = Vec::new();
-        response_msg.write_to(&mut v)?;
-        try_ready!(self.transport.poll_write(&v));
-
-        // Start timeout to remove the association if no connect is received. We start it only if a new association has just been added,
-        // possible only with version 1.
-        if let JetMessage::JetAcceptRsp(accept_rsp) = response_msg {
-            if accept_rsp.version == 1 && accept_rsp.status_code == StatusCode::OK {
-                start_remove_association_future(self.executor_handle.clone(), self.jet_associations.clone(), accept_rsp.association);
+    fn handle_set_transport(&mut self) -> Poll<(), io::Error> {
```
Fotosmile: After we have written the response, depending on the version, we set up the transport

src/jet_client.rs
```diff
@@ -144,224 +151,331 @@ impl Future for JetMsgReader {
...
+                    }
+                }
+                _ => {
+                    // No jet message exist with version different than 1 or 2
+                    unreachable!()
+                }
+            };
+
+            if let Some(remove_association_future) = self.remove_association_future.take() {
+                self.executor_handle.spawn(remove_association_future);
```
Fotosmile: And if everything is ok, then do spawn for remove_association_future (as before)

src/jet_client.rs
```diff
@@ -144,224 +151,331 @@ impl Future for JetMsgReader {
...
                         }
-                    }
+                    };
 
+                    if let Some(candidate) = candidate {
+                        // The accept request has been received before and a transport is available to open the proxy
+                        if let Some(server_transport) = candidate.take_transport() {
+                            candidate.set_state(CandidateState::Connected);
+
+                            self.server_transport = Some(server_transport);
```
Fotosmile: We get the transport from the candidate (he is our target server), temporarily save, and also save the association_id, candidate_id, which will be needed for the result.

src/websocket_client.rs
```diff
@@ -22,102 +23,167 @@ pub struct WebsocketService {
...
+
+                            let fut = req
+                                .into_body()
+                                .on_upgrade()
+                                .map(move |upgraded| {
+                                    if let Ok(mut jet_assc) = jet_associations_clone.lock() {
+                                        if let Some(assc) = jet_assc.get_mut(&association_id) {
+                                            if let Some(candidate) = assc.get_candidate_mut(candidate_id) {
+                                                candidate.set_state(CandidateState::Accepted);
+                                                candidate.set_transport(JetTransport::Ws(WsTransport::new_http(
```
Fotosmile: set transport for the candidate

src/websocket_client.rs
```diff
@@ -22,102 +23,167 @@ pub struct WebsocketService {
...
+                                .into_body()
+                                .on_upgrade()
+                                .map(move |upgraded| {
+                                    if let Ok(mut jet_assc) = jet_associations_clone.lock() {
+                                        if let Some(assc) = jet_assc.get_mut(&association_id) {
+                                            if let Some(candidate) = assc.get_candidate_mut(candidate_id) {
+                                                if (candidate.transport_type() == TransportType::Ws || candidate.transport_type() == TransportType::Wss)
+                                                    && candidate.state() == CandidateState::Accepted {
+                                                    {
+                                                        let server_transport = candidate
```
Fotosmile: Here the same thing happens as in jet_client.rs: take transport from the candidate, set the client's nb_bytes_read/written, and create the proxy.

src/transport/mod.rs
```diff
@@ -120,13 +131,179 @@ impl AsyncWrite for JetTransport {
...
     fn nb_bytes_read(&self) -> u64;
     fn set_packet_interceptor(&mut self, interceptor: Box<dyn PacketInterceptor>);
 }
 
 pub trait JetSink: Sink {
-    fn shutdown(&mut self) -> std::io::Result<()>;
     fn nb_bytes_written(&self) -> u64;
 }
+
+struct JetStreamImpl<T: AsyncRead> {
```
Fotosmile: Moved out the same JetStream / JetSink implementations from tcp and ws

src/transport/mod.rs
```diff
@@ -120,13 +131,179 @@ impl AsyncWrite for JetTransport {
...
+
+impl<T: AsyncRead> Stream for JetStreamImpl<T> {
+    type Item = usize;
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        let peer_addr = &self.peer_addr_str;
+        let mut written = 0;
+        loop {
+            if let Some(mut reservation) = self.buffer.reserve(PART_LEN) {
```
Fotosmile: BipBuffer usage: try to reserve the buffer. If it's Some, then we will read to the buffer, truncate and send for the Sink

src/proxy.rs
```diff
@@ -24,21 +29,20 @@ impl Proxy {
         server_transport: T,
         client_transport: U,
     ) -> Box<dyn Future<Item = (), Error = io::Error> + Send> {
-        let jet_sink_server = server_transport.message_sink();
-        let mut jet_stream_server = server_transport.message_stream();
+        let (client_writer, server_reader) = bip_buffer_with_len(BIP_BUFFER_LEN);
```
Fotosmile: Create bip buffer's reader and writer for the server and client, and pass them to the split_transport methods.

src/proxy.rs
```diff
@@ -76,43 +80,45 @@ impl Proxy {
...
-                    client = &client_addr
-                );
-
-                Ok(())
-            }).then(|result| {
-            SESSION_IN_PROGRESS_COUNT.fetch_sub(1, Ordering::Relaxed);
-            result
-        }))
+        Box::new(
+            f1.select2(f2)
```
fdubois1: I remember that I have hesitated between a select and a join.  If we call a select, could it be possible that we don't forward some packets ?  Let say we have these events :
- Client sent a terminate packet to the server
- Client close his connection
- Server tries to send data to client --> it fails
Result : We stop everything... But we didn't forward the terminate packet from the client to the server. And it is often there that we have the reason of the disconnection...  

I understand that with a join, the first thing that was done is a shutdown on the stream so we could lost it as well...

Is it something that we should pay attention or it should be ok ?

Fotosmile: I also hesitated about this case, but actually there are no problems thanks to Forward future :)
this case will behave like this:
1. The client sends a terminate packet to Jet;
2. The client closes the connection;
3. Jet reads terminate packet via Stream Future and sends the bytes to the Sink Future. And then, Forward Future calls Sink Future until it returns AsyncSink::Ready (so, all terminate packet is redirected);
4. Stream Future is polled again, but underlying stream returns 0 bytes read, so we finish the Stream Future. Forward Future closes Sink Future on the Stream Future's Async::Ready(None).

Result: Jet forwards the terminate packet from the client to the server, and then finishes this Future. After that Select Future works and stops another one Forward Future (and this works exactly like join + shutdown).
So, it should be ok :)

PR #68, opened 2020-03-18
Fix env var name prefixed with PICKY instead of JET

bcortier-devolutions:


awakecoding:
My bad, I should have actually looked at the changes XD

PR #73, opened 2020-04-16
Move the code in a library

ekse:
This is to deal with a limitation of cargo-fuzz, fuzzers can only use code from librairies. The workaround is to move the code to lib.rs and import it in main.rs. I'm currently working on fuzzers, they will be added in another pull request.

Tests are passing with the exception of `rdp_with_nla_ntlm` which is also failing on the master branch.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,27 @@
+#[macro_use]
+extern crate serde_json;
+#[macro_use]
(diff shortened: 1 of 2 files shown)


awakecoding:
@fdubois1 je pense pas que ca change grand chose?

PR #91, opened 2020-08-31
CI test script runnability fixes

pacmancoder:
This PR fixed multiple Clippy warnings which are causing test.sh CI script to fail.

Also this PR sets rdp_with_nla_ntlm integration test as #[ignored]; 
This test fails because it requires that the test environment should have an ironrdp_client application in PATH, which is not usually the case. Maybe separate refactoring for this test is required to eliminate the additional environment dependencies. 

Note that out Travis CI is still looks broken and should be fixed, builds are not triggered automatically, and last builds (~9-10 months ago) were failed (mostly because of the problems which were solved in this PR).

### DIFF
--- a/tests/rdp.rs
+++ b/tests/rdp.rs
@@ -80,7 +80,10 @@ fn run_client() -> Child {
     client_command.spawn().unwrap_or_else(|e| panic!("{:?}", e))
 }
 
+// NOTE: The following test is disabled by default as it requires specific environment with
+// ironrdp_client executable in PATH variable
 #[test]
+#[ignore]
 fn rdp_with_nla_ntlm() {
     let mut identities_file = tempfile::NamedTempFile::new().expect("failed to create a named temporary file");
     let rdp_identities = vec![RdpIdentity::new(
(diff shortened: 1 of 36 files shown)


tests/rdp.rs
```diff
@@ -77,10 +77,13 @@ fn run_client() -> Child {
         client_command.args(&["--domain", domain]);
     }
 
-    client_command.spawn().expect("failed to run IronRDP client")
+    client_command.spawn().unwrap_or_else(|e| panic!("{:?}", e))
```
bcortier-devolutions, line 80: Does this correspond to some clippy warning? `expect` (like it was previously) or `unwrap` (equivalent to the current code but less contrived) should be good enough.
```suggestion
    client_command.spawn().unwrap()
```

pacmancoder: Thank you, this change is redundant indeed, I am accidentally committed it. I will return `expect` back.

bcortier-devolutions (approved):
Thank you! LGTM.

PR #92, opened 2020-08-31
JWT-based routing routing feature

pacmancoder:
This feature implements the feature described [here ](https://devolutions.atlassian.net/browse/WAYK-1917)(Internal ticket).

Basically, this feature introduces simple connection redirection (without traffic inspection) based on the JWT token contents. Destination host is provided to the Jet using dst_hst claim. JWT token itself can be injected into the Preconnection PDU blob using "pcb" option for MSTSC/FreeRDP.

When launching Jet proxy, --provisioner-public-key-file should be specified to provide PEM file with the public key for JWT token validation. If it is not specified - JWT token validation is skipped.

Also, it was required to bump The IronRDP version, so a small number of changes were introduced to fix compilation issues.

### Use JWT token for RDP connection in MSTSC
1. Open MSTSC
2. Enter a JET address in the "computer" field
3. Press the "Save As..." button under the "Connection settings" panel to save ".RDP" file to you PC
4. Open saved ".RDP" file with the text editor
5. Append string "pcb:s:<JWT token>"  to the end of the file (e.g: pcb:s:eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOj....)
6. Save file
7. In MSTSC press "Open..." and select your edited file
8. Done. You can start the connection

### Use JWT token for RDP connection in FreeRdp
For FreeRDP token can be provided using "/pcb" CLI parameter

### JWT token generation for testing
To simplify the testing process of the feature, a small JWT token generation utility was created "jwt-routing-gen"
Utility usage example:
```
jwt-routing-gen.exe --destination=192.168.1.1 --signing-key=D:\keys\private.pem --validity-period=3m`
```
- This utility will print the generated JWT token to stdout.
- Signing key pair for testing can be generated using https://travistidwell.com/jsencrypt/demo
- jwt-routing-gen sources and Win64 binary are attached to the current PR
[jwt-routing-gen-src.zip](https://github.com/Devolutions/devolutions-jet/files/5150355/jwt-routing-gen-src.zip)
[jwt-routing-gen-win64.zip](https://github.com/Devolutions/devolutions-jet/files/5150360/jwt-routing-gen-win64.zip)

This ticket requires https://github.com/Devolutions/IronRDP/pull/29 to be merged to correctly handle Preconnection PDU parsing.
Also, it would be great to merge https://github.com/Devolutions/devolutions-jet/pull/91.

After merging PR's above, IronRDP's version change in Cargo.toml is needed, and cargo clippy/fix/fmt should be run again.


Cargo.toml
```diff
@@ -42,7 +42,7 @@ chrono = { version = "0.4", features = ["serde"] }
 failure = "0.1"
 x509-parser = "0.6.0"
 sspi = "0.3.0"
-ironrdp = "0.4"
+ironrdp = { git = "https://github.com/Devolutions/IronRDP", branch = "preconnection-pdu-improvements" }
```
pacmancoder, line 45: Temporary change before preconnection-pdu-improvements IronRDP branch is berged to the master and pushed to the crates.io

awakecoding:
@pacmancoder @bcortier-devolutions and I are having trouble figuring out how we should be launching the devolutions-jet relay with a proper RDP listener for TCP-based relaying. Do you have such a command-line sample, or more information how to properly set it up? I'm not even sure which listener should actually be used, and if the RDP TCP-based relayed has been mixed alongside the previous RDP TLS relaying code. We need an identities.json file for the RDP TLS relaying, but we don't need one for the TCP-based relaying. I'll let you know if we figure it out today, otherwise we'll wait for further instructions from you.

pacmancoder:
@awakecoding The Jet relay start command is the same as before, except we need add an aditional `--provisioner-public-key-file` option to specify path to public pem key file for JWT signature validation:
```
devolutions-jet --jet-instance 127.0.0.1 --routing-url rdp://anything.com --listener tcp://0.0.0.0:3389 --protocol "rdp"  --identities-file D:/identities.json --provisioner-public-key-file D:/keys/public.pem
```
--jet-instance can be anything, --routing-url is just required to start with "rdp://" and does not make any effect for now. Identities file is still mandatory because it is used during "normal" Jet operation, even it is not used for the actual JWT routing. So, you can just place some stub data inside identities.json file.

awakecoding (approved):
LGTM

PR #97, opened 2020-10-09
Initial system service support

awakecoding:
This is part of the work required to make the new Devolutions Gateway package for Devolutions Server. I added the following:

- support for launching as a system service using ceviche-rs
- support for JSON-based configuration file support
- try loading configuration files from global path first
- fallback to loading from CLI and environment variables
- toggle between "console" and "service" mode with --service

It's not very clean, but it shouldn't affect our current usage inside Windows containers. I also added a PowerShell script with helper functions to deal with the key pair generation, token generation, and Windows service registration/unregistration.


src/config.rs
```diff
@@ -65,15 +95,180 @@ pub struct Config {
...
+
+fn get_default_hostname() -> Option<String> {
+    hostname::get().ok()?.into_string().ok()
+}
+
 impl Config {
+    pub fn load() -> Option<Self> {
+        let args: Vec<String> = env::args().collect();
+
+        let service_mode = args.contains(&"--service".to_string());
```
bcortier-devolutions, line 170: Ça devrait suffire
```suggestion
        let service_mode = env::args().any(|arg| arg == "--service");
```

bcortier-devolutions (approved):
Nice!

PR #99, opened 2020-10-16
WAYK-2198: allow hostnames in addition to IPs

bcortier-devolutions:


awakecoding (approved):
LGTM

awakecoding:
I think you need to run cargo format to fix the build

PR #101, opened 2020-10-19
Jet bugfix

pacmancoder:
Thi PR fixes the following:
- Fixed random connection hang when using mstsc client (was caused by incorrect AcceptConnection  logic, which caused tokio to never call poll no the future  again)
- Fixed missing support for non-RDP8 connections for (In FreeRDP connections without /gfx nor /rfx and connections only with /rfx)

The second part of the bugfix scope is in the following IronRDP PR: https://github.com/Devolutions/IronRDP/pull/32 
NOTE: IronRDP PR should be merged first and IronRDP crate reference in  Cargo.toml should be updated

Fixes: WAYK-2182, WAYK-2190

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 target/
 .DS_Store
 .idea/
+.vscode/
+# we don't want Cargo.lock for jet-proto (not a binary), also TODO: maybe we should use a workspace?
+jet-proto/Cargo.lock
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1228,7 +1228,7 @@ dependencies = [
 [[package]]
 name = "ironrdp"
 version = "0.4.1"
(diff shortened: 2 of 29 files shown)


awakecoding:
@pacmancoder I have merged the corresponding pull request in IronRDP

pacmancoder:
@awakecoding I have fixed merge conflicts and bumped up IronRDP version for this PR.
I have also found that "cargo clippy --all-targets --all-features -- -D warnings" shows a lot of errors after rebasing this branch to the new master, If you don't mind, I can fix these issues in this PR.

FYI: I am using ci/test.sh script to validate Jet locally, It would be great if we can add this script execution to our pipelines, so Jet will be checked not only for buildability but also for formatting, clippy and tests.

awakecoding:
@pacmancoder let's just merge these functional changes first to avoid mixing them with formatting changes.

PR #105, opened 2020-10-20
Clean up legacy association creation code

bcortier-devolutions:


fdubois1 (commented):
Before doing this, I would ask @awakecoding  if he wants to change the specification for that. If we change it, maybe we would like to remove completely the create_association request.  

But we could continue to use a create association in a setup with only one JET so I don't know if we want to completely remove that...

awakecoding:
@fdubois1 in its current form it remains broken by design, unless we literally make the call on all "farm members" by looping over the list and making the same call on each one of them individually. That's some serious duct tape solution, but it avoids the trouble of using a centralized database. I have mixed feelings about removing it entirely - is it still using the API key, or doing the same kind of validation as done with the candidate gathering phase?

awakecoding:
@bcortier-devolutions if you can just modify it to do the same token validation as for the candidate gathering, I think it would be good enough. We won't be using it, but it would at least remove the API key usage for this call.

bcortier-devolutions:
@awakecoding done :+1:

PR #109, opened 2020-10-26
Cli update

fdubois1:
- All env. variable are now DGATEWAY_XYZ

Parameters added:
- farm-name
- application-protocols (only rdp and wayk can be specified for now. Only rdp is used to replace old rdp parameter)

Parameters renamed:
- jet-instance --> hostname

Parameters removed:
- rdp

Parameters updated:
- certificate-file and private-key-file can specify a full path OR a filename and the config path will be added
- listeners has to be defined with * instead of <jet_instance> (example : -l tcp://*:8080,tcp://*:8080)

Other:
- Health check will return the string : "Devolutions Gateway "<hostname>" is alive and healthy."

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -66,7 +66,7 @@ pub enum Protocol {
 
 #[derive(Debug, Clone)]
 pub struct ListenerConfig {
-    pub url: Url,
+    pub internal_url: Url,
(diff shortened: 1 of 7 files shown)


awakecoding (approved):
LGTM

PR #124, opened 2020-12-09
Async crates upgrade

pacmancoder:
This PR upgrades devolutions-gateway project to the new async crates - futures 0.3 (with async & await), tokio 0.3 (in some places tokio 0.2 was required for third-party libraries e.g saphir & hyper), and a lot of dependencies were also upgraded to the newest versions with tokio 0.3 support.

Besides this PR, some additional changes were performed in the other repos along the way:
- Saphir graceful shutdown fix  https://github.com/richerarc/saphir/pull/128
- Applied our already existing rustls patch for disabling server name identification https://github.com/Devolutions/rustls/tree/v0.18.1-patched


Cargo.toml
```diff
@@ -48,21 +47,33 @@ failure = "0.1"
...
-tungstenite = "0.10"
+hyper = "0.13"
 sha1 = "0.6.0"
 base64 = "0.9.1"
 slog-stdlog = "4.0"
 ring = "0.16.0"
 spsc-bip-buffer = { git = "https://github.com/Devolutions/spsc-bip-buffer.git", branch = "master" }
 indexmap = "1.0"
 
+[dependencies.saphir]
```
pacmancoder, line 58: TODO: change to mainstream repo / published version when PR https://github.com/richerarc/saphir/pull/128 will be merged

src/jet_client.rs
```diff
@@ -14,21 +18,21 @@ use jet_proto::{
...
     jet::{
         association::Association,
         candidate::{Candidate, CandidateState},
         TransportType,
     },
     transport::JetTransport,
-    utils::association::{RemoveAssociation, ACCEPT_REQUEST_TIMEOUT_SEC},
+    utils::association::{RemoveAssociation, ACCEPT_REQUEST_TIMEOUT},
     Proxy,
 };
```
pacmancoder, line 38: JetAssociationsMap above was changed to tokio Mutex, which should play more nicely with async code

awakecoding:
@bcortier-devolutions

Cargo.toml
bcortier-devolutions: Looks like it just got merged :tada:

src/http/controllers/health.rs
```diff
@@ -1,39 +1,31 @@
...
 }
 
-fn health(controller: &ControllerData, _req: &SyncRequest, res: &mut SyncResponse) {
-    build_health_response(res, &controller.config.hostname);
-}
-
-pub fn build_health_response(res: &mut SyncResponse, hostname: &str) {
-    res.status(StatusCode::OK)
-        .body(format!("Devolutions Gateway \"{}\" is alive and healthy.", hostname));
+pub fn build_health_response(config: &Config) -> (u16, String) {
```
bcortier-devolutions, line 27: I think you could return `StatusCode` directly instead of `u16`
Doing so you can remove all the `as_u16` in all handlers

src/http/controllers/jet.rs
```diff
@@ -1,198 +1,147 @@
...
-                    return;
+                    return (StatusCode::FORBIDDEN.as_u16(), None);
                 }
                 Ok(_) => { /* alright */ }
             }
         }
 
         // create association
-        let mut jet_associations = self.jet_associations.lock().unwrap(); // no need to deal with lock poisoning
+        let mut jet_associations = self.jet_associations.lock().compat().await; // no need to deal with lock poisoning
```
bcortier-devolutions, line 139: You can remove this comment since there is no lock poisoning check with tokio Mutex

bcortier-devolutions (changes requested):
Great job!
I reviewed 18/41 files for now, I'll continue later

bcortier-devolutions (changes requested):
Really good work!
I left some suggestions, but it looks good. I was also able to run some Wayk sessions without trouble.

bcortier-devolutions:
I'll merge now so that we can start adding new features on mainstream without creating conflicts. You could address my comments via another PR, or I can do it myself depending on what is more convenient for you. Let me know :slightly_smiling_face:

pacmancoder:
Sure, I will resolve the issues described in the comments and prepare the new PR 👍

src/http/controllers/health.rs
pacmancoder: Thanks, I didn't know that it is possible to use StatusCode directly -- in saphir docs examples there is the only example with u16 return code type :smile:

src/http/controllers/jet.rs
pacmancoder: Yep, forgot to clean-up the comments, thanks!

src/jet_client.rs
```diff
@@ -168,7 +161,8 @@ impl Future for JetMsgReader {
                 jet_proto::JET_MSG_HEADER_SIZE
             );
 
-            Ok(Async::NotReady)
+            cx.waker().clone().wake();
```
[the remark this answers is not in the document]
pacmancoder: Well, Indeed, it looks like it was a bug because normally we should call `wake` to ensure our future will be called if we are returning Pending manually (not while propagating it from another async function, but when actually returning if the first time in the call chain). In the tokio 0.1 it could be done with `futures::task::current().notify();` (e.g. in the old version of src/utils/association.rs we were doing so when call to try_lock of the non-async mutex returned `Err`. 
As I understand tokio 0.1 still _could_ poll the future when it's not notified anything, but it is not guaranteed, so maybe that's because we aren't faced any consequences of this bug yet.

src/jet_client.rs
```diff
@@ -196,25 +189,29 @@ impl HandleAcceptJetMsg {
...
+    fn handle_create_response(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Result<Vec<u8>, io::Error>> {
+        let Self {
+            jet_associations,
+            request_msg,
+            remove_association_future,
+            association_uuid,
+            config,
+            ..
+        } = self.deref_mut();
+        let (status_code, association) = if let Ok(mut jet_associations) = jet_associations.try_lock() {
```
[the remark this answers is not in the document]
pacmancoder: I agree, will change that 👍

src/proxy.rs
```diff
@@ -90,25 +95,27 @@ impl Proxy {
...
+                .as_mut()
+                .set_packet_interceptor(Box::new(interceptor.clone()));
+            jet_stream_client.as_mut().set_packet_interceptor(Box::new(interceptor));
         }
 
+        // Create trait object wrappers to achieve Stream + Sized type
+        let jet_stream_server = SizedStream::new(jet_stream_server);
+        let jet_stream_client = SizedStream::new(jet_stream_client);
+        let jet_sink_client = SizedSink::new(jet_sink_client);
+        let jet_sink_server = SizedSink::new(jet_sink_server);
```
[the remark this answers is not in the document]
pacmancoder: Thank you for pointing this out! I was previously imported StreamExt from the tokio-utils crate, which indeed, required Sized: https://docs.rs/futures-util/0.3.8/futures_util/stream/trait.StreamExt.html#method.forward later I am switched to StreamExt  from futures and didn't realized that it does not require Sized 😄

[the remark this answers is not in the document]
bcortier-devolutions: Oh! I see now! :smile:

src/jet_client.rs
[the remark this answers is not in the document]
bcortier-devolutions: Great, thank you!

src/jet_client.rs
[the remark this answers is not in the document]
bcortier-devolutions: I understand, good catch!

src/http/controllers/health.rs
bcortier-devolutions: Haha maybe doc could be improved to discover this more easily: you can actually return anything implementing their `Responder` trait, including:
https://docs.rs/saphir/2.8.0/saphir/responder/trait.Responder.html#foreign-impls

PR #143, opened 2021-02-04
Rdp connectivity

pacmancoder:
This TR implements RDP rendezvous connections. Basically, this provides possibility to connect over Gateway with the following steps:
- Agent RDP machine connects to the Gateway over JET-TCP tunnel, opened by Jetsocat (Jetsocat on the one side connects to the local RDP port, and to the Gateway using assicoation/candidate id from the other side)
- Client machine connects to the Agent over Gateway via mstsc or xfreerdp by providing session token which includes association id for Gateway inside it). In case of WaykNow this token generates inside WaykBastion.


jetsocat/src/tcp_proxy.rs
```diff
@@ -0,0 +1,186 @@
...
+            association_id,
+            candidate_id,
+            auto_reconnect,
+        }
+    }
+
+    pub async fn serve(self, log: Logger) -> Result<()> {
+        debug!(log, "Performing rendezvous connect...");
+
+        loop {
```
awakecoding, line 44: @pacmancoder can you explain the reason for the loop? I believe it may be related to the problem where RDP makes multiple TCP connections during the negotiation step? Features like RDP auto-reconnect probably won't work without requesting a new JWT to open the connection, because the original JWT would be expired. The problem I could see is this might loop forever, we need a proper exit condition.

pacmancoder: Yes, the reasoning behind this is that mstsc reconnects during negotiation step. This happens once during the connection. Current jwt token validity interval, which is set from den is enough to perform reconnection. Maybe one improvement which we can make is to limit possible reconnection instead of loop-until-fail approach (the loop currently exits when Gateway kills association because of inactivity)

jetsocat/src/main.rs
```diff
@@ -113,13 +114,33 @@ fn listen_command() -> Command {
...
     let res = ServerArgs::parse("listen", c).and_then(|args| {
         let log = setup_logger(args.common.logging);
-        run(log.clone(), jetsocat::server::listen(args.common.addr, args.pipe, log))
+        run(log.clone(), jetsocat::server::listen(args.common.addr, args.cmd, log))
+    });
+    exit(res);
+}
+
+fn proxy_command() -> Command {
+    let cmd = Command::new("tcp-proxy")
```
CBenoit, line 129: Since this is mostly specific code for RDP rendezvous mode, I suggest to rename "tcp-proxy" to "rdp-tunnel" or "rdp-forward" for now. At least I'm in favor of using "tunnel" or "forward" instead of "proxy" (possible confusion with SOCKS/HTTP proxy also supported by jetsocat)

CBenoit (changes requested):
Good job!
As an aside, I noticed jetsocat logs were empty when started by wayknow service:
image
Do you have an idea?

jetsocat/src/main.rs
pacmancoder: The current code in jetsocat for "tcp-proxy" could handle any tcp port handling, so maybe it is better to rename it to "tcp-tunnel-accept"? What do you think?

jetsocat/src/tcp_proxy.rs
pacmancoder: Well, in the case of RDP this code is indeed required to make RDP forwarding function normally. However, in general, for general tcp-forwarding, I think this is optional. Let's pass auto-reconnect retries  from CLI (--max-reconnects=3), what do you think?

CBenoit: Indeed I prefer the `--max-reconnects` approach!

jetsocat/src/main.rs
CBenoit: I asked Marc-André about whether we should make it rdp specific or not, I think he will clarify this soon. If we keep it general over TCP then `tcp-tunnel-accept` sounds good to me! :+1:

CBenoit: @pacmancoder Hi! I discussed with Marc-André, we agreed on "jet-tcp-accept" :slightly_smiling_face:

jetsocat/src/tcp_proxy.rs
CBenoit: @pacmancoder I discussed with Marc-André and we agreed on renaming to `--max-reconnection-count` with a default value of 0 (only initial connection).
Loop would be:
```rust
for _ in 0..=2 {
    …
}
```
(notice the ..**=**)

jetsocat/Cargo.toml
```diff
@@ -13,6 +13,8 @@ rustls = ["async-tungstenite/tokio-rustls"]
 native-tls = ["async-tungstenite/tokio-native-tls"]
 
 [dependencies]
+jet-proto = { path = "../jet-proto" }
+uuid = { version = "0.7.1", features = ["v4"] }
```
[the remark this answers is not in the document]
pacmancoder: We are using the older crate version (0.7.x) in jet-proto crate which we depend on, upgrade it jet-proto and revolutions gateway too?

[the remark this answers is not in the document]
CBenoit: Yes, there is no breaking change :+1:

pacmancoder:
> 
> 
> Good job!
> As an aside, I noticed jetsocat logs were empty when started by wayknow service:
> image
> Do you have an idea?

That's strange, log are initializing identiacally to other jetsocat commands... not yet investigated what can cause this to happen

pacmancoder:
@CBenoit Hm, I looked at logs by myself, can't reproduce this issue. on my machine logs contain log messages, they are not empty :(

pacmancoder:
FYI @awakecoding @CBenoit Updated PR, Fixed all issues discussed above

CBenoit (approved):
Great thank you! I'm merging! :tada:

CBenoit (approved):

PR #147, opened 2021-03-16
Session recording inside gateway

AnastasiiaRomaniuk:
- added handling of the jet-tp claim
- added ffi to communicate with dll
- fixed tests

### DIFF
--- a/devolutions-gateway/src/config.rs
+++ b/devolutions-gateway/src/config.rs
@@ -490,9 +490,9 @@ impl Config {
                 Arg::with_name(ARG_PLUGINS)
                     .long("plugin")
                     .value_name("PATH")
-                    .help("An path where the plugin is located including the plugin name and plugin extension.")
+                    .help("A path where the plugin is located including the plugin name and plugin extension.")
                     .long_help(
-                        "An path where the plugin is located including the plugin name and plugin extension. \
+                        "A path where the plugin is located including the plugin name and plugin extension. \
                     The plugin will be loaded as dynamic library. \
                     For example, on linux  - home/usr/libs/libplugin.so \
                     on Windows - D:\\libs\\plugin.dll.",
@@ -507,9 +507,9 @@ impl Config {
                 Arg::with_name(ARG_RECORDING_PATH)
                     .long("recording-path")
                     .value_name("PATH")
-                    .help("An path where the recording of the session wil be located.")
+                    .help("A path where the recording of the session will be located.")
                     .long_help(
-                        "An path where the recording will be saved. \
+                        "A path where the recording will be saved. \
                     If not set the TEMP directory will be used.",
                     )
                     .takes_value(true)
--- a/devolutions-gateway/src/config.rs
+++ b/devolutions-gateway/src/config.rs
@@ -1,15 +1,15 @@
+use crate::plugin_manager::PLUGIN_MANAGER;
 use cfg_if::cfg_if;
 use clap::{crate_name, crate_version, App, Arg};
 use picky::key::{PrivateKey, PublicKey};
 use picky::pem::Pem;
 use serde::{Deserialize, Serialize};
+use slog_scope::debug;
 use std::env;
 use std::fs::File;
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
 use url::Url;
-use slog_scope::debug;
-use crate::plugin_manager::PLUGIN_MANAGER;
 
 const ARG_API_KEY: &str = "api-key";
 const ARG_APPLICATION_PROTOCOLS: &str = "application-protocols";
@@ -490,9 +490,7 @@ impl Config {
                 Arg::with_name(ARG_PLUGINS)
                     .long("plugin")
                     .value_name("PATH")
-                    .help(
-                        "An path where the plugin is located including the plugin name and plugin extension.",
-                    )
+                    .help("An path where the plugin is located including the plugin name and plugin extension.")
                     .long_help(
                         "An path where the plugin is located including the plugin name and plugin extension. \
                     The plugin will be loaded as dynamic library. \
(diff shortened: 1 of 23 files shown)


jet-proto/Cargo.toml
```diff
@@ -12,3 +12,22 @@ httparse = "1.3.4"
 http = "0.1.18"
 
 hex-literal = "0.2"
+
+[dev-dependencies]
```
awakecoding, line 16: hum... where are there so many *dev-dependencies* all of sudden? dev dependencies are only required for building, but not for the actual crate itself: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies

this looks like a mistake?

devolutions-gateway/src/http/controllers/jet.rs
```diff
@@ -106,29 +114,38 @@ impl JetController {
...
 
         // check the session token is signed by our provider if unrestricted mode is not set
-        if !self.config.unrestricted {
-            match validate_session_token(self.config.as_ref(), &req) {
-                Err(e) => {
-                    slog_scope::error!("Couldn't validate session token: {}", e);
-                    return (StatusCode::UNAUTHORIZED, None);
-                }
-                Ok(expected_id) if expected_id != association_id => {
+        let jet_tp_claim = match validate_session_token(self.config.as_ref(), &req) {
```
awakecoding, line 117: I'm not sure just by looking at the code what would happen if the jet_tp claim is absent from the token? Devolutions Gateway needs to function properly with older versions of the code. If the jet_tp claim is omitted, it can be assumed not to record anything. Since the tokens are signed by Wayk Bastion anyway, there is no issue with trust. only versions of Wayk Bastion that can enable session recording will set this field, older versions will omit it.

devolutions-gateway/src/http/controllers/jet.rs
```diff
@@ -64,31 +64,39 @@ impl JetController {
...
         // check the session token is signed by our provider if unrestricted mode is not set
-        if !self.config.unrestricted {
-            match validate_session_token(self.config.as_ref(), &req) {
-                Err(e) => {
-                    slog_scope::error!("Couldn't validate session token: {}", e);
-
-                    return (StatusCode::UNAUTHORIZED, ());
-                }
-                Ok(expected_id) if expected_id != association_id => {
+        let jet_tp_claim = match validate_session_token(self.config.as_ref(), &req) {
```
awakecoding, line 67: same comment here as in the other place where the jet_tp claim is extracted: does this correctly handle the case where the token is omitted? old versions of Wayk Bastion will sign tokens that don't contain this claim, and we need to handle it properly (no recording when the claim is absent).

awakecoding (changes requested):
I left a few comments in the code, but can you confirm the following with me:

none of the existing functionality should be impacted by merging this, such that Devolutions Gateway without plugins installed, and session tokens that don't have the jet_tp claim will still be processed as they used to be (no recording, nothing breaks).

If this is the case, just go over the comments and it should be good for a merge.

devolutions-gateway/src/interceptor/pcap_recording.rs
```diff
@@ -0,0 +1,83 @@
...
+
+            let (status, message_id) = parser.parse_message(data, data.len(), is_from_server);
+
+            if !parser.is_message_constructed() {
+                return;
+            } else if message_id == PacketsParser::NOW_UPDATE_MSG_ID {
+                let size = parser.get_size();
+                let image_data = parser.get_image_data();
+                if option_recorder.is_some() {
+                    let recorder = option_recorder.as_ref().unwrap();
```
fdubois1, line 65: instead of doing an unwrap, I think we could use something like that : 
`if let Some(recorder) = option_recorder.as_ref()`

devolutions-gateway/src/jet_client.rs
```diff
@@ -73,6 +83,73 @@ impl JetClient {
...
+    response: HandleConnectJetMsgResponse,
+    tls_acceptor: TlsAcceptor,
+) -> Result<(), io::Error> {
+    let mut recording_interceptor: Option<PcapRecordingInterceptor> = None;
+    let association_id = response.association_id;
+
+    let associations = jet_associations.lock().await;
+    if let Some(association) = associations.get(&association_id) {
+        if let Some(jet_tp_claim) = association.get_jet_tp_claim() {
+            if jet_tp_claim.eq(EXPECTED_JET_TP_VALUE) && config.plugins.is_some() {
```
fdubois1, line 127: Similar to what we did in den-server, I think jet_tp_claim should be an enum with all possibles values. So instead of looking if we have a specific string, we could validate an enum

fdubois1 (changes requested):
I left a few comments in the code. No big issue. Good job !!

jet-proto/Cargo.toml
AnastasiiaRomaniuk: I added all of these, because the tests were failing as they were not able to resolve the functions that belong to those dependencies. This all belongs to the tests folder in the jet-proto directory.

devolutions-gateway/src/http/controllers/jet.rs
AnastasiiaRomaniuk: yes, the jet_tp claim has the type Option<>. So if the jet_tp claim will be absent, the value will be set to None and the recording will not start.

devolutions-gateway/src/http/controllers/jet.rs
AnastasiiaRomaniuk: yes, the jet_tp claim has the type Option<>. So if the jet_tp claim will be absent, the value will be set to None and the recording will not start.

awakecoding (dismissed):
LGTM

devolutions-gateway/src/config.rs
```diff
@@ -473,6 +485,42 @@ impl Config {
...
+                    .use_delimiter(true)
+                    .value_delimiter(";")
+                    .takes_value(true)
+                    .number_of_values(1),
+            )
+            .arg(
+                Arg::with_name(ARG_RECORDING_PATH)
+                    .long("recording-path")
+                    .value_name("PATH")
+                    .help("An path where the recording of the session wil be located.")
```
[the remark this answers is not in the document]
AnastasiiaRomaniuk: done

devolutions-gateway/src/config.rs
```diff
@@ -473,6 +485,42 @@ impl Config {
                     .long("service")
                     .takes_value(false)
                     .help("Enable service mode"),
+            )
+            .arg(
+                Arg::with_name(ARG_PLUGINS)
+                    .long("plugin")
+                    .value_name("PATH")
+                    .help("An path where the plugin is located including the plugin name and plugin extension.")
```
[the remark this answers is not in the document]
AnastasiiaRomaniuk: done

devolutions-gateway/src/jet_client.rs
AnastasiiaRomaniuk: done

devolutions-gateway/src/interceptor/pcap_recording.rs
AnastasiiaRomaniuk: done

fdubois1 (approved):
LGTM

PR #149, opened 2021-03-26
Move Dockerfiles out of the project root

awakecoding:
The dev and production release pipelines will need to be updated after this change.


fdubois1 (approved):
LGTM

PR #153, opened 2021-04-12
Bugfix session recording

AnastasiiaRomaniuk:
Added creating of additional frames by timeout.
Added path api for the feature sogar integration.
Updated api according to the changes in the libarary.

### DIFF
--- a/devolutions-gateway/src/interceptor/pcap_recording.rs
+++ b/devolutions-gateway/src/interceptor/pcap_recording.rs
@@ -1,6 +1,6 @@
 use crate::interceptor::{PacketInterceptor, PeerInfo};
 use crate::plugin_manager::{PacketsParser, Recorder, PLUGIN_MANAGER};
-use slog_scope::debug;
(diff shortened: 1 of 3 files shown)


devolutions-gateway/src/interceptor/pcap_recording.rs
```diff
@@ -46,15 +92,27 @@ impl PacketInterceptor for PcapRecordingInterceptor {
         debug!("New packet intercepted. Packet size = {}", data.len());
 
         let server_info = self.server_info.lock().unwrap();
+        let is_from_server = source_addr.unwrap() == server_info.addr;
```
fdubois1, line 95: source_addr: Could it be none and panic ?

AnastasiiaRomaniuk: for now, the answer is not. But I added the check up in case if something will change.

fdubois1 (approved):
LGTM

PR #165, opened 2021-06-11
Jetsocat fixes

CBenoit:
* Fix process piping.
Process was closed early.
To fix this, a handle is kept alive until piping operation ends.

* Fix WebSocket read buffer advancing.
Buffer was not marked as filled causing WebSocket to close itself.

* Flush on write. `sftp` works thanks to this.

* Add `ws-listen` pipe. Useful for debugging but might be used for other purposes as well.


awakecoding (approved):
LGTM

PR #169, opened 2021-07-15
Don't use sogar_cli to push files to remote registry 

fdubois1:
1 - Sogar update
2 - Don't use sogar_cli to push files. We use sogar_core since we already have a dependency on it
3 - Remove sogar_path from config, not needed since we don't use sogar_cli


CBenoit (approved):
LGTM!

PR #172, opened 2021-07-30
Wayk 2564 - Add endpoint /sessions to list sessions + Fix tests

fdubois1:
A GatewaySession contains all JET fields from a session token (jet_ap, jet_aid, etc.)

A few changes in the token definition : 
jet_tp has been removed. 
jet_rec has been added: recording policy: true/false
jet_flt has been added: filtering policy: true/false

I will update the spec, this is why this PR is a draft, I will push the spec as soon as possible, but the code should not change


awakecoding (approved):
LGTM

CBenoit:
I’ll have a look after vacations, you can merge before if you prefer

PR #173, opened 2021-08-02
Fix powershell build: Remove nuget cert revocation check

fdubois1:
try to remove the nuget cert revocation check and see if it fixes the issue on the CI


awakecoding (approved):
LGTM

PR #174, opened 2021-08-03
Wayk 2566 - Review all authorization in Devolutions Gateway

fdubois1:
I reviewed all authorization. Before, authorization was done on each route. The auth middleware was only there to validate the api_key. Now, the auth middleware validate that we have a token signed by the provisioner private key. It exists (for now), 3 types of token (see below). If a valid token is passed, the request is accepted. Then, there is a guard applied on each route to accept only a specific token type. For example, GET /sessions accept only a "scope" token with "gateway.sessions.read" as scope. Otherwise, forbidden is returned. 

Here are the different tokens : 
 
**Association token defined like this :** 
```
{
  "iat": 1628002185,
  "nbf": 1628002185,
  "exp": 1628002305,
  "type": "association",
  "jet_ap": "wayk",
  "jet_cm": "rdv",
  "jet_aid": "0e8cc760-133d-4422-9b5e-77ce39d7a204"
}
```

**Scope token defined like this :** 
```
{
  "iat": 1628001490,
  "nbf": 1628001490,
  "exp": 1628001610,
  "type": "scope",
  "scope": "gateway.sessions.read"
}
```

The scope supported for now are : "gateway.sessions.read" and "gateway.associations.read". For now, I didn't remove the endpoint to get association (GET /association) in case it would be useful along the road, but I added security on it and scope "gateway.associations.read" is required to do that request.

**Bridge token is defined like this :**
```
{
  "iat": 1628001490,
  "nbf": 1628001490,
  "exp": 1628001610,
  "type": "bridge",
  "target": "https://google.com"
}
```

Other things to note :
- The token validation is done with a leeway of 10 minutes (as specify in the specification).
- The jet_aid should always be specified in the association token. BUT, the current DVLS version doesn't set it so to be backward compatible, an ID will be generated if it is not specified, but all future implementation should specify an ID
- I updated the spec to add description of different scopes and different API (/sessions, /bridge).
- I removed API_KEY, it can no longer be used to authorize a request
- I removed the unrestricted mode. It is always restricted and a token as to be provided


awakecoding (approved):
LGTM

PR #176, opened 2021-08-03
WAYK-2567: Add endpoint to get logs (GET /diagnostics/logs)

fdubois1:
It requires a scope token with scope set to "gateway.diagnostics.read"


awakecoding (approved):
LGTM

PR #182, opened 2021-08-12
Implement JMUX multiplexing protocol

Dezdor:
Implemented JMUX multiplexing protocol for tcp:// and tcp-listen:// connection types
    JMUX spec: https://github.com/awakecoding/qmux/blob/protocol-update/SPEC.md

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -754,7 +754,7 @@ dependencies = [
 
 [[package]]
 name = "devolutions-gateway"
-version = "2021.1.5"
+version = "2021.1.6"
 dependencies = [
  "base64 0.9.3",
  "byteorder",
@@ -1516,6 +1516,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "idalloc"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "559e07f7d3b6a6543b8f3a3f09fef45c468e3b6a1d49dd02369fb79363fae3a8"
+
+[[package]]
 name = "idna"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1640,16 +1646,19 @@ dependencies = [
 
 [[package]]
 name = "jetsocat"
-version = "2021.1.5"
+version = "2021.1.6"
 dependencies = [
  "anyhow",
  "async-tungstenite",
+ "byteorder",
  "dirs-next",
  "futures-channel",
  "futures-io",
  "futures-util",
+ "idalloc",
  "jet-proto",
  "jetsocat-proxy",
+ "min-max",
  "proxy_cfg",
  "seahorse",
  "slog",
@@ -1837,6 +1846,15 @@ checksum = "2684d4c2e97d99848d30b324b00c8fcc7e5c897b7cbb5819b09e7c90e8baf212"
 dependencies = [
  "mime",
  "unicase",
+]
+
+[[package]]
+name = "min-max"
+version = "0.1.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe75516622e34f7e712e0f60b6a9bb1b2df281d1985c11a21fa5dabc1e34885f"
+dependencies = [
+ "partial-min-max",
 ]
 
 [[package]]
@@ -2258,6 +2276,12 @@ dependencies = [
  "smallvec 1.6.1",
  "winapi 0.3.9",
 ]
+
+[[package]]
+name = "partial-min-max"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6448add382c60bbbc64f9dab41309a12ec530c05191601042f911356ac09758c"
 
 [[package]]
 name = "pcap-file"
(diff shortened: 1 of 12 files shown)


awakecoding:
@bcortier-devolutions if you can review this PR please

CBenoit:
@awakecoding looks like something related to openssl and MacOS is failing in CI, do you have an idea?

awakecoding:
> @awakecoding looks like something related to openssl and MacOS is failing in CI, do you have an idea?

it's probably related to the recent chances we've done in the conan packages, let me take a look

CBenoit:
@Dezdor Hello!

Can you send me a few usage examples?

Here is an usage that didn’t worked as I expected:
image

Another one:
image

What is the expected usage with the current API?

Dezdor:
JMUX protocol works only with TCP (for another protocols it doesn't have a sense)

For an example (my test cases):
1.
image

```
./jetsocat forward tcp-listen-multiplexing://127.0.0.1:8001 tcp://127.0.0.1:8002 --log-term
                            \ A1 /                        \  B1 /

./jetsocat forward tcp-listen://127.0.0.1:8002 tcp-multiplexing://127.0.0.1:8003 --log-term
                            \ A2 /                        \  B2 /
```
In this case, we have next "channel":` A1 -> B1 -> A2 -> B2`.  At the start and end of this "channel" must be `tcp-listen-multiplexing` and `tcp-multiplexing`. In the middle can be TCP and WS.

2. 
image
```
./jetsocat forward tcp-listen-multiplexing://127.0.0.1:9001 tcp-multiplexing://127.0.0.1:9002 --log-term
```

---
your first case has no sense

The second one need to change to

1 console (up)
`jetsocat forward tcp-listen-multiplexing://localhost:9999 tcp-multiplexing://localhost:10000 --log-term`
2 console (right)
`jetsocat forward  tcp-listen://localhost:10000 stdio`
3 console (down)
`jetsocat forward stdio tcp://localhost:9999`

CBenoit (approved):
I see, thank you!

Here is my review report.

### Review points

- Audited new dependencies: no problem.

- About packets encoding and decoding:
  - Parsing a `JmuxMsgChannelOpen` message containing a destination URL which is not valid UTF-8 is causing a panic instead of bubbling up the error. Same applies for `JmuxMsgChannelOpenFailure`’s description.
  - Otherwise, everything looks correct (tests are thorough).

- `Marshaler` trait is forcing `Vec` heap allocation which is inefficient for our use case.
	The right approach is to use `AsyncWrite` trait and/or [`Bytes`](https://docs.rs/bytes/1.0.1/bytes/struct.Bytes.html)/[`BytesMut`](https://docs.rs/bytes/1.0.1/bytes/struct.BytesMut.html) objects.
	[`tokio_util::codec`](https://docs.rs/tokio-util/0.6.7/tokio_util/codec/index.html) module is specifically designed to facilitate this approach.

- As is, JMUX-related CLI API is allowing user to express indesirable configurations.
  Briefly, there are two distinct sides (as you pointed):

		[Client] (TCP socks server <-> JMUX-client wrapped pipe) <-> [Proxy] (JMUX-server wrapped pipe <-> variable TCP destinations)

  Both are not expected to be as flexible as the generic `forward <PIPE> <PIPE>` but should rather be more constrained in specific ways:

  - On [Client] side, a binding address and a generic pipe on which to send and receive JMUX-client stream.
  - On [Proxy] side, a generic pipe from which to receive and send JMUX-server stream.

  A good CLI API should guide user into expressing a valid configuration and avoid ending up bloated with incompatible options as much as possible.

  In current implementation:

  - It’s not possible to properly express the [Proxy] side which should be a single JMUX-server wrapped pipe.
  Destination must be determined through the `destinationUrl` field in the `CHANNEL OPEN` message, but in the current CLI API the [Proxy] side is also specifying the destination url and `destinationUrl` field is simply ignored.

  [Relevant part](https://github.com/Devolutions/devolutions-gateway/blob/f7541b87832bd22567ef9011c96fa0863519fa7c/jetsocat/src/jmux.rs#L150):

	```rust
	pub async fn jmux_connect_loop(address: String, log: Logger) -> Result<Pipe> {
		…
		JMUXChannelMsg::Open(msg) => {
			…
			let address = address.clone();
			tokio::spawn({
				let connection_id = msg.sender_channel_id;
				let window_size = msg.initial_window_size;
				let maximum_packet_size = msg.maximum_packet_size;
				…
				let (read_half, write_half) = match TcpStream::connect(address.clone()).await {
		…
	```

  - [Client] side is kind of fine, but the TCP listener approach (as opposed to SOCKS5) should be configured along a destination URL (which should not be configured on the proxy side as said above).

- Point above is simplified by considering channels can only be opened from "client" to a "proxy", and it’s fine because AFAIK you were asked to implement it that way.
	In fact both sides are equally able to act as a normal JMUX proxy: both sides are just "JMUX peers" and "[either side may open a channel](https://github.com/awakecoding/qmux/blob/eb1261cb3f78599c13834a3c260c5e50dfc31145/SPEC.md#channels)".

### My changes

- Rebase on master and clean commit message.

- Update dependencies to latest.

- Rename module `crate::jmux::jmux_proto` into `crate::jmux::proto`.

### Remarks / Future work

- Maybe a sub-crate `jmux` or `jmux-proto` for protocol-specific code could have been interesting (similar to `jetsocat-proxy`, better separation, reusability and compilation time).

- SOCKS5 TCP listener.

- Support bi-lateral channel opening.

------

A follow up PR will be open.

PR #189, opened 2021-10-20
GitHub workflow

thenextman:
Port the CI build into GitHub workflow

- A patch is required for `embed-resource` build dependency, I raised a PR for the issue. Alternatively we could manually set the Windows SDK include paths
- Artifacts are not code-signed, as the relevant secrets are not yet available. Once they're made available in a properly configured environment, code-signing will probably be refactored into new "release {dev/production}" jobs. Currently the workflow prints a notice that artifacts are *not* code-signed.

Performance is roughly on-par with Azure, since we are able to parallelize the builds


CBenoit (approved):
Good news :tada:

PR #195, opened 2021-11-01
CI and release pipelines

thenextman:
- Only enable LTO when building master
- Put rust channel into a toolchain file
- Skeleton yml files for release pipeline


awakecoding (approved):
LGTM

PR #200, opened 2021-11-09
Fix docker image in release pipeline

thenextman:
Ensure we set the Docker base image when generating a release


awakecoding (approved):
LGTM

PR #206, opened 2021-11-24
DPS-5255: support for alternate hosts

CBenoit:
A new optional claim `dst_hst` is added to the association token.
A list of target addresses can be specified through it.
A connection will be attempted to each host successively until one succeeds, or all fail.

Todo:

- [x] implementation
- [x] tests


CBenoit:
@fdubois1 PR is now ready for review.

fdubois1 (approved):
LGTM

PR #207, opened 2021-11-24
WAYK-2591: Update installer

thenextman:
Remove Wayk references from the installer. I tested by installing the version 2021.1.6; then building a new 2021.1.7 and upgrading. The firewall rules are properly recreated.

`ProductDescription` could be localized, but note that only the en-US version is displayed anywhere.

I removed some redundant CI files.


CBenoit (approved):
LGTM

PR #209, opened 2021-11-29
jmux-proxy: improve API

CBenoit:
- Use a bounded channel for the API request instead of an unbounded one
  (provides backpressure)

  - Internal channels are still unbounded because this is faster and the
    JUMX protocol's windowing is already providing backpressure.

- Use a oneshot channel for the API response.

- Builder pattern to configure the proxy instead of the 6-parameters
  function.


fdubois1 (approved):
LGTM

PR #210, opened 2021-12-01
DPS-5297: mitigate token reuse

CBenoit:
If a given token is reused by a peer with a different IP address, request is rejected.
A cache with a few information extracted from the token is maintained to make this possible.


fdubois1 (approved):
LGTM

PR #211, opened 2021-12-02
DPS-5332: default port based on protocol claim

CBenoit:
A custom `serde::de::Deserialize` trait implementation is used in order to dynamically validate target addresses based on the application protocol claim provided in the same token:

- If the application protocol claim is set to a known protocol for which the default port is known, then the port can be omitted.
- Otherwise, the port must be included or token will be rejected.

This PR also adds timeout to host connections.


fdubois1 (approved):
LGTM

PR #214, opened 2021-12-07
Custom cargo profile

thenextman:
Update the toolchain to rust 1.57 and use a custom "production" cargo profile to enable LTO on master builds.


CBenoit (approved):
LGTM!

PR #216, opened 2021-12-07
Enable releases

thenextman:
Fix issues with the release workflow:
- Uncomment the final steps of the release pipeline to actually allow the release
- Fix the syntax and auth of the `gh release create` command
- Don't run the workflow if the specified run isn't from master, or it didn't complete successfully
- Don't fail the workflow if the PS module version is already published


CBenoit (approved):
LGTM!

PR #222, opened 2021-12-13
Run tests in CI

thenextman:


CBenoit (commented):
Thank you!

When running `cargo test`, `--workspace` argument should be appended as well otherwise it only run tests for `jetsocat` and `devolutions-gateway` (the default members of the workspace). I guess a new parameter to `tlk.ps1` would do the trick?

CBenoit (approved):
LGTM :+1:

PR #231, opened 2022-01-31
KDC proxy in DevolutionsGateway

RRRadicalEdward:
Hi! :slightly_smiling_face: 
The task is to implement the `KDC` proxy for the `Kerberos` protocol inside `DevolutionsGateway`. `Kerberos` is a computer network authentication protocol that works with tickets to allow nodes communicating over a non-secure network to prove their identity to one another in a secure manner. 
`KDC` proxy is used when clients don't know where the target `KDC` server is or can't reach it.`KDC` proxy adds an additional layer of security as it allows to deploy `KDC` servers to private networks.

What has been implemented:
- `picky-krb` subproject with all the needed for`KDC` proxy data types. See 5.2 Basic Kerberos Types [Kerberos protocol specification](https://www.rfc-editor.org/rfc/rfc4120.txt).
-  KDC proxy configuration added to the `Config`.
- `KdcProxyController` with `/KdcProxy` `HTTPS` endpoint that transmits incoming `KDC-PROXY-MESSAGE`'s `kerb-message` to the target `KDC` server and returns `KDC` server reply to the client.

Useful links:
- [KDC Proxy Protocol specification](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-KKDCP/%5bMS-KKDCP%5d.pdf)
- [Kerberos protocol specification](https://www.rfc-editor.org/rfc/rfc4120.txt)
- [Kerberos terms explanation](https://zeroshell.org/kerberos/Kerberos-definitions/)


devolutions-gateway/src/http/controllers/kdc_proxy.rs
```diff
@@ -0,0 +1,160 @@
...
+    kdc_reply_message: &mut Vec<u8>,
+) -> Result<(), HttpErrorStatus> {
+    let mut buff = Vec::new();
+    let mut len = 0;
+
+    loop {
+        match connection.read_buf(&mut buff).await {
+            Ok(_) => {
+                if len == 0 {
+                    len = u32::from_be_bytes(buff[0..4].to_vec().try_into().unwrap()) as usize;
```
CBenoit, line 135: You should check that `buff.len() >= 4` before

devolutions-gateway/src/http/controllers/kdc_proxy.rs
```diff
@@ -0,0 +1,160 @@
...
+        Ok(Builder::new()
+            .body(kdc_proxy_reply_message.to_vec().unwrap())
+            .status(200))
+    }
+}
+
+async fn read_kdc_reply_message(
+    connection: &mut TcpStream,
+    kdc_reply_message: &mut Vec<u8>,
+) -> Result<(), HttpErrorStatus> {
```
CBenoit, line 127: General comment for this function: I don't think you need an intermediate buffer. Am I wrong?

devolutions-gateway/src/http/controllers/kdc_proxy.rs
```diff
@@ -0,0 +1,160 @@
...
+                .write_all(&kdc_proxy_message.kerb_message.0 .0)
+                .await
+                .map_err(|e| {
+                    slog_scope::error!("{:?}", e);
+                    HttpErrorStatus::internal("Unable to send the message to the KDC server")
+                })?;
+
+            read_kdc_reply_message(&mut connection, &mut kdc_reply_message).await?;
+        } else if scheme == "udp" {
+            let mut buff = vec![0; 1024];
```
CBenoit, line 87: Can you explain why this buffer has this size? Is 1024 the upper bound for a response in UDP mode?
Also, you may use an array instead of a vec: `[0; 1024]`

devolutions-gateway/src/http/controllers/kdc_proxy.rs
```diff
@@ -0,0 +1,160 @@
...
+
+            let port = portpicker::pick_unused_port().ok_or_else(|| HttpErrorStatus::internal("No free ports"))?;
+            let udp_socket = UdpSocket::bind(format!("127.0.0.1:{}", port)).await.map_err(|e| {
```
CBenoit, line 90: Don't use `format!` for that, use `("127.0.0.1", port)` instead

CBenoit (changes requested):
This is looking quite good.

Also +1 for separating KDC asn.1 stuff into a different crate!
About that, I would prefer you move this crate into the `crates/` folder which I plan to use for the non-main and non-application crates in general :slightly_smiling_face: 

Thank you!

devolutions-gateway/src/http/controllers/kdc_proxy.rs
RRRadicalEdward: We need to read a reply from the `KDC` server, then wrap the reply into `KdcProxyMessage`, and send it back to the client. So a buffer is needed. Do you have different a approach?

devolutions-gateway/src/http/controllers/kdc_proxy.rs
RRRadicalEdward: I thought that a ticket could not be more than 1K. I rechecked and the upper bound for a response in UDP mode is 64K. 
I think 2k of a reziable buffer would be enough.

CBenoit: Yeah 64K sounds like a lot. You can bump to 2K or stay at 1K, but add a comment about that maybe :+1:

devolutions-gateway/src/http/controllers/kdc_proxy.rs
CBenoit: Notice in my suggestion there is no risk to read past the size of the message (I think we may eat to much bytes in your version, but I don't know if it's an actual problem)

Cargo.toml
```diff
@@ -13,6 +13,7 @@ members = [
 default-members = [
     "devolutions-gateway",
     "jetsocat",
+    "crates/picky-krb",
```
CBenoit, line 16: It should not be default (it's compiled as part of devolutions-gateway)

CBenoit (changes requested):

devolutions-gateway/src/http/controllers/kdc_proxy.rs
RRRadicalEdward: Done. Thank you.

Cargo.toml
RRRadicalEdward: Done

CBenoit (approved):
Looks good to me! Thank you :tada:

PR #236, opened 2022-02-26
Initial work to share more code between Jetsocat and Gateway

CBenoit:
This removes quite a bunch of legacy code, tidy up the project workspace and hierarchy so it's easier to share and test.

A few additional integration tests and benchmarks for forwarding are also included.

Related to:
- https://devolutions.atlassian.net/browse/DGW-1
- https://devolutions.atlassian.net/browse/DGW-2

It took me more work than expected, and I couldn't do as much as I wanted.
For now, I disabled rendezvous mode. It should be fine, because we don't use it with DVLS so it must be okay to release.
However, this breaks Gateway for people still using Wayk Bastion, though it's not officially supported anymore (cc @fdubois1).
I'll re-add support for rendezvous mode in a future PR.

Commits are independents and may be reviewed separately / partially.


awakecoding (approved):

CBenoit:
Confirmed working in the lab.

PR #241, opened 2022-03-08
Bump version to 2022.1.1

CBenoit:


thenextman (approved):
LGTM. I'd suggest also updating the copyrights [here](https://github.com/Devolutions/devolutions-gateway/blob/38b136d63026997023c7a849612399527f641e5d/powershell/DevolutionsGateway/DevolutionsGateway.psd1#L25) and [here](https://github.com/Devolutions/devolutions-gateway/blob/44483a6c7388e965d946111f998f8dab9e8f6354/package/Linux/gateway/debian/copyright#L6).

CBenoit:
Good catch!

PR #244, opened 2022-03-15
DGW-18: parse query parameters to retrieve token

CBenoit:
Check for JWT in requests with the shape `/route?token=<JWT>`.
This allow configuring KDC client to use `/KdcProxy?token=<JWT>` to identify itself.


awakecoding (approved):
LGTM

PR #257, opened 2022-05-09
picky-krb: New data types and messages that are used in Kerberos authentication.

RRRadicalEdward:
I've added new data types and messages that are used in `Kerberos` authentication:
* Added `Kerberos` constants (key usages, messages ids).
* `KrbResult` is added for easier decoding responses from the `KDC server`.
* `GSS-API `types.

During implementation, I thought: Would it be better to move the  `picky-krb` to the `picky-rs` repository? I mean, types from the `picky-krb` are used in the `devolutions-gateway` and in the `sspi-rs`. Would `picky-rs` be the more suitable place for them?

Notes:
* Pay attention to the `ApplicationTag0` type. It's a workaround for the `ApplicationTag` replacing wrapped `SequenceTag` behavior. 

Useful links:
* [RFC 4120](https://datatracker.ietf.org/doc/html/rfc4120) - Kerberos RFC.
* [RFC 4121](https://datatracker.ietf.org/doc/html/rfc4121) - The Kerberos Version 5 Generic Security Service Application Program Interface (GSS-API) Mechanism.


CBenoit:
> Would it be better to move the picky-krb to the picky-rs repository?

Yeah, I think it would be good.
I’ll start by reviewing this PR, but if you could open a pull request copying `picky-krb` over there afterward, it would be perfect for me. 🙂

crates/picky-krb/src/data_types.rs
```diff
@@ -152,28 +162,247 @@ pub type Ticket = ApplicationTag<TicketInner, 1>;
...
+/// KERB-ERROR-DATA ::= SEQUENCE {
+///     data-type [1] INTEGER,
+///     data-value [2] OCTET STRING OPTIONAL
+/// }
+/// ```
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+pub struct KerbErrorData {
+    data_type: ExplicitContextTag1<IntegerAsn1>,
+    #[serde(default)]
+    data_value: Optional<Option<ExplicitContextTag2<BitStringAsn1>>>,
```
CBenoit, line 184: As I see you’ve been making fields public, I want to make sure that these are intentionally privates

crates/picky-krb/src/data_types.rs
```diff
@@ -152,28 +162,247 @@ pub type Ticket = ApplicationTag<TicketInner, 1>;
...
+
+/// [RFC 4120](https://datatracker.ietf.org/doc/html/rfc4120#section-5.2.7.5)
+///
+/// ```not_rust
+/// ETYPE-INFO2              ::= SEQUENCE SIZE (1..MAX) OF ETYPE-INFO2-ENTRY
+/// ```
+pub type EtypeInfo2 = Asn1SequenceOf<EtypeInfo2Entry>;
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct KrbResult<T>(pub Result<T, KrbError>);
```
CBenoit, line 333: Is it used outside of tests? What is the intended usage pattern?

As a general remark, I was expecting a `Result<T, KrbError>` instead of a newtype pattern.

I’m pretty sure you can implement `Deserialize` on it like done below on it, and get the benefits of using the `?` operator.

crates/picky-krb/src/data_types.rs
```diff
@@ -385,5 +614,257 @@ mod tests {
     }
 
     #[test]
-    fn test_authorization_data() {}
+    fn test_authenticator() {
```
CBenoit, line 617: General comment: avoid prefixing test functions with `test_`
```suggestion
    fn authenticator() {
```
It’s already in a `tests` module + it only appears when issuing a `cargo test` command + there is already a `#[test]` annotation.

I suggest changing that for all the tests.

crates/picky-krb/src/data_types.rs
RRRadicalEdward: А `KDC` response can be some valid data or encoded `KrbError`.  `KrbResult` is used for simplifying the decoding.

RRRadicalEdward: Done

crates/picky-krb/src/data_types.rs
RRRadicalEdward: Done

crates/picky-krb/src/data_types.rs
RRRadicalEdward: I made these fields also public

RRRadicalEdward:
> This looks good to me! Could you open a pull request to add `picky-krb` in `picky` repository instead? We will merge there 🙂

Sure

RRRadicalEdward:
> This looks good to me! Could you open a pull request to add `picky-krb` in `picky` repository instead? We will merge there 🙂

Done(see https://github.com/Devolutions/picky-rs/pull/146).
I will open a new PR in devolutions-gateway that removes the `picky-krb` crate when It's merged in picky-rs.

CBenoit:
Closing since it’s merged in picky-rs repository by https://github.com/Devolutions/picky-rs/pull/146

PR #258, opened 2022-05-09
feat(dgw): `*` in JMUX token for "allow all" rule

CBenoit:
If any host is specified as `*`, the JMUX proxy should be in allow all mode.


awakecoding (approved):
LGTM

PR #266, opened 2022-05-26
feat(jetsocat): process watcher option

CBenoit:
Add `--watch-parent` and `--watch-process` CLI options.
Jetsocat will close itself automatically when watched process is not running anymore.

Issue: ARC-68


awakecoding (approved):
LGTM

PR #267, opened 2022-05-30
feat: migrate logging infratructure to `tracing`

CBenoit:
This also adds support for file rotation and auto-deletion (DGW-25) and bump the version to publish a new version.


awakecoding (approved):
LGTM

PR #269, opened 2022-06-04
feat: HTTP proxy listener

CBenoit:
-  HTTP proxy listener now handles both HTTPS (tunneling) proxy requests and HTTP (regular forwarding).
-  Optional `ap` claim to JMUX tokens. This claim is used to specify the application protocol expected of over JMUX tunnel. This is mostly used to find a good default port when unspecified.
- Duplicate `/jmux` endpoint under `/jet`.


awakecoding (approved):
LGTM

PR #279, opened 2022-06-20
ci: upload artifacts to OneDrive

CBenoit:


.github/workflows/release.yml
```diff
@@ -339,3 +339,68 @@ jobs:
...
+        if: needs.preflight.outputs.dl-strategy == 'cli'
+        shell: pwsh
+        env:
+          GITHUB_TOKEN: ${{ secrets.DEVOLUTIONSBOT_TOKEN }}
+        run: gh run download ${{ needs.preflight.outputs.run }} -n devolutions-gateway --repo $Env:GITHUB_REPOSITORY
+
+      ## workflow_call: The same artifacts persist across the entire run, so the PowerShell/DevolutionsGateway directory will still exist from the CI workflow
+      - name: Manage artifacts
+        shell: pwsh
+        run: Remove-Item (Join-Path PowerShell DevolutionsGateway) -Recurse -ErrorAction Ignore
```
CBenoit, line 382: This was from other jobs, I’m not exactly sure if it’s required in this case but just in case

CBenoit (commented):
This is now ready for review!

.github/workflows/release.yml
```diff
@@ -339,3 +339,68 @@ jobs:
...
+          Move-Item -Path "./windows/x86_64/DevolutionsGateway-x86_64-${version}.msi" -Destination "$destinationFolder/DevolutionsGateway-x86_64-${version}.0.msi"
+
+      - name: Upload to OneDrive
+        uses: ./.github/workflows/upload-onedrive
+        with:
+          client_id: ${{ secrets.ONEDRIVE_AUTOMATION_CLIENT_ID }}
+          client_secret: ${{ secrets.ONEDRIVE_AUTOMATION_CLIENT_SECRET }}          
+          remote: releases
+          destination_path: /Gateway/${{ steps.prepare.outputs.version }}
+          conflict_behavior: replace
```
fdubois1, line 404: Pour une release, je crois que c'est préférable de faire un crash si il y a un conflit plutot que de remplacer le fichier. On ne voudrait pas écraser une vieille release par erreur.

.github/workflows/release.yml
```diff
@@ -339,3 +339,68 @@ jobs:
...
+          files: ${{ steps.prepare.outputs.files-to-upload }}
```
fdubois1, line 405: Le fichier ici possède bien les 4 chiffres (2022.2.2.0) ??  J'essaie de suivre, mais je crois que c'est dans le preflight que ça se passe.

.github/workflows/release.yml
thenextman: It's not needed because in next step you select the files you'll upload, so the presence of the PowerShell/DevolutionsGateway directory doesn't matter. It's harmless to leave it.

.github/workflows/release.yml
thenextman: Yes, the directory and .msi both get a 4-part version (ending in .0)

thenextman:
One small piece of feedback is that the other jobs in release workflow have two checks before running:

`skip-publishing` - If the preflight job sees an existing GitHub release with the same version, it will set this to true and skip remaining jobs (*unless* dry-run is specified)

`dry-run` - allows running the workflow via workflow_dispatch *without* actually generating a release

You might consider following or adapting the logic to prevent surprises. The `dry-run` could simply skip the final upload step.

.github/workflows/release.yml
```diff
@@ -210,7 +210,7 @@ jobs:
     name: GitHub release
     runs-on: ubuntu-18.04
     environment: build-and-publish
-    needs: [preflight]
+    needs: preflight
     if: needs.preflight.outputs.skip-publishing == 'false' || github.event.inputs.dry-run == 'true'
```
thenextman, line 214: The `dry-run` flag isn't respected. You should skip the final "Upload to OneDrive" step if it's true.

.github/workflows/release.yml
CBenoit: I’ll add a comment at the appropriate place to highlight this!

.github/workflows/jetsocat-nuget.yml
```diff
@@ -21,8 +21,16 @@ jobs:
...
+        uses: actions/checkout@v3
+        with:
+          repository: Devolutions/actions
+          ref: master
+          token: ${{ secrets.DEVOLUTIONSBOT_TOKEN }}
+          path: ./.github/workflows
+
       - name: Publish to Artifactory
-        uses: Devolutions/actions/nuget-publish-jfrog@master
+        uses: ./.github/workflows/nuget-publish-jfrog@master
```
dion-gionet, line 33: No need for @master when checking out the actions repo manually
```suggestion
        uses: ./.github/workflows/nuget-publish-jfrog
```

.github/workflows/release.yml
```diff
@@ -210,7 +210,7 @@ jobs:
     name: GitHub release
     runs-on: ubuntu-18.04
     environment: build-and-publish
-    needs: [preflight]
+    needs: preflight
```
dion-gionet, line 213: Are you sure removing the brackets is gonna work?

dion-gionet (changes requested):

.github/workflows/release.yml
CBenoit: Yes, that’s how it’s done at the other places too.
Generally speaking, when a list can be provided, GitHub workflow accepts both a single item or a list. Since we’re using the single item syntax in most place I just unified it.

.github/workflows/jetsocat-nuget.yml
CBenoit: Let me change that 👍

PR #293, opened 2022-08-08
feat(dgw): add general claim `jet_gw_id`

CBenoit:
When this claim is specified, a given token can only be used on a Gateway with the very same ID.


awakecoding (approved):
LGTM

PR #298, opened 2022-08-17
Temporary work around?

MaximeMorin-Devolutions:
Might want to check request headers to return the appropriate values instead.


CBenoit (approved):
Thank you!

PR #299, opened 2022-08-17
Misc improvements following config overhaul

CBenoit:
- Update README to reflect latest changes
- Make sure "additional" configuration options are not removed when pushing config (for instance "ProvisionerPrivateKeyFile" that is used by the PowerShell cmdlet)
- Remove dependency on clap (there is nothing justifying its use anymore)


awakecoding (approved):
LGTM

PR #322, opened 2022-09-14
fix(dgw): remove TLS options from default configuration

CBenoit:
It's not always required and when specified it causes an error on start up because files (obviously) do not exist.


fdubois1 (approved):
LGTM

PR #327, opened 2022-09-20
fix(jmux-proxy): properly cancel proxy task

CBenoit:
Previously, JMUX proxy task wasn't properly shut down because tokio tasks are detached by default (similar to `std::thread::spawn`). This adds a helper wrapper to explicitely specify whether a task should be joined or detached.

Issue: DGW-54


thenextman (approved):
LGTM

PR #331, opened 2022-09-21
feat(dgw): smaller token reuse interval for RDP sessions

CBenoit:
With this change, we do not allow reuse for RDP sessions more than a few seconds following the previous use. The interval is 10 seconds which is expected to give plenty of time to RDP handshake and negotiations. Once this interval is exceeded, we consider the RDP session is fully started and the same token can't be reused anymore.

Two reasons why this is beneficial:

- Security wise: the reuse interval is considerably shortened
- Feature wise: more efficient forced RDP session termination

Regarding the second point: Windows’ mstsc will keep alive the session by re-opening it immediately. Because we allow token reuse in a limited fashion for RDP, as long as the association token is not expired, the terminate action has effectively no visible effect (besides that multiple sessions occurred). Reducing the reuse interval greatly improves the situation.

### DIFF
--- a/devolutions-gateway/src/token.rs
+++ b/devolutions-gateway/src/token.rs
@@ -842,7 +842,7 @@ fn validate_token_impl(
 
                     if now > bucket.get().last_use_timestamp + MAX_REUSE_INTERVAL_SECS {
                         anyhow::bail!(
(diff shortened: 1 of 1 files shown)


awakecoding (approved):
LGTM

PR #333, opened 2022-09-22
ci: feeling lucky with runners

CBenoit:
We have issues with these runners, I'm trying to change and see if it's any better.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -129,7 +129,7 @@ jobs:
           - os: windows
             runner: windows-2019
           - os: macos
-            runner: macos-10.12
+            runner: macos-latest
           - os: linux
             runner: ubuntu-20.04
         exclude:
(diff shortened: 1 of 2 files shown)


richerlariviere (approved):
lucky indeed

.github/workflows/package.yml
```diff
@@ -129,7 +129,7 @@ jobs:
           - os: windows
             runner: windows-2019
           - os: macos
-            runner: macos-10.11
+            runner: macos-10.12
```
dion-gionet, line 132: Macos 10 is deprecated https://github.com/actions/runner-images

dion-gionet (changes requested):

dion-gionet (approved):

.github/workflows/package.yml
CBenoit: I changed to latest (currently 11)

PR #342, opened 2022-10-04
feat (SBOM) Moved sbom generation to the package workflow

sbaker-devolutions:


.github/workflows/package.yml
```diff
@@ -116,6 +117,14 @@ jobs:
           name: changelog
           path: CHANGELOG.md
 
+      - name: Get version
+        id: get-version
+        shell: pwsh
+        run: |
+          $Version = Get-Content VERSION -TotalCount 1
+          echo "::set-output name=version::$Version"
+          echo "::notice::Releasing artifacts for version $Version from run ${{ steps.get-run.outputs.run }}"
```
thenextman, line 126: "Packaging", not Releasing :)

.github/workflows/release.yml
```diff
@@ -347,36 +347,20 @@ jobs:
...
-          client_id: ${{ secrets.ONEDRIVE_AUTOMATION_CLIENT_ID }}
-          client_secret: ${{ secrets.ONEDRIVE_AUTOMATION_CLIENT_SECRET }}
-
-      - name: Set execute permission on cargo-cyclonedx
-        run: |
-          chmod +x /home/runner/.cargo/bin/cargo-cyclonedx
-
-      - name: Generate SBOM
-        uses: devolutions/actions/cyclonedx-rust-generate@master
+          name: devolutions-gateway
```
thenextman, line 356: This isn't right, you want "bom.xml"

.github/workflows/release.yml
```diff
@@ -347,36 +347,20 @@ jobs:
...
-        uses: actions/upload-artifact@v3
-        with:
-          name: bom.xml
-          path: bom.xml
+      - name: Download artifacts (cli)
+        if: needs.preflight.outputs.dl-strategy == 'cli'
+        shell: pwsh
+        env:
+          GITHUB_TOKEN: ${{ secrets.DEVOLUTIONSBOT_TOKEN }}
+        run: gh run download ${{ needs.preflight.outputs.run }} -n devolutions-gateway --repo $Env:GITHUB_REPOSITORY
```
thenextman, line 363: Same, artifact name appears wrong

.github/workflows/package.yml
```diff
@@ -310,4 +319,42 @@ jobs:
...
+
+      - name: Set execute permission on cargo-cyclonedx
+        run: |
+          chmod +x /home/runner/.cargo/bin/cargo-cyclonedx
+
+      - name: Generate SBOM
+        uses: devolutions/actions/cyclonedx-rust-generate@master
+
+      - name: Save SBOM
+        uses: actions/upload-artifact@v2
```
thenextman, line 357: Can be @v3

.github/workflows/release.yml
```diff
@@ -347,36 +347,20 @@ jobs:
...
 
       - name: Upload SBOM
```
thenextman, line 365: You should probably check if: dry-run == false here

.github/workflows/release.yml
sbaker-devolutions: commited the change forgot to push 😅

.github/workflows/release.yml
sbaker-devolutions: same here

.github/workflows/release.yml
thenextman, line 365: You should check if dry-run == false here, otherwise you'll get BOM uploads for dry runs (where we run the workflow to check it but without actually publishing anything)

thenextman (approved):
LGTM

PR #345, opened 2022-10-06
ci: fix tlk.ps1 eating stdout

CBenoit:
stdout of invoked commands was eaten by the PowerShell pipeline stream. Appending `| Out-Host` make it clear that we don’t want that.


thenextman (approved):
Works for me!!

Sylfwood (approved):

PR #348, opened 2022-10-24
Add optional configuration steps to .msi installer

thenextman:
Adds the basic configuration options from Devolutions Server Console directly to the Devolutions Gateway installer. 

Customization via the installer is only available for new installs, and only if the user has PowerShell 5.1 at least. To configure an existing installation using the installer, the user must uninstall first before re-running the installer.

Options are configured in multiple steps with a summary page at the end. Each step is backed by a custom action that will perform basic validation as well as format the values into the necessary PowerShell commands.

Notes:

- The installer now displays the version number on the initial "Welcome" page
- The installer will now query the service startup type of an existing Devolutions Gateway service, and re-apply it after repairing or upgrading. This prevents a bug where that setting could be lost if a user customized it directly post-installation.
- To try and make things clearer, the .wxs now uses "namespaces" (as far as possible) for different IDs (P.* for properties, CA.* for custom actions, etc).
- The .wxs file(s) are also refactored to try and include common parts together. The existing "InitConfig" action is switched to use WiX quiet execution and prevent it popping a console window will it runs.
- The GitHub runners are switched to windows-2022 to allow using the v143 build tools (VS 2022). Note that `msbuild` is now required to be in your PATH to package Devolutions Gateway on Windows.


dion-gionet (approved):

ci/tlk.ps1
```diff
@@ -385,43 +385,54 @@ class TlkRecipe
...
+            & 'candle.exe' '-nologo' $InputFiles $WixArgs "-dPlatform=$TargetArch" | Out-Host
             $OutputFile = "$($this.PackageName)_${Culture}.msi"
         
             if ($Culture -eq 'en-US') {
                 $OutputFile = "$($this.PackageName).msi"
             }
         
-            & 'light.exe' "-v" "-nologo" $ObjectFiles "-cultures:${Culture}" "-loc" "$($this.PackageName)_${Culture}.wxl" `
+            & 'light.exe' "-nologo" $ObjectFiles "-cultures:${Culture}" "-loc" "$($this.PackageName)_${Culture}.wxl" `
                 "-out" $OutputFile $WixArgs "-dPlatform=$TargetArch" "-sice:ICE61" | Out-Host
```
CBenoit, line 431: You changed `"-nologo"` for `'-nologo'` at one place and not the other. I don’t think this is a big deal but just in case there is a deeper reason I’m unaware of

package/Windows/Actions/stdafx.cpp
```diff
@@ -0,0 +1,4 @@
+#include "stdafx.h"
+
+// TODO: reference any additional headers you need in STDAFX.H
+// and not in this file
```
CBenoit, line 4: What is the reason for this? Compilation speed? Is that really a TODO?

awakecoding (approved):

CBenoit (approved):
Nice! 💯

package/Windows/Actions/stdafx.cpp
thenextman: > Compilation speed? 

Probably

> Is that really a TODO?

Probably not 😂 This is actually boilerplate generated by the project template. I'll just remove the todos (I don't think I even looked at this file before!).

ci/tlk.ps1
thenextman: No deeper reason. I iterated this file a bit, and that was probably a change that came in via my editor and didn't really notice.

PR #353, opened 2022-10-27
fix(installer): recursively reset permissions on the ProgramData directory

thenextman:
The installer invokes calcs.exe to reset the permissions on %programdata%\Devolutions\Gateway

However, if a user or tool has previously edited permissions on a file within the directory (and removed the "inheritance" flag), those permissions will not be reset. By adding the `/t` switch to [calcs](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cacls), we can ask it to traverse subdirectories and files:

> /t | Changes ACLs of specified files in the current directory and all subdirectories.

### DIFF
--- a/package/Windows/DevolutionsGateway.wxs
+++ b/package/Windows/DevolutionsGateway.wxs
@@ -92,7 +92,7 @@
     <Property Id="P.PK_CMD" Secure="yes" />
 
     <SetProperty Id="CA.SetProgramDataPermissions" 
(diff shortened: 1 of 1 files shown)


CBenoit:
cc @MathieuMorrissette 
Tagging VM-1061

MathieuMorrissette (approved):
LGTM

CBenoit (approved):

PR #363, opened 2022-12-07
feat(installer): start the Gateway service at install time

thenextman:
If the Gateway service startup type is "Automatic", the installer will now attempt to start the service after configuring it. This is done using a C++ custom action, since it's not possible to use conditions in the WiX `ServiceControl` element.

`StartService` can fail for fairly benign reasons, so failing to start the service does not cause the installation to fail.

The service will not be started if it's startup type is on-demand.

Finally, this behaviour can be overridden by setting `P.DGW.NO_START_SERVICE` via `msiexec`.

Additionally, file pickers inside the installer now always present an "All Files" option. When picking the provisioner public key, .pem is offered before .key to prevent users accidentally selecting the certificate private key in this step.

### DIFF
--- a/package/Windows/Actions/CustomAction.cpp
+++ b/package/Windows/Actions/CustomAction.cpp
@@ -13,6 +13,8 @@ enum Errors
 	InvalidHost = 30000
 };
 
+constexpr auto DG_SERVICE_NAME = L"DevolutionsGateway";
+
 /// <summary>
 /// Write a message to the installer log
 /// </summary>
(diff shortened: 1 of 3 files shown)


package/Windows/Actions/CustomAction.cpp
```diff
@@ -831,7 +941,7 @@ UINT __stdcall ValidateCertificate(MSIHANDLE hInstall)
 
 		for (int i = 0; i < lstrlenW(szCertPass); i++)
 		{
-			szValBuf[i] = L'�';
+			szValBuf[i] = L'�';
```
awakecoding, line 944: this comes off as 'ï¿½' in GitHub - I just assume it's encoded the right way in the source file?

awakecoding (approved):

package/Windows/Actions/CustomAction.cpp
thenextman: It doesn't even show in the local diff. I will double-check my editor didn't mangle this before merging.

PR #372, opened 2023-01-11
fix(pwsh): nil UUID when creating an empty DGatewayConfig

CBenoit:
Without this patch, the nil UUID is used as the "missing" value instead of $null.

Issue: DGW-73


thenextman (approved):
LGTM

PR #373, opened 2023-01-13
fix(installer): fix configuration creation with installer

thenextman:
- On first time installs, always initialize the default config file before applying any PowerShell configuration commands. This ensures that the ID is properly created.
- Properly pass only the hostname (instead of the full URI) to `Set-DGatewayHostname`
- Fix deprecation warnings in GitHub Actions

Issue: DGW-72
Issue: DGW-73


richerlariviere (approved):
lgtm

awakecoding (approved):

PR #376, opened 2023-01-16
fix(installer): avoid unicode char literals

thenextman:
Unicode character literals in source files can be problematic, depending on the editor and encoding. Instead, avoid the issue by masking the character with an asterisk instead of a unicode "bullet".

Issue: DGW-74


CBenoit (approved):
LGTM!

PR #379, opened 2023-01-27
fix(dgw): workaround rustls’ ClientConfig reuse issue

CBenoit:
I’m unsure why reusing the ClientConfig is an issue.

rustls doc says:

> Making one of these can be expensive,
> and should be once per process rather than once per connection.

source: https://docs.rs/rustls/0.20.8/rustls/struct.ClientConfig.html

In our case, this doesn’t work, so I’m creating a new ClientConfig from scratch each time (slower).

I opened an issue on rustls side:
https://github.com/rustls/rustls/issues/1186


thenextman (approved):
LGTM

PR #380, opened 2023-01-27
fix(ci): build the jetsocat nuget in package.yml

thenextman:
Build the jetsocat nuget package as part of the packaging workflow (instead of the old standalone workflow, which just took the latest release from GitHub).

If running the package workflow manually, the version number of the package may be specified; else it defaults to the current date.


CBenoit (approved):
LGTM! Thank you

PR #382, opened 2023-01-30
feat(jetsocat): use rustls-native-certs

CBenoit:
Let rustls use the plateform’s native certificate store.


CBenoit:
cc @jesseGalarneau

PR #384, opened 2023-02-03
ci(jetsocat): add Linux binary to nuget package

thenextman:


awakecoding (approved):
we're still skipping the jetsocat linux arm64 builds, right?

thenextman:
> we're still skipping the jetsocat linux arm64 builds, right?

Affirmative, it's not built

CBenoit (approved):

PR #395, opened 2023-03-04
fix powershell/build.ps1 output path

awakecoding:
use csproj explicitly with dotnet publish -o parameter (output path) since recent .NET SDK makes it an error rather than a warning on a .sln (default), breaking the build


CBenoit (approved):
Thank you!

PR #397, opened 2023-03-13
feat(installer): improved error handling and rollback in Windows installer

thenextman:
PowerShell configuration commands are now executed as custom actions instead of WixSilentExec. Errors are tracked and, if the installer is running with UI, an appropriate error message is shown to the user. For first time installs, if the installation fails, files that may have been created by the configuration process are cleaned up.

PowerShell command output is redirected to a temporary file; in the case of an error we provide the user the path to that file. A general command execution error will display a string error value. 

Custom actions are refactored slightly for consistency and readability:

- Internal functions now only return `void`, `BOOL`, or `HRESULT` where possible. Errors are always handled as `HRESULT` and other results (e.g. Win32 error codes, `LSTATUS`, null references) are converted to `HRESULT` and handled with the different WiX macros (e.g. `ExitOnWin32Error`).
- Consolidate on `WixGetProperty` instead of `MsiGetProperty` and be careful to release the resulting strings (`ReleaseStr`)
- Consolidate on `nullptr` instead of `NULL`

Issue: DGW-76
Issue: DGW-78

### DIFF
--- a/package/Windows/Actions/CustomAction.cpp
+++ b/package/Windows/Actions/CustomAction.cpp
@@ -4,6 +4,11 @@
 enum Errors
 {
 	NoError = 0,
+	CommandExecPublicKeyFailure = 29989,
+	CommandExecCertificateFailure = 29990,
+	CommandExecListenersFailure = 29991,
+	CommandExecAccessUriFailure = 29992,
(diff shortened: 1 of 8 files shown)


CBenoit (approved):
🚀

PR #409, opened 2023-04-24
update /jetc/jrec/push, token claims, file layout

awakecoding:
- use recording.json, recording.webm instead of session.json, session.webm
- use /jet/jrec/push instead of /jet/jrec (/jet/jrec/pull will be added next)
- use fileType=webm query param instead of jet_rft private claim in token
- add jet_rop=push|pull private claim in token (must match /jetc/jrec/push|pull)


CBenoit (approved):
LGTM! Small nits (see below). Feel free to ignore. You need to run `cargo fmt` before we can merge.

PR #412, opened 2023-04-26
initial session recording listing (/jet/jrec/list)

awakecoding:
Here's my first shot at implementing session recording listing using a scope token. It's just half the work, here's what I don't like about it:

- not enough error handling (probably panics if directory is empty)
- scope token cannot be reused, not sure how to make it reusable
- result is a vector of strings, maybe we'd like a vector of objects
- result is not sorted, file creation time sorting would be best

### DIFF
--- a/devolutions-gateway/src/http/controllers/mod.rs
+++ b/devolutions-gateway/src/http/controllers/mod.rs
@@ -6,6 +6,7 @@ pub mod heartbeat;
 pub mod http_bridge;
 pub mod jrl;
 pub mod kdc_proxy;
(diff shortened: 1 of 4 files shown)


CBenoit:
My first shot at implementing session recording listing using a scope token.
It's just half the work, here's what I don't like about it:

- not enough error handling (probably panics if directory is empty)
- scope token cannot be reused, not sure how to make it reusable
- result is a vector of strings, maybe we'd like a vector of objects
- result is not sorted, file creation time sorting would be best

PR #416, opened 2023-04-26
feat(dgw): /jet/jrec/pull/{id}/{filename} endpoint

CBenoit:
Recording files can be fetched using this new endpoint and a JREC token with the `jet_rop` operation set to `pull`.


awakecoding (approved):
LGTM

PR #423, opened 2023-05-03
feat(ci): Build jet-doctor and tokengen in CI

thenextman:
Build **jet-doctor** and **tokengen** in CI for Windows, Linux (x86_64) and macOS (x86_64, arm64 and universal).

Windows and macOS executables are code-signed.

Additionally, the macOS universal executables are marked 'executable' and packaged into a disk image (.dmg) which is further signed and notarized. This is suitable for distribution to a user without needing to bypass Gatekeeper on the client system.

Build time is not optimized (i.e. slow) but it's not expected to run this workflow often.

### DIFF
--- a/.github/workflows/build-tools.yml
+++ b/.github/workflows/build-tools.yml
@@ -1,0 +1,229 @@
+name: Build Tools
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


awakecoding (approved):

CBenoit (approved):
Thank you!

PR #430, opened 2023-05-08
fix(dgw): do not enforce scheme in /jet/fwd

CBenoit:
This was inconsistent with other routes such as /jet/jmux where `dst_hst` will have the `http` or `https` scheme, but this is simply used as a filter policy and Devolutions Gateway will not wrap the stream further into an "`https` protocol layer".

Instead, we rely on the requested uri to choose between plain TCP and TLS wrapping at proxy level (i.e.: /jet/fwd/tcp vs /jet/fwd/tls).


awakecoding (approved):

CBenoit:
Rebased on master.

PR #441, opened 2023-05-18
fix(ci): limit builds on forked PRs, optimize CI workflow

thenextman:
- Although it's possible to check if we're on a fork (and even to check for `dependabot` builds by checking `github.actor`), it's not possible to exclude a job based on a matrix condition. Instead, generate the build matrices dynamically using PowerShell. This allows us to drop the Windows builds in case secrets are not available (e.g. a PR from a fork). A warning will be printed in this case.
- Revert to using `secrets.ARTIFACTORY_READ_TOKEN`. This token is now available for builds from `dependabot`.
- Don't bootstrap conan or OpenSSL when building for non-Windows.
- Drop the `jetsocat` build for the Windows x86 target.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,6 @@ env:
   conan-version: 1.43.4
 
 jobs:
-
   preflight:
     name: preflight
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


CBenoit (approved):
Nice dynamic matrix trick

Nit: instead of `fix(ci):`, there is a commit type `ci`
```
ci: limit builds on forked PRs, optimize CI workflow
```
I use this to put related commits in a "Continuous Integration" section in our changelog:
image

PR #454, opened 2023-06-02
chore: update CHANGELOG.md

CBenoit:
Since this bug was patched in a dependency, the changelog generator did not catch it. Next time I should make it clear in the dependency update commit.


thenextman (approved):
LGTM

PR #458, opened 2023-06-06
fix(jetsocat}: fix MSBuild targets for Windows

thenextman:


awakecoding (approved):
LGTM - I think that's exactly the problem with Windows on ARM!

PR #483, opened 2023-06-27
fix(dgw): truncated payload after PCB reading

CBenoit:
Too many bytes are consumed when PCB string is missing the null-terminator.

Indeed, until now the number of bytes to consume was found by computing the size of the previously decoded PCB when re-encoded. IronRDP will always encode PCB string with a null-terminator (just like mstcs client). This is generally correct, but will cause payload truncation when the received PCB string did not originally contain the null-terminator.

This patch is changing this. The "cursor API" is used instead, and cursor position after reading the PCB can be used to find the number of bytes actually read (even if re-encoding the PDU would give a different result).

Issue: DGW-97

### DIFF
--- a/devolutions-gateway/src/rdp_pcb.rs
+++ b/devolutions-gateway/src/rdp_pcb.rs
@@ -50,9 +50,24 @@ pub fn extract_association_claims(
     }
 }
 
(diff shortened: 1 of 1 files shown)


thenextman (approved):
Nice catch!

PR #499, opened 2023-07-27
chore: Updated and regenerated angular client for angular 16

LucFauvel:


devolutions-gateway/openapi/ts-angular-client/config.json
```diff
@@ -11,10 +11,9 @@
...
   "enumUnknownDefaultCase": false,
   
-  "ngMinVersion": "14.0.0",
-  "ngPackagrMinVersion": "14.0.0",
-  "rxjsMinVersion": "7.4.0",
-  "tsickleMinVersion": "0.46.0",
-  "tsMinVersion": "4.4.0",
-  "zonejsMinVersion": "0.11.4"
+  "ngMinVersion": "16.1.0",
+  "ngPackagrMinVersion": "16.1.0",
```
CBenoit, line 15: Are we depending on features available starting angular "16.1", or "16.0" is enough?
Resulting package.json will have the "^16.0.0" version requirement (>=16.0.0 <17.0.0)
What do you think?

CBenoit (commented):
Thank you!

devolutions-gateway/openapi/ts-angular-client/config.json
LucFauvel: @CBenoit we could that, we probably changed it during our testing because we had issues with peer dependencies

LucFauvel: @CBenoit also wanted to make sure, should I bump the client version to 0.6.0?

CBenoit: Yes, let’s bump the minor number :+1:

CBenoit (approved):
LGTM :rocket:

PR #502, opened 2023-08-03
fix(dgw): error 500 when recording folder is missing

CBenoit:
When listing the recordings, if the recording directory does not exist, it means that there is no recording yet (and the folder will be created later). However, Devolutions Gateway is attempting to read this folder anyway and the HTTP error 500 (Internal Server Error) is returned. This patch fixes this by returning an empty list as appropriate.

Issue: DGW-99


pacmancoder (approved):
LGTM

PR #510, opened 2023-08-15
fix(dgw): typo in TLS forward route

CBenoit:
The name of the endpoint was wrong, and thus /jet/fwd/tls was returning the 404 Not Found status.
Furthermore, the `with_tls` option was not properly set.

Issue: DGW-102


pauldumais (approved):
LGTM

PR #573, opened 2023-10-18
remove a lot of old stuff

awakecoding:
yank container-related stuff, unused farm name and farm member configs, compatibility for obsolete parameter names. good old cleanup


CBenoit (approved):
:broom: :+1:

PR #575, opened 2023-10-19
fix(dgw)!: adjust ngrok options

CBenoit:
Some ngrok options are not making much sense for Devolutions Gateway.

- PROXY protocol: we do not handle PROXY protocol in Devolutions Gateway and instead make use of Conn::peer_addr to find the original client IP.
- Basic Authentication: we have our own way to handle the authentication using Json Web Tokens.
- Schemes: only HTTPS should be used when exposing a Devolutions Gateway on internet.

The `Authtoken` key was also renamed to `AuthToken` for readability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Currently, stable options are:
 
 - `Ngrok`: JSON object describing the ngrok configuration for ingress listeners:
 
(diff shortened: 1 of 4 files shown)


awakecoding (approved):
LGTM

PR #625, opened 2023-12-22
feat(dgw): HTTP API for standalone web application

CBenoit:
## Summary

This adds the HTTP API required by the standalone web application.

- `GET /`: Redirects to `GET /jet/webapp/client`
- `GET /jet/webapp/client/*`: Serves the clientless web application.
- `POST /jet/webapp/app-token`: Signs `WEBAPP` tokens.
- `POST /jet/webapp/session-token`: Signs session tokens.

These new endpoints are exposed only when the web application is enabled. Server will reply with the default 401 Unauthorized response when accessing these endpoints with the web application is disabled.

## Security considerations

In order to prevent password brute forcing, the endpoint `POST /jet/webapp/app-token` is rate-limited per username/IP pair. The maximum number of attempts per minute can be configured and is set to 10 by default. When the rate is exceeded, a WARN-level log is emitted.
The maximum lifetime for the returned WEBAPP token can be configured and is set to 8 hours by default.
Responses returned by both `POST /jet/webapp/app-token` and `POST /jet/webapp/session-token` endpoints include the `Cache-Control` header set to `no-cache, no-store` in order to prevent web browsers from keeping the tokens stored around more than necessary.
~~However, the WEBAPP token is passed to the GET /jet/webapp/client endpoint via query parameters.~~

**EDIT**: we will not pass the WEBAPP token via query params. The login view / page will be included in the single page application and the token only need to live inside a single page.

cc @MathieuMorrissette @pdugre 

## Demonstration (OUTDATED)

[webapp-http-api-demo.webm](https://github.com/Devolutions/devolutions-gateway/assets/3809077/adac14be-6f61-4732-a0df-5af1a49673b3)

All the curl commands and outputs are to be found in the `COOKBOOK.md` file, updated by this PR as well.


pauldumais (approved):
Looks good to me.

PR #630, opened 2024-01-04
feat(dgw): groundwork for network scan feature

irvingoujAtDevolution:
This is the very beginning for Network Scanner. 

1. Add raw socket support for Tokio, as ICMP requires raw socket only and Gateway looks very async oriented (correct me if I'm wrong).
2. Added basic ping 
3. created three crates
 a. network-scanner: contains core logic of network scanning techniques
 b. network-scanner-net: contains network utilities, including raw socket and other network related code
 c. network-scanner-proto: contains protocols struct definition, protocol struct <-> bytes parsers. (ICMP v4 for now, ICMP v6 coming up and more in the future)


crates/network-scanner-net/src/tokio_raw_socket.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+impl AsyncRead for TokioRawSocketStream {
+    fn poll_read(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
+    ) -> std::task::Poll<std::io::Result<()>> {
+        let mut buffer = [0u8; 8096];
```
Irvingouj, line 49: This would work as OS primitives support non-blocking raw socket, as per socket2 documentation.

crates/network-scanner-net/src/tokio_raw_socket.rs
```diff
@@ -0,0 +1,122 @@
...
+        match self.socket.read(&mut buffer) {
+            Ok(size) => {
+                tracing::trace!("Read {} bytes, buffer = {:?}", size, &buffer[..size]);
+                buf.put_slice(&buffer[..size]);
+                std::task::Poll::Ready(Ok(()))
+            }
+            Err(e) => {
+                if e.kind() == std::io::ErrorKind::WouldBlock {
+                    cx.waker().wake_by_ref();
```
irvingoujAtDevolution, line 58: To be improved later, we should use OS primitives to wake up Tokio

crates/network-scanner-net/Cargo.toml
CBenoit, line 1: issue: For all the new `Cargo.toml`, configure keys as in [jmux-proto](https://github.com/Devolutions/devolutions-gateway/blob/bb10fdb37d25d21f18d325dec47d1ab626dd2234/crates/jmux-proto/Cargo.toml): `version = "0.0.0"`, `authors = ["Devolutions Inc. <infos@devolutions.net>"]` and `publish = false`.

crates/network-scanner/src/ping.rs
```diff
@@ -0,0 +1,54 @@
+use std::net::Ipv4Addr;
+
+use network_scanner_net::tokio_raw_socket::TokioRawSocketStream;
+use network_scanner_proto::icmp_v4;
+
+use crate::NetworkScanError;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+type NetowrkScanResult<T> = Result<T, NetworkScanError>;
+
+pub async fn ping(ip: impl Into<Ipv4Addr>) -> NetowrkScanResult<()> {
```
CBenoit, line 10: issue: This will cause the whole function to be inlined and compiled in the consumer crate.

suggestion: You can use the trick [mentioned here](https://github.com/Devolutions/IronRDP/blob/33c66edfc6ddc8fd47d83b3ef4ddf9b1875a4bc2/STYLE.md#avoid-monomorphization). It’s also fine for this function to not be generic at all given that we’ll not use it more than once. At you preference.

crates/network-scanner/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+    IoError(std::io::Error),
+    Other(String),
+    ProtocolError(String),
+}
+
+impl From<std::io::Error> for NetworkScanError {
+    fn from(e: std::io::Error) -> Self {
+        NetworkScanError::IoError(e)
+    }
+}
```
CBenoit, line 14: note: I would be fine with using `anyhow` in this crate, given that everything is getting stringified here. `anyway::Error` is arguably better at keeping the error structured.

crates/network-scanner/src/ping.rs
```diff
@@ -0,0 +1,54 @@
...
+    let mut socket = TokioRawSocketStream::connect(ip)
+        .await
+        .map_err(|e| NetworkScanError::IoError(e))?;
+
+    let time = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map_err(|e| NetworkScanError::Other(format!("Failed to get time: {}", e)))?
```
CBenoit, line 17: nitpick: [Error messages should start by a lower case letter](https://github.com/Devolutions/IronRDP/blob/33c66edfc6ddc8fd47d83b3ef4ddf9b1875a4bc2/STYLE.md#formatting-of-error-messages).

This applies to all the places, including the `.expect("…")`.

crates/network-scanner/src/ping.rs
```diff
@@ -0,0 +1,54 @@
...
+    let packet = icmp_v4::Icmpv4Packet::parse(&buffer[..size])
+        .map_err(|e| NetworkScanError::ProtocolError(format!("Failed to parse ICMP packet: {:?}", e)))?;
+
+    match packet.message {
+        icmp_v4::Icmpv4Message::EchoReply {
+            identifier,
+            sequence,
+            payload,
+        } => {
+            tracing::info!("Received echo reply: {:?} {:?} {:?}", identifier, sequence, payload);
```
CBenoit, line 43: nitpick: [We could record fields and benefit from structured logging](https://github.com/Devolutions/IronRDP/blob/33c66edfc6ddc8fd47d83b3ef4ddf9b1875a4bc2/STYLE.md#logging).

crates/network-scanner/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+network-scanner-net ={ path = "../network-scanner-net" }
+network-scanner-proto ={ path = "../network-scanner-proto" }
+tokio = { version = "1.35.1", features = ["io-util"] }
+tracing = "0.1.40"
+
+[dev-dependencies]
+tokio = { version = "1.35.1", features = ["full"] }
```
CBenoit, line 15: suggestion: Enable only the required features instead of `full`.

crates/network-scanner-proto/src/icmp_v4.rs
```diff
@@ -0,0 +1,387 @@
...
+}
+
+impl Icmpv4Packet {
+    pub fn parse(bytes: impl AsRef<[u8]>) -> Result<Self, PacketParseError> {
+        let mut bytes = bytes.as_ref();
+        let mut packet_len = bytes.len();
+        if bytes.len() < 28 {
+            return Err(PacketParseError::PacketTooSmall(packet_len));
+        }
+        // NOTE(jwall) Because we use raw sockets the first 20 bytes are the IPv4 header.
```
CBenoit, line 243: question: Who / what is jwall? Did you copy paste code from somewhere else?

crates/network-scanner-net/src/tokio_raw_socket.rs
CBenoit, line 1: question: Did you consider wrapping https://crates.io/crates/tokio_socket2? The implementation looks good to me and is very short. We can also consider vendoring it (copying the code with a notice).

CBenoit (commented):
Thank you! I appreciate that the code is separated in multiple crates.

crates/network-scanner-proto/src/icmp_v4.rs
irvingoujAtDevolution: yes, I copy and pasted code from a MIT licensed library, he did not have everything I wantted so I copy and modified it a little bit.

crates/network-scanner-net/src/tokio_raw_socket.rs
irvingoujAtDevolution: looks like it uses 
```
tokio::io::unix::AsyncFd
```
correct me if I'm wrong but I don't think it work on Windows.

crates/network-scanner-proto/src/icmp_v4.rs
irvingoujAtDevolution: https://github.com/zaphar/icmp-socket

crates/network-scanner/src/ping.rs
irvingoujAtDevolution: updated

crates/network-scanner/src/lib.rs

crates/network-scanner/src/ping.rs
irvingoujAtDevolution: updated

crates/network-scanner-proto/src/icmp_v4.rs
```diff
@@ -0,0 +1,400 @@
...
+}
+
+/// Error type returned by parsing the ICMP packets.
+#[derive(Debug)]
+pub enum PacketParseError {
+    /// Not enough bytes to properly parse the packet from.
+    PacketTooSmall(usize),
+    /// An unrecognized ICMP type.
+    UnrecognizedICMPType(u8),
+}
```
CBenoit: suggestion: You could also implement the `std::error::Error` trait. Side bonus: `anyhow` is working better with error types implementing it.

crates/network-scanner-net/Cargo.toml
irvingoujAtDevolution: updated

crates/network-scanner/src/ping.rs
```diff
@@ -0,0 +1,49 @@
...
+use network_scanner_net::tokio_raw_socket::TokioRawSocketStream;
+use network_scanner_proto::icmp_v4;
+
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tracing::trace;
+
+pub async fn ping(ip: Ipv4Addr) -> anyhow::Result<()> {
+    let mut socket = TokioRawSocketStream::connect(ip)
+        .await
+        .map_err(|e| anyhow::anyhow!(e))?;
```
CBenoit, line 12: suggestion: With anyhow, you don’t need `map_err` in such cases.
```rust
let mut socket = TokioRawSocketStream::connect(ip).await?;
```
But you could use [`with_context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.with_context) to add more context to the error:
```rust
let mut socket = TokioRawSocketStream::connect(ip)
    .await
    .with_context(|| format!("couldn’t connect to {ip}"))?;
```

crates/network-scanner/src/ping.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    let time = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map_err(|e| anyhow::anyhow!(e))?
```
CBenoit, line 16: suggestion: Re-introduce the context strings, but using the `anyhow::Context` trait. Same at other places.
```suggestion
        .context("cannot access system time")?
```

crates/network-scanner-proto/src/icmp_v4.rs
CBenoit: Okay. The repository has not been updated for two years now, so it may not be maintained anymore. Otherwise sending a PR to add the missing bits is the way to go.

However, even if the Apache 2 is indeed permissive, you can’t just take the code without the original license like this.
I encourage you to peruse the licensing conditions: https://www.apache.org/licenses/LICENSE-2.0

First, you can’t change the license of the existing code, because Devolutions is not the copyright holder.
In the past, we [directly asked the copyright holder of a project we forked in order to re-license it](https://github.com/KizzyCode/serde_asn1_der-rust/issues/1#issuecomment-559145397).
For this reason, you need to copy the the [LICENSE](https://github.com/zaphar/icmp-socket/blob/master/LICENSE) and put it in `network-scanner-proto`. Also add the [`license` key in the Cargo.toml](https://github.com/zaphar/icmp-socket/blob/02b0714302db717936be987f006edba32cdf0d38/Cargo.toml#L8).
Usually, we would add a line to add ourselves to the file, but I see the original author didn’t modify it to add themselves, so I’m unsure if we should in this case.

Also [add the original author in the `authors` key](https://github.com/zaphar/icmp-socket/blob/02b0714302db717936be987f006edba32cdf0d38/Cargo.toml#L6).

In addition to that, add a README with a notice which makes it clear we include code from the other project:
```
This project includes portions of [icmp-socket](https://github.com/zaphar/icmp-socket), held by Jeremy Wall, 2022.
```

Next time, don’t forget to mention when you copy code from another project. It’s critical to respect the licensing conditions.

crates/network-scanner-net/src/tokio_raw_socket.rs
CBenoit: You are right. Let’s reconsider this later.

CBenoit:
FYI, I changed the scope of the PR to "Devolutions Gateway" (dgw) because this is the component ultimately actually affected by this change (even if it’s not yet integrated). I also made the name a bit shorter.

crates/network-scanner/src/ping.rs
irvingoujAtDevolution: updated

crates/network-scanner-proto/src/icmp_v4.rs
irvingoujAtDevolution: Thank you very much. Will do better next time.

crates/network-scanner/Cargo.toml

crates/network-scanner/src/ping.rs
```diff
@@ -0,0 +1,50 @@
...
+        sequence: 0,
+        payload: time.to_be_bytes().to_vec(),
+    };
+
+    let packet = icmp_v4::Icmpv4Packet::from_message(echo_message);
+    socket.write(&packet.to_bytes(true)).await?;
+    let mut buffer = [0u8; icmp_v4::ICMPV4_MTU];
+    let size = socket.read(&mut buffer).await.map_err(|e| anyhow::anyhow!(e))?;
+
+    let packet = icmp_v4::Icmpv4Packet::parse(&buffer[..size]).map_err(|e| anyhow::anyhow!(e))?;
```
CBenoit, line 31: You forgot to re-introduce the context at a few places.

CBenoit (approved):
LGTM. I left another nitpick comment, but this can be addressed in a future PR. Good job!

PR #635, opened 2024-01-10
refactor(dgw): redesign network-scan-net crate

irvingoujAtDevolution:
It is better to do this now than later, the socket interface should not implement the tokio AsyncRead and AsyncWrite interface, as it does more than just read and write. In this PR, we mainly want to keep the interface align with socket2::Socket, but in async. 

Currently all it does just throw the blocking job to a blocking worker thread, even though this might not be the best performant choice, but will work for now since network scanning is not a long running and frequently invoked task. Will looking for improvement in the future.


crates/network-scanner/src/ping.rs
```diff
@@ -1,34 +1,66 @@
...
+
+    let inited_buf = buffer[..size].as_ref();
 
-    let packet = icmp_v4::Icmpv4Packet::from_message(echo_message);
-    socket.write(&packet.to_bytes(true)).await?;
-    let mut buffer = [0u8; icmp_v4::ICMPV4_MTU];
-    let size = socket.read(&mut buffer).await.map_err(|e| anyhow::anyhow!(e))?;
+    let buffer = inited_buf
+        .iter()
+        .map(|u| unsafe { u.assume_init() })
```
irvingoujAtDevolution, line 59: btw, I could not find anything that for convert MaybeUinit<u8> to regular u8 without using unsafe keyword, if you are aware of better way of doing this, let me know.

crates/network-scanner-net/src/tokio_raw_socket.rs
```diff
@@ -1,122 +1,124 @@
...
+            let (usize, socket_addr) = socket.recv_from(&mut inner_buf)?;
+            rx.blocking_send(inner_buf)
+                .map_err(|_| std::io::Error::new(std::io::ErrorKind::Other, "Channel failed to send"))?;
+            Ok::<(usize, SockAddr), std::io::Error>((usize, socket_addr))
+        })
+        .await??;
+
+        let inner_buf = tx.recv().await.ok_or(std::io::Error::new(
+            std::io::ErrorKind::Other,
+            "Channel failed to receive",
```
CBenoit, line 61: nitpick: Errors should start by lower case.

crates/network-scanner/src/ping.rs
```diff
@@ -1,50 +1,112 @@
...
+            anyhow::bail!("payload does not match for echo reply");
+        } else {
+            Ok(())
         }
+    } else {
+        anyhow::bail!("received non-echo reply");
     }
+}
+
+pub fn block_ping(ip: Ipv4Addr) -> anyhow::Result<()> {
```
CBenoit, line 74: naming: 
```suggestion
pub fn blocking_ping(ip: Ipv4Addr) -> anyhow::Result<()> {
```

CBenoit (commented):
High level remark: in `tokio_raw_socket`, I see we are essentially building an asynchronous API for `TokioRawSocket` by wrapping all the network calls into a `spawn_blocking` call, each time. I also think using `spawn_blocking` is the path of least suffering, and that it should perform well enough given that network scanning will likely happen sporadically. At this point, it seems to me that the code could be much simpler, efficient and not tied to tokio by just writing synchronous code, and only then wrapping into a single top-level `spawn_blocking`. This would result in less heap allocations, less copies, less synchronization (Mutex) and less thread spawning (although I believe tokio is using a thread pool for the blocking threads?).

That is, you could write this:
```rust
pub fn blocking_ping(ip: Ipv4Addr) -> anyhow::Result<()> {
    /* all the blocking networking code */
}

pub async fn ping(ip: Ipv4Addr) -> anyhow::Result<()> {
    tokio::task::spawn_blocking(move || { blocking_ping(ip) }).await?
}
```

Maybe the `network-scanner-net` crate could then be removed completely?

crates/network-scanner/src/ping.rs
CBenoit: There is no safe way to do this without using unsafe code. MaybeUninit is a building block specifically for writing unsafe code.

However, here is a better way to write this code:
```diff
diff --git a/crates/network-scanner/src/ping.rs b/crates/network-scanner/src/ping.rs
index 3044309..6cbbf0e 100644
--- a/crates/network-scanner/src/ping.rs
+++ b/crates/network-scanner/src/ping.rs
@@ -91,14 +91,11 @@ pub fn block_ping(ip: Ipv4Addr) -> anyhow::Result<()> {
         .recv_from(&mut buffer)
         .with_context(|| format!("failed to receive packet from {}", ip))?;

-    let inited_buf = buffer[..size].as_ref();
-
-    let buffer = inited_buf
-        .iter()
-        .map(|u| unsafe { u.assume_init() })
-        .collect::<Vec<u8>>();
+    // SAFETY: `recv_from` returns the number of bytes written into the buffer, so the `size` first
+    // elements are in an initialized state.
+    let filled = unsafe { assume_init(&buffer[..size]) };

-    let packet = icmp_v4::Icmpv4Packet::parse(&buffer[..size]).with_context(|| format!("cannot parse icmp packet"))?;
+    let packet = icmp_v4::Icmpv4Packet::parse(filled).with_context(|| format!("cannot parse icmp packet"))?;

     if let icmp_v4::Icmpv4Message::EchoReply { payload, .. } = packet.message {
         if payload != verifier {
@@ -110,3 +107,15 @@ pub fn block_ping(ip: Ipv4Addr) -> anyhow::Result<()> {
         anyhow::bail!("received non-echo reply");
     }
 }
+
+/// Assume the `buf`fer to be initialised.
+///
+/// # Safety
+///
+/// It is up to the caller to guarantee that the MaybeUninit<T> elements really are in an initialized state.
+/// Calling this when the content is not yet fully initialized causes undefined behavior.
+// TODO: replace with `MaybeUninit::slice_assume_init_ref` once stable.
+// https://github.com/rust-lang/rust/issues/63569
+unsafe fn assume_init(buf: &[MaybeUninit<u8>]) -> &[u8] {
+    &*(buf as *const [MaybeUninit<u8>] as *const [u8])
+}
```

Assumptions made by the unsafe code should be clearly documented. I also vendored a function from the standard library which is not yet stabilized in latest Rust version.

irvingoujAtDevolution:
> High level remark: in `tokio_raw_socket`, I see we are essentially building an asynchronous API for `TokioRawSocket` by wrapping all the network calls into a `spawn_blocking` call, each time. I also think that using `spawn_blocking` is the path of least suffering, and that it should perform well enough given that network scanning will likely happen sporadically. At this point, it seems to me that the code could be much simpler, efficient and not tied to tokio by just writing synchronous code using `socket2`, and only then wrap it into a single top-level `spawn_blocking`. This would result in less heap allocations, less copies, less synchronization (Mutex) and less thread spawning (although I believe tokio is using a thread pool for the blocking threads, so this should be fine in theory).
> 
> That is, you could write this:
> 
> ```rust
> pub fn blocking_ping(ip: Ipv4Addr) -> anyhow::Result<()> {
>     /* all the blocking networking code */
> }
> 
> pub async fn ping(ip: Ipv4Addr) -> anyhow::Result<()> {
>     tokio::task::spawn_blocking(move || { blocking_ping(ip) }).await?
> }
> ```
> 
> And probably remove the whole `network-scanner-net` crate?

I agree that, for the ping function at least, spawn_blocking the blocking ping is much better, however, I would argue that it is still necessary to keep the network-scan-proto crates for following reasons.
1. Subsequent functionalities/protocols may needs an async api with finer control on the socket
2. Network scan by it's natures needs to be done asynchronously, we may have up to hundreds machine to scan in a short period of time, I don't thing we want to spawn up a thread for each machine we are scanning. Though, at this moment, we are using blocking thread to achieve this, but as specified in the comment section of `TokioRawSocket`, we will implement this using OS async primitives.

crates/network-scanner/src/ping.rs
irvingoujAtDevolution: thank you, this looks much better!

irvingoujAtDevolution: updated

crates/network-scanner/src/ping.rs
```diff
@@ -1,50 +1,112 @@
...
 
-use anyhow::Context;
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
-use tracing::trace;
+#[macro_export]
+macro_rules! create_echo_request {
+    () => {{
+        let time = std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .map_err(|e| anyhow::anyhow!(e))?
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/network-scanner-net/src/tokio_raw_socket.rs
```diff
@@ -12,6 +15,14 @@ pub struct TokioRawSocket {
     socket: Arc<Mutex<socket2::Socket>>,
 }
 
+macro_rules! lock_socket {
+    ($socket:expr) => {
+        $socket
+            .lock()
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, format!("failed to lock socket: {}", e)))?
+    };
+}
```
CBenoit, line 24: suggestion: We typically don’t go to such extends to handle poisoning.

```rust
my_mutex.lock().unwrap();
```

Is typically enough. Poisoning will happen when another thread holding the mutex panicked, so it’s generally fine to propagate the panic.

Alternatively, `parking_lot` is providing a Mutex without poisoning (that we already use in Devolutions Gateway).

```
my_mutex.lock();
```

CBenoit (approved):
LGTM. I left a new suggestion, but the code is already good for merging so you can address that in a further PR. Thank you!

CBenoit:
> I agree that, for the ping function at least, spawn_blocking the blocking ping is much better, however, I would argue that it is still necessary to keep the network-scan-proto crates for following reasons.
> 
>     1. Subsequent functionalities/protocols may needs an async api with finer control on the socket

I’ll trust your judgment here. If we’re going to have a proper implementation of `TokioRawSocket` which properly integrate with Tokio scheduler, then it makes sense to develop against this API now.

>     2. Network scan by it's natures needs to be done asynchronously, we may have up to hundreds machine to scan in a short period of time, I don't think we want to spawn up a thread for each machine we are scanning. Though, at this moment, we are using blocking thread to achieve this, but as specified in the comment section of `TokioRawSocket`, we will implement this using OS async primitives in the near feature.

Of course we don’t want a thread per machine to scan. However, there are many fine ways to not do that. For example, we could use [`rayon`](https://crates.io/crates/rayon) and async message passing to stream back the results as they become available. Ideally, we want proper integration with tokio so that we can fully benefit from the async model. But this takes a proper integration with tokio (instead of `spawn_blocking`). As discussed, this can be seen as a future improvement.

PR #636, opened 2024-01-10
feat(dgw) : network scan ip v4 broadcasting

irvingoujAtDevolution:
Supporting network scan ipv4 boardcasting, this is a draft, waiting for https://github.com/Devolutions/devolutions-gateway/pull/635 to be merged first

### DIFF
--- a/crates/network-scanner-net/Cargo.toml
+++ b/crates/network-scanner-net/Cargo.toml
@@ -11,7 +11,7 @@ anyhow = "1.0.79"
 crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
 futures = "0.3.30"
 parking_lot = "0.12.1"
-polling = "3.3.2"
+polling = {git = "https://github.com/irvingoujAtDevolution/polling.git"}
 socket2 = { version = "0.5.5", features = ["all"] }
 thiserror = "1.0.56"
 tokio-stream = "0.1.14"
--- a/crates/network-scanner-net/Cargo.toml
+++ b/crates/network-scanner-net/Cargo.toml
@@ -9,10 +9,12 @@ publish = false
 [dependencies]
 anyhow = "1.0.79"
 crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
+futures = "0.3.30"
 parking_lot = "0.12.1"
 polling = "3.3.2"
 socket2 = { version = "0.5.5", features = ["all"] }
 thiserror = "1.0.56"
+tokio-stream = "0.1.14"
 tracing = "0.1.40"
 
 [dev-dependencies]
--- a/crates/network-scanner/Cargo.toml
+++ b/crates/network-scanner/Cargo.toml
@@ -12,7 +12,7 @@ futures = { version = "0.3.30", features = ["io-compat"] }
 network-scanner-net ={ path = "../network-scanner-net" }
 network-scanner-proto ={ path = "../network-scanner-proto" }
 socket2 = "0.5.5"
-tokio = { version = "1.35.1", features = ["io-util", "rt"] }
+tokio = { version = "1.35.1", features = ["rt"] }
 tracing = "0.1.40"
 
 [dev-dependencies]
--- a/crates/network-scanner/Cargo.toml
+++ b/crates/network-scanner/Cargo.toml
@@ -8,11 +8,11 @@ publish = false
 
 [dependencies]
 anyhow = "1.0.79"
-futures = { version = "0.3.30", features = ["io-compat"] }
+futures = { version = "0.3.30" }
 network-scanner-net ={ path = "../network-scanner-net" }
 network-scanner-proto ={ path = "../network-scanner-proto" }
 socket2 = "0.5.5"
-tokio = { version = "1.35.1", features = ["rt"] }
+tokio = { version = "1.35.1" }
 tracing = "0.1.40"
 
 [dev-dependencies]
--- a/crates/network-scanner/Cargo.toml
+++ b/crates/network-scanner/Cargo.toml
@@ -8,10 +8,11 @@ publish = false
 
 [dependencies]
 anyhow = "1.0.79"
+futures = { version = "0.3.30", features = ["io-compat"] }
 network-scanner-net ={ path = "../network-scanner-net" }
 network-scanner-proto ={ path = "../network-scanner-proto" }
 socket2 = "0.5.5"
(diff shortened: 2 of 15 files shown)


crates/network-scanner-net/examples/block_raw.rs
CBenoit, line 1: suggestion: This example does use any API of any of our crate so we can remove it.

crates/network-scanner-net/examples/send_all.rs
CBenoit, line 1: suggestion: This example is only exercising the `network-scanner-proto` crate, so it should be part of `network-scanner-proto` instead.

crates/network-scanner-net/Cargo.toml
```diff
@@ -9,6 +9,10 @@ publish = false
...
 [dependencies]
 anyhow = "1.0.79"
+etherparse = "0.13.0"
 socket2 = { version = "0.5.5", features = ["all"] }
-tokio = { version = "1.35.1", features = ["io-util","rt"] }
+tokio = { version = "1.35.1", features = ["io-util","rt","sync"] }
 tracing = "0.1.40"
+
+[dev-dependencies]
+network-scanner-proto = { path = "../network-scanner-proto" }
```
CBenoit, line 18: Following the above, you can remove the dev dependency as well.

crates/network-scanner/examples/block_boardcast.rs
```diff
@@ -0,0 +1,36 @@
...
+    for result in iterator {
+        match result {
+            Ok(ping_res) => {
+                println!("received result {:?}", &ping_res);
+            }
+            Err(e) => match e.downcast_ref::<std::io::Error>() {
+                Some(e) => match e.kind() {
+                    std::io::ErrorKind::TimedOut => {
+                        tracing::info!("Timed out");
+                        exit(0);
```
CBenoit, line 23: issue: `exit(0)` is not running any destructor and is generally not necessary.

> Note that because this function never returns, and that it terminates the process, no destructors on the current stack or any other thread’s stack will be run. If a clean shutdown is needed it is recommended to only call this function at a known point where there are no more destructors left to run; or, preferably, simply return a type implementing [Termination](https://doc.rust-lang.org/std/process/trait.Termination.html) (such as [ExitCode](https://doc.rust-lang.org/std/process/struct.ExitCode.html) or Result) from the main function and avoid this function altogether

suggestion: Use `return Ok(())` if you want to exit with a success status code. Alternatively, you used `break` at the other places, which have the same behavior. Yet another alternative is to just return the error instead of handling it this way. I don’t really see any special logic happening other than breaking out of the loop and return `Ok(())` anyway. Maybe you didn’t mean to break the loop at the other places? (In order to continue processing of the remaining results.)

As an aside, a timeout can probably be considered an error too. (`Err` instead of `Ok`.)

crates/network-scanner/examples/boardcast.rs
```diff
@@ -0,0 +1,35 @@
...
+                tracing::info!("received result {:?}", &res)
+            }
+            Err(e) => {
+                if let Some(e) = e.downcast_ref::<std::io::Error>() {
+                    // if is timeout, say timeout then break
+                    if let std::io::ErrorKind::TimedOut = e.kind() {
+                        tracing::info!("timed out");
+                        exit(0)
+                    }
+                }
```
CBenoit, line 28: question: Do we need to specifically handle the timeout case?

crates/network-scanner/Cargo.toml
```diff
@@ -8,9 +8,12 @@ publish = false
 
 [dependencies]
 anyhow = "1.0.79"
+get_if_addrs = "0.5.3"
 network-scanner-net ={ path = "../network-scanner-net" }
 network-scanner-proto ={ path = "../network-scanner-proto" }
+socket2 = "0.5.5"
 tokio = { version = "1.35.1", features = ["io-util"] }
+tokio-stream = "0.1.14"
```
CBenoit, line 16: question: I don’t see any utility provided by this crate used in the new code. Am I missing something?

suggestion: Use `futures-core` instead.

crates/network-scanner/Cargo.toml
```diff
@@ -8,9 +8,12 @@ publish = false
 
 [dependencies]
 anyhow = "1.0.79"
+get_if_addrs = "0.5.3"
```
CBenoit, line 11: issue: This crate is unmaintained since 2019 (the repository has been archived).

question: Did you consider this? https://lib.rs/crates/if-addrs
This is maintained fork.

CBenoit (commented):
Partial review. The diff also contains changes from the other PR so it’s a bit harder. I’ll wait for you to fix conflicts etc.
General comment: try to run [`typos`](https://crates.io/crates/typos-cli)

crates/network-scanner-net/examples/block_raw.rs
irvingoujAtDevolution: Sorry, this was meant to keep as a draft, not really ready for review yet, I was waiting for the other PR to be merged so I can rebase and clean up the code on this branch. Still, thanks for your review, I will fix them.

crates/network-scanner/src/lib.rs
```diff
@@ -1 +1,2 @@
+pub mod boardcast;
```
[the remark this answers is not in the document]

crates/network-scanner-net/examples/send_all.rs
irvingoujAtDevolution: updated

crates/network-scanner/examples/block_boardcast.rs
irvingoujAtDevolution: updated.

As of the timeout error, it is important to notice that timeout is different from other error. When searching the network, we do not know exactly how long it takes for some computer to respond back to us, so it is necessary to set a timeout for the operation. if the error kind is timeout, that means this is operation is finished as expected (there is no other way to finish it).

crates/network-scanner/examples/boardcast.rs
irvingoujAtDevolution: yes, please see the above comments

irvingoujAtDevolution:
Not Ready for Review!

irvingoujAtDevolution:
Blocked By https://github.com/Devolutions/devolutions-gateway/pull/665

CBenoit:
Converted to draft until #665 is merged

CBenoit:
@irvingoujAtDevolution Can you rebase on top of master? I recommend you to squash your commits before doing so, to avoid conflict hell.

PR #638, opened 2024-01-11
feat(installer): new Windows installer built using WixSharp

thenextman:
A complete rewrite of the Devolutions Gateway Windows installer using [wixsharp](https://github.com/oleg-shilo/wixsharp).

The following key points are drivers for the change:

- A far easier to understand programming model than directly developing WiX (wxs) source code. The wixsharp script is transpiled to WiX source code and uses the standard WiX toolchain to build the final msi so there is no lock-in and standard WiX solutions are easy to integrate.
- Custom actions are developed in-line (rather than as a submodule) and more closely integrated into the installation script.
- Using managed code throughout is much more expressive than the previous XML/C++ mix, and helps prevent many common classes of error (for example; installer custom properties are now defined in a T4 template and can be mostly used in a type-safe way; common mistakes can be caught at compilation time rather than through obscure MSI runtime errors).
- wixsharp provides a built in abstraction to use WinForms for custom UI which clearly improves on the developer and runtime experience of Windows Installer UI authored using WiX.

 ### Notes and Quirks
- Custom actions still lean heavily on Win32 (now via p/invoke) for two reasons:
1. Reuse of existing, tested functionality
2. `advapi32.dll` provides a more comprehensive API for controlling Windows Services than .NET Framework
- Unavoidably Windows Installer abstractions leak up into the wixsharp code (as they do with a WiX codebase) and a good understanding of the technology is helpful (understanding execution sequences, property availability, conditionals, etc). Sadly this is not simply "an installer written in C#".
- There exists a bug in Windows Installer itself when using external UI whereby, when executing the install as a non-admin user, the UAC prompt is _sometimes_ opened minimized in the taskbar (msiexec does not pass the parent window handle to `ShellExecute`). To a non-attentive user this a simply appear as though the installer is hung. There exists no good workaround for this issue. As a remediation, if the install does not proceed for 4 seconds, the user is prompted to check the taskbar:

Screenshot 2024-01-11 at 13 33 37

- Configuration commands are still executed by shelling out to PowerShell, this was a direct port from the previous installer. A future improvement would be to execute the PowerShell in-script using a PowerShell runtime.

### TODO

- Build system integration. The installer can be built with `msbuild` but it remains to replace the existing project in tlk.ps1 and update the workflows appropriately.
- Some missing localizations. Some text was not localized in the old installer (mainly due to logic having to be built in C++ custom actions, where the installer database is not easy to access - such functionality has now been elevated to the UI).
- Now that we can define a UI with a properly flowing layout, we should provide some more detailed descriptions of the various configuration options exposed by the installer.
- I note that the installer has not been updated to allow configuring a certificate from the Windows certificate store. The user must provide a file.


CBenoit (approved):
Excellent work! Thank you!!

PR #640, opened 2024-01-15
feat(installer): add Windows sytem certificate support to managed installer

thenextman:
Adds support for using Windows certificate store parameters in the managed Windows installer.

If certificate selection is required (i.e. an https listener is being created), the user will now be prompted with "External" and "System" options for certificate selection. "External" functions as before; "System" allows the user to select the store location and name, and search for certificates by thumbprint or subject name.

The user must select _one_ certificate in this mode. If multiple certificates match their search, they will be forced to choose one. The certificate mode, store location, name and certificate subject will be configured in the Gateway .cfg file.

https://github.com/Devolutions/devolutions-gateway/assets/2573400/c342785d-f221-43e8-bec6-7307458a3530

Notes:

- The "Search" screen was easier to implement than a generic certificate browser, hopefully it's use is intuitive enough.
- The `CurrentService` store location isn't available; there's no .NET API to interact with this certificate store. If this is a requirement then the user should configure the certificate using PowerShell. I'm not sure if this is a niche use case (why wouldn't the user just put the certificate in the local machine store?).
- Only the .NET built-in [store names](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.storename?view=net-8.0) are exposed. We could allow the user to input their own store name?
- Additionally, I'm trusting that these values are what the Gateway expects (I assume so as, for example "StoreName.My" maps to the default "My" value in the Gateway readme)
- Finally, the certificate subject name in .NET is a full distinguished name. I don't think this is what the configuration file wants, so I'm calling [X509Certificate2.GetNameInfo](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.getnameinfo?view=net-8.0) with a value of `X509NameType.SimpleName`. It _appears_ to be correct.

This is an incremental PR: the managed installer is not currently published. Several localizations are missing and the summary install page is temporarily disabled.

### DIFF
--- a/package/WindowsManaged/.editorconfig
+++ b/package/WindowsManaged/.editorconfig
@@ -2,3 +2,6 @@
 [*]
 end_of_line = lf
 insert_final_newline = true
+
+[*.tt]
+end_of_line = crlf
--- a/package/WindowsManaged/Actions/CustomActions.cs
+++ b/package/WindowsManaged/Actions/CustomActions.cs
@@ -58,21 +58,35 @@ namespace DevolutionsGateway.Actions
 
             try
             {
-                if (string.IsNullOrEmpty(session.Get(GatewayProperties._CertificatePassword)))
+                Constants.CertificateMode mode = session.Get(GatewayProperties._CertificateMode);
+
+                if (mode == Constants.CertificateMode.External)
                 {
-                    command = string.Format(
-                        Constants.ImportDGatewayCertificateWithPrivateKeyCommandFormat,
-                        session.Get(GatewayProperties._CertificateFile),
-                        session.Get(GatewayProperties._CertificatePrivateKeyFile));
+                    if (string.IsNullOrEmpty(session.Get(GatewayProperties._CertificatePassword)))
+                    {
+                        command = string.Format(
+                            Constants.ImportDGatewayCertificateWithPrivateKeyCommandFormat,
+                            session.Get(GatewayProperties._CertificateFile),
+                            session.Get(GatewayProperties._CertificatePrivateKeyFile));
+                    }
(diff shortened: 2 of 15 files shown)


awakecoding:
For CurrentService: no worries, leave that one out of the UI. We've added it for advanced use cases only, so editing the json is file.

For custom store names, we can wait until a customer asks for it, and they may as well just edit the json. No need to bloat the installer UI when we expect the vast majority of users to go with built-in store names.

The configuration options in the JSON were actually mapped to Kestrel options, [as discussed here](https://github.com/Devolutions/devolutions-gateway/pull/576#issuecomment-1772675506). You can find the Kestrel certificate loader logic [in the code here](https://github.com/dotnet/aspnetcore/blob/main/src/Servers/Kestrel/Core/src/CertificateLoader.cs).

I don't recall adding an option to select the certificate by thumbprint though, since Kestrel didn't have one?

thenextman:
Thanks!

> I don't recall adding an option to select the certificate by thumbprint though, since Kestrel didn't have one?

No: sorry for the confusion. The thumbprint / subject name just allows the user to _search_ for the certificate they want to use. They must search for and select a _single_ certificate, and the information encoded to the config is the location, store and subject name of the selected certificate.

(An additional "Thumbprint" property was added to the installer, but this is just for state management in the UI - if you go forward, and then come back to this page, I want to keep the certificate selection you made).

Hope it's clear.

thenextman:
For the store name and subject name, everything looks good after checking the kestrel source.

CBenoit (approved):
Excellent job!

PR #641, opened 2024-01-15
feat(dgw): initial web application UI for standalone mode

kristahouse:
First WIP of Standalone Gateway UI.


awakecoding:
@kristahouse are all the generated files excluded from git? I'm not familiar enough with those files types to tell, and there are a lot of them. Otherwise, just from trying it out locally, it looks good enough for a first merge.

awakecoding:
@thenextman @bcortier-devolutions

awakecoding (approved):

CBenoit (commented):
Small remark: should we move up all the files under `webapp/web/` to be in `webapp` instead?
I’m not sure we need to nest much more than that. What do you think?

PS: Renaming the PR to match conventional commit specification.

CBenoit:
Merging now, we can make the other adjustments later.

PR #650, opened 2024-01-16
ci: build managed installer in workflows

thenextman:
Integrate the managed wixsharp Windows installer into the build system.

> [!IMPORTANT] 
> The previous WiX installer is still built. In CI and packaging it gets a "-legacy" suffix on the .msi filename. In release, only the legacy installer is currently released - this is a safety hatch in case we need to make a new release between now and 2024.1. Once we are ready to commit; the legacy installer can be removed from the build system.
> These changes add a fair amount of complexity to the build system, which can be removed again once we're committed (2024.1)

In Dev and CI; the managed installer is built using the wixsharp project and msbuild. This is convenient and produces a multi-language MSI in a single build step.

In Packaging, it's necessary to code sign the new runtime components of the installer:

1. tlk.ps1 takes a new "PackageOption" parameter to control what we want to do
2. We use the `generate` option: this builds the installer's managed code, WiX source code, and generates a batch file that invokes standard WiX tooling to build the final MSI.
2(a). This is done twice; for both the en-US and fr-FR localization
3. We call AzureSignTool to code sign the new managed runtime components
4. We use the `assemble` option to build the final MSI
4(a). The en-US MSI is built
4(b). The fr-FR MSI is built
4(c). We create a transform (.mst) between the en-US and fr-FR MSIs
4(d). The transform is embedded in the en-US MSI, and it's language list is updated
6. The MSI itself is code-signed as normal

As regards (3); the managed devolutionsgateway.exe (custom UI and actions) is code signed successfully but the other assemblies (WixSharp DLLs) are not - AzureSignTool claims they are already signed (but they aren't). For now, this appears "good enough" and the new managed MSI no longer triggers SmartScreen on my workstation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -436,8 +436,15 @@ jobs:
             $Env:DGATEWAY_PSMODULE_PATH = "${{ steps.load-variables.outputs.dgateway-psmodule-output-path }}"
             $Env:DGATEWAY_WEBCLIENT_PATH = Join-Path "webapp" "client"
           }
-
+            
           ./ci/tlk.ps1 package -Platform ${{ matrix.os }} -Architecture ${{ matrix.arch }} -CargoProfile ${{ needs.preflight.outputs.rust-profile }}
+
+          if ($Env:RUNNER_OS -eq "Windows") {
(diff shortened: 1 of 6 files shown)


awakecoding (approved):

CBenoit (approved):
Thank you!

PR #651, opened 2024-01-17
feat(dgw): store users list in a users.txt file

CBenoit:
This changes the `Users` option into `UsersPath` for specifying a path to a users.txt file.
By default, the path is `%ProgrameData%/Devolutions/Gateway/users.txt`.

For each line such as `<user>:<hash>`:

* `<user>`: The name of the user.
* `<hash>`: Hash of the password in the PHC string format.

Blank lines and lines starting by `#` are ignored.

Issue: DGW-127

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,16 +203,18 @@ Stable options are:
     * **LoginLimitRate** (_Integer_): The maximum number of login requests for a given username/IP pair
         over a minute (default is `10`).
 
-    * **Users** (_Array_): A list of users authorized to access the web application when using the
-        `Custom` authentication method.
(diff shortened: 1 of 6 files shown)


thenextman (approved):

awakecoding:
Maybe call it UsersFile or UsersFilePath instead of UsersPath?

CBenoit:
> Maybe call it UsersFile or UsersFilePath instead of UsersPath?

Good call. I opened a PR for that: https://github.com/Devolutions/devolutions-gateway/pull/652

PR #655, opened 2024-01-18
Experimental: implementation of asyn io for socket2

irvingoujAtDevolution:
This is a branch of experiment of non-blocking implementation for `TokioAsyncSocket`. I have emailed the maintainer of Mio and socket2, @Thomasdezeeuw, they have decided not to integrate socket2 into Mio hence it cannot be integrated into Tokio. A fork of Mio would not work since tokio did not expose handle for the event-loop.Thus, We would not be able to integrate socket2 into Tokio without fork both Mio and Tokio, which is a lot of overhead. 

Therefore, I created a mini async-io event-loop that constantly polling on a second thread. The API will not change much except for the creation of the socket.

@CBenoit Please take a look and I appreciate any feedback! (This does not necessarily needs to be merged, for discussion only!)


crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,150 @@
...
+        let id = self.runtime.lock().await.add_socket(&socket)?;
+        Ok(TokioRawSocket::from_socket(socket, id, self.clone())?)
+    }
+
+    pub(crate) fn awake(&self, socket: Arc<Socket>, id: Event, waker: std::task::Waker) -> anyhow::Result<()> {
+        let runtime = self.runtime.clone();
+        tokio::task::spawn(async move {
+            let runtime = runtime.lock().await;
+            runtime.awake_when_ready(socket, id, waker).await
+        });
```
CBenoit, line 146: suggestion: Do not use the Mutex from tokio, but the one from parking_lot instead. I suspect we could remove the dependency on `tokio` from the crate entirely.

crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,150 @@
...
+use polling::{Event, Events};
+use socket2::Socket;
+use tokio::sync::{Mutex};
+
+use crate::tokio_raw_socket::TokioRawSocket;
+
+#[derive(Debug)]
+pub struct AsyncIoRuntime {
+    poller: Arc<polling::Poller>,
+    socket_id: usize,
```
CBenoit, line 16: suggestion: Use an `AtomicUsize` so that you don’t need to take `&mut AsyncIoRuntime`.

crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,150 @@
...
+    loop_handle: Option<std::thread::JoinHandle<()>>,
```
CBenoit, line 17: question: What is the purpose of having this? If we want a way to shutdown the runtime, we need message passing to tell the polling thread to stop looping. It would be better to support that properly. A simple approach would be to have an `AtomicBool` that we check after the call to `wait`, and use [`notify`](https://docs.rs/polling/latest/polling/struct.Poller.html#method.notify) to make it return immediately.

note: Can be changed to `std::thread::JoinHandle<()>`, without the option if we spawn the thread eagerly at creation time.

crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,150 @@
...
+
+use polling::{Event, Events};
+use socket2::Socket;
+use tokio::sync::{Mutex};
+
+use crate::tokio_raw_socket::TokioRawSocket;
+
+#[derive(Debug)]
+pub struct AsyncIoRuntime {
+    poller: Arc<polling::Poller>,
```
CBenoit, line 15: suggestion: Could be `poller: polling::Poller` if we can work with `Arc<AsyncIoRuntime>`.

crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,150 @@
...
+    socket_id: usize,
+    loop_handle: Option<std::thread::JoinHandle<()>>,
+    waiting_list: Arc<Mutex<Vec<SocketWaiter>>>,
+}
+
+#[derive(Debug)]
+struct SocketWaiter {
+    id: usize,
+    waker: std::task::Waker,
+}
```
CBenoit, line 25: suggestion: Use `Arc<Mutex<HashMap<usize, std::task::Waker>>>`, which seems to be the appropriate data structure for all the operations here, and remove this struct.

CBenoit (commented):
I believe you got the right direction for this. :+1: 

suggestion: We can probably simplify the API a little bit by eagerly spawning the polling thread instead of having to call `start_loop`.

suggestion:  Use interior mutability (`Mutex` / atomics) so that `AsyncIoHandle` can be replaced by `Arc<AsyncIoRuntime>`.
You can see [this pattern in tokio too](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html). (The whole API of `Runtime` is taking `&mut Self`).

suggestion: Maybe rename to `Socket2Runtime`, just to make it clear that the only purpose of this runtime is to provide a direct, async interface to `socket2`, and nothing else. (We can’t use this runtime to replace tokio.)

Taking into account all the comments, here is how the runtime would be defined:

```rust
pub struct Socket2Runtime {
    poller: polling::Poller,
    next_socket_id: AtomicUsize,
    is_terminated: AtomicBool,
    waiting_list: Arc<Mutex<HashMap<usize, std::task::Waker>>>,
}

impl Socket2Runtime {
    pub new() -> Arc<Self> {
        let runtime = Self {
            …
        };

        let runtime = Arc::new(runtime);

        // Spawn the polling thread…

        runtime
    }
}
```

crates/network-scanner-net/src/tokio_raw_socket.rs
```diff
@@ -109,3 +89,44 @@ impl TokioRawSocket {
...
+                    let socket = self.socket.clone();
+                    let id = self.id;
+                    if self
+                        .handle
+                        .awake(socket.clone(), Event::readable(id), cx.waker().clone())
+                        .is_err()
+                    {
+                        tracing::warn!("failed to awake socket");
+                        cx.waker().wake_by_ref();
+                    }
```
CBenoit, line 123: issue: If we fail at registering the interest for this socket, then I think we should return an error instead of starting a busy loop and slowing down the whole application.

irvingoujAtDevolution:
> I believe you got the right direction for this. 👍
> 
> suggestion: We can probably simplify the API a little bit by eagerly spawning the polling thread instead of having to call `start_loop`.
> 
> suggestion: Use interior mutability (`Mutex` / atomics) so that `AsyncIoHandle` can be replaced by `Arc<AsyncIoRuntime>`. You can see [this pattern in tokio too](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html). (The whole API of `Runtime` is taking `&mut Self`).
> 
> suggestion: Maybe rename to `Socket2Runtime`, just to make it clear that the only purpose of this runtime is to provide a direct, async interface to `socket2`, and nothing else. (We can’t use this runtime to replace tokio.)
> 
> Taking into account all the comments, here is how the runtime would be defined:
> 
> ```rust
> pub struct Socket2Runtime {
>     poller: polling::Poller,
>     next_socket_id: AtomicUsize,
>     is_terminated: AtomicBool,
>     waiting_list: Arc<Mutex<HashMap<usize, std::task::Waker>>>,
> }
> 
> impl Socket2Runtime {
>     pub new() -> Arc<Self> {
>         let runtime = Self {
>             …
>         };
> 
>         let runtime = Arc::new(runtime);
> 
>         // Spawn the polling thread…
> 
>         runtime
>     }
> }
> ```

Updated!

crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,150 @@
...
+        domain: socket2::Domain,
+        ty: socket2::Type,
+        protocol: Option<socket2::Protocol>,
+    ) -> anyhow::Result<TokioRawSocket> {
+        let socket = socket2::Socket::new(domain, ty, protocol)?;
+        let id = self.runtime.lock().await.add_socket(&socket)?;
+        Ok(TokioRawSocket::from_socket(socket, id, self.clone())?)
+    }
+
+    pub(crate) fn awake(&self, socket: Arc<Socket>, id: Event, waker: std::task::Waker) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/network-scanner-net/src/async_io.rs
irvingoujAtDevolution: udpated

crates/network-scanner-net/src/async_io.rs
irvingoujAtDevolution: updated

crates/network-scanner-net/src/async_io.rs
irvingoujAtDevolution: updated

crates/network-scanner-net/src/async_io.rs
irvingoujAtDevolution: updated

crates/network-scanner-net/src/lib.rs
```diff
@@ -1 +1,11 @@
-pub mod tokio_raw_socket;
+pub mod async_io;
+pub mod async_raw_socket;
```
CBenoit, line 2: naming: I would suggest this:
```suggestion
pub mod runtime;
pub mod socket;
```

CBenoit (commented):
Sorry, I’ll review this again a bit later.

crates/network-scanner-net/Cargo.toml
```diff
@@ -9,6 +9,15 @@ publish = false
...
+crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
+futures = "0.3.30"
+parking_lot = "0.12.1"
+polling = "3.3.2"
+
 socket2 = { version = "0.5.5", features = ["all"] }
-tokio = { version = "1.35.1", features = ["io-util","rt","sync"] }
+thiserror = "1.0.56"
 tracing = "0.1.40"
+tracing-subscriber = "0.3.18"
```
CBenoit, line 20: issue: `tracing-subscriber` should not be a dependency of this crate. Maybe a dev-dependency?

crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,152 @@
...
+use socket2::Socket;
+
+use crate::async_raw_socket::AsyncRawSocket;
+
+#[derive(Debug)]
+pub struct Socket2Runtime {
+    poller: polling::Poller,
+    next_socket_id: AtomicUsize,
+    is_terminated: AtomicBool,
+    map: Arc<Mutex<HashMap<usize, std::task::Waker>>>,
```
CBenoit, line 23: suggestion: Maybe this doesn’t need to be wrapped in an `Arc`

CBenoit (commented):
Nice! This is promising.

crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,152 @@
...
+                    self.poller
+                        .modify(socket, event)
+                        .map_err(|e| tracing::warn!(error = ?e, "Event registration failed"))
+                        .ok(); // cannot handle this error
+                    tracing::trace!("event registered successfully");
+                }
+                tracing::warn!("register loop terminated")
+            })?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
irvingoujAtDevolution: yes, acquiring the lock to the map is potentially blocking, Tokio will crash.

crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,152 @@
...
+        // event loop will terminate after this
+        // register loop will terminate becase of sender is dropped after this.
+    }
+}
+
+impl Socket2Runtime {
+    pub fn new() -> anyhow::Result<Arc<Self>> {
+        let poller = polling::Poller::new()?;
+        //unbounded channel performance is worse, but prevents sender from blocking, in case of Tokio crashes completely
+        let (sender, receiver) = crossbeam::channel::unbounded();
```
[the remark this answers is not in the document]
irvingoujAtDevolution: sorry, I did not explain clearly what was the concern. 
If the channel is full, then the sender would block and will crashes the thread that polls the socket's future.
So by setting it as unbounded, we made sure that the sending end of the channel will never block. 

Furthermore, since the the receiving end is running on a dedicated loop, it is not very likely to stack up the channel.
I agree that a sized channel might be better depending on the usage.

crates/network-scanner-net/src/async_raw_socket.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn drop(&mut self) {
+        self.runtime
+            .remove_socket(&self.socket)
+            .map_err(|e| tracing::error!("failed to remove socket from poller: {:?}", e))
+            .ok(); // ignore the error
+    }
+}
+
+impl AsyncRawSocket {
+    // prevent direct instantiation
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Agree, Updated

crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,152 @@
...
+
+impl Drop for Socket2Runtime {
+    fn drop(&mut self) {
+        self.is_terminated.store(true, std::sync::atomic::Ordering::SeqCst);
+        self.poller
+            .notify()
+            .map_err(|e| tracing::error!("failed to notify poller: {:?}", e))
+            .ok();
+        // event loop will terminate after this
+        // register loop will terminate becase of sender is dropped after this.
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/network-scanner-net/src/async_io.rs
irvingoujAtDevolution: agree,updated.

crates/network-scanner-net/src/async_raw_socket.rs
```diff
@@ -0,0 +1,134 @@
...
+        &'a self,
+        buf: &'a mut [MaybeUninit<u8>],
+    ) -> impl Future<Output = std::io::Result<(usize, SockAddr)>> + 'a {
+        RecvFromFuture {
+            socket: self.socket.clone(),
+            buf,
+            id: self.id,
+            runtime: self.runtime.clone(),
+        }
+    }
```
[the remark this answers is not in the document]
irvingoujAtDevolution: agree, updated

crates/network-scanner-net/src/async_io.rs
```diff
@@ -0,0 +1,152 @@
...
+                            let key = event.key;
+                            if let Some(waker) = map.get(&key) {
+                                waker.wake_by_ref();
+                                tracing::trace!(?key, "waker called");
+                                map.remove(&key);
+                            }
+                        }
+                    }
+                }
+                self.is_terminated.store(true, std::sync::atomic::Ordering::SeqCst);
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/network-scanner-net/src/async_raw_socket.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub fn set_broadcast(&self, broadcast: bool) -> std::io::Result<()> {
+        self.socket.as_ref().set_broadcast(broadcast)
+    }
+}
+
+struct RecvFromFuture<'a> {
+    pub socket: Arc<socket2::Socket>,
+    pub buf: &'a mut [MaybeUninit<u8>],
+    pub id: usize,
+    pub runtime: Arc<Socket2Runtime>,
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/network-scanner-net/Cargo.toml
irvingoujAtDevolution: updated

crates/network-scanner-net/examples/tcp.rs
```diff
@@ -0,0 +1,63 @@
...
+    let clone = socket_addr;
+    let handle = tokio::task::spawn(async move {
+        tracing::info!("connecting to cern using tokio");
+        let mut stream = tokio::net::TcpStream::connect(clone).await?; // making sure tokio io runtime does not collide with socket2 runtime
+        tracing::info!("conncted to cern using tokio");
+        let _ = stream.write(b"GET / HTTP/1.0\r\n\r\n").await?;
+        let buf = &mut [0u8; 1024];
+        let size = stream.read(buf).await?;
+        Ok::<Vec<_>, anyhow::Error>(buf[..size].to_vec())
+    }); // making sure tokio io event loop does not collide with socket2 event loop
```
CBenoit, line 36: question: I don’t quite understand what is making sure there is no collision? Also, what is a collision in this case? The socket2 event loop is started in a different thread, so my understanding is that it is running in parallel without requiring any special handling. I understand what you are trying to achieve is called "concurrency" using async tasks. Is that correct?

crates/network-scanner-net/Cargo.toml
```diff
@@ -6,9 +6,23 @@ edition = "2021"
...
+tokio = { version = "1.35.1", features = [
+    "rt",
+    "sync",
+    "macros",
+    "rt-multi-thread",
+    "time",
+    "net"
+] }
+trust-dns-resolver = "0.23.2"
```
CBenoit, line 28: issue: `trust-dns` project has been rebranded to "Hickory DNS" and has been moved to the https://github.com/hickory-dns/hickory-dns organization and repo

question: Do we really need to bring in this crate just for the example? Can’t we just use the built-in [`ToSocketAddrs` trait](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html)? There is also [an async version provided by Tokio](https://docs.rs/tokio/latest/tokio/net/trait.ToSocketAddrs.html).

suggestion: Remove the dependency if it’s not absolutely necessary.

crates/network-scanner-net/examples/udp.rs
```diff
@@ -0,0 +1,52 @@
...
+        tracing::info!("size: {}, addr: {:?}", size, addr);
+        socket.send_to(inited, &addr).await?;
+    }
+}
+
+unsafe fn init_buf(buf: &mut [MaybeUninit<u8>], size: usize) -> &[u8] {
+    let buf = &mut buf[..size];
+
+    std::mem::transmute::<&mut [MaybeUninit<u8>], &mut [u8]>(buf) as _
+}
```
CBenoit, line 52: suggestion: I suggest [this version instead](https://github.com/Devolutions/devolutions-gateway/blob/7157ad608278278d63096297ae1d6c04b768a984/crates/network-scanner/src/ping.rs#L50C1-L60C2).

crates/network-scanner/src/ping.rs
```diff
@@ -34,7 +34,7 @@ pub fn blocking_ping(ip: Ipv4Addr) -> anyhow::Result<()> {
     // elements are in an initialized state.
     let filled = unsafe { assume_init(&buffer[..size]) };
 
-    let packet = icmp_v4::Icmpv4Packet::parse(filled).with_context(|| format!("cannot parse icmp packet"))?;
+    let packet = icmp_v4::Icmpv4Packet::parse(filled).with_context(|| "cannot parse icmp packet".to_string())?;
```
CBenoit, line 37: nitpick: more concise
```suggestion
    let packet = icmp_v4::Icmpv4Packet::parse(filled).context("cannot parse icmp packet")?;
```

crates/network-scanner-net/src/async_io.rs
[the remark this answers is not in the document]
CBenoit: If the thread panics because of the `expect`, then the atomic will not be updated. Did you mean this?

```rust
if let Some(err) = self.poller.wait(&mut events, None) {
    error!(…);
    break;
}
```

crates/network-scanner-net/src/async_raw_socket.rs
```diff
@@ -0,0 +1,134 @@
...
+    runtime: Arc<Socket2Runtime>,
+    id: usize,
+}
+
+impl Drop for AsyncRawSocket {
+    fn drop(&mut self) {
+        self.runtime
+            .remove_socket(&self.socket)
+            .map_err(|e| tracing::error!("failed to remove socket from poller: {:?}", e))
+            .ok(); // ignore the error
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/network-scanner-net/examples/tcp.rs
irvingoujAtDevolution: When I mentioned the possibility of a collision, I was referring to ensuring that there is no conflict in accessing the underlying IOCP/epoll or kqueue systems. Since both Mio and Polling essentially use the same system calls, I just want to make sure that two separate IO event loops can work in conjunction with two separate instances of the OS's asynchronous IO primitives without any issues.

crates/network-scanner-net/src/async_io.rs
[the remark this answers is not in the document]
irvingoujAtDevolution: Update:
Since this will not be running on the main thread, a short blocking call is ok

crates/network-scanner-net/examples/tcp.rs
CBenoit: I see! Maybe you can you write this down in the comment, I think it would be useful to have the context

crates/network-scanner-net/Cargo.toml
irvingoujAtDevolution: you are right, updated

crates/network-scanner-net/examples/udp.rs
irvingoujAtDevolution: updated

crates/network-scanner-net/examples/tcp.rs
irvingoujAtDevolution: updated comments

crates/network-scanner-net/src/async_io.rs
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/network-scanner-net/src/async_io.rs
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

irvingoujAtDevolution:
Looks like CI is broken

irvingoujAtDevolution:
> Looks like CI is broken

CI is working agin

CBenoit:
@irvingoujAtDevolution The `multiple_tcp` test never ends on Windows. Can you check that?

crates/network-scanner-net/examples/broadcast.rs
```diff
@@ -32,3 +32,32 @@ pub async fn main() -> anyhow::Result<()> {
...
+//     socket.send_to(&echo_request, &SockAddr::from(addr))?;
+
+//     for i in 0..10 {
+//         let mut buf = [MaybeUninit::<u8>::uninit(); 1024];
+//         let (size, addr) = socket.recv_from(&mut buf)?;
+//         tracing::info!("counter = {i} size: {}, addr: {:?}", size, addr.as_socket_ipv4());
+//     }
+
+//     Ok(())
+// }
```
CBenoit, line 63: Leftover?

CBenoit:
@irvingoujAtDevolution Unfortunately, there is still a problem.

image

Note that it doesn’t always trigger. I ran the test 4 times, and it hanged once.

I’m considering disabling the faulty tests and merging now.

CBenoit (approved):
@irvingoujAtDevolution As mentioned, I disabled the tests temporarily. We’ll need to investigate that and re-enable again. I’ll merge now so we can collaborate more easily and push incremental improvements.

PR #663, opened 2024-01-24
fix(dgw): debug option to set the webapp path

CBenoit:
~~This also removes the `DGATEWAY_WEBAPP_PATH` env variable.~~
The `DGATEWAY_WEBAPP_PATH` env variable is conserved.

For example:
```json
{
  "WebApp": {
    "Enabled": true,
    "Authentication": "Custom",
    "StaticRootPath": "./webapp"
  },
}
```

This will reproduce the behavior of the initial implementation, finding the webapp folder in the current working directly.

cc @kristahouse

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,9 @@ Stable options are:
 
         Blank lines and lines starting by `#` are ignored.
 
+    * **StaticRootPath** (_FilePath_): Path to the static files for the standalone web application.
+        This is an advanced option which should typically not be changed.
+
 - **VerbosityProfile** (_String_): Logging verbosity profile (pre-defined tracing directives).
 
     Possible values:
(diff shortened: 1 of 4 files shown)


awakecoding:
@CBenoit can we keep the DGATEWAY_WEBAPP_PATH environment variable? I actually like it better if the replacement is a non-stable option under __debug__, as I'd really like to have a PowerShell code snippet to bootstrap the local dev configuration. If the WebAppPath option can be stabilized alongside the rest of the WebApp options, I could expose it properly in the PowerShell cmdlets, and that would work too. Even then, the environment variable is handy, I'd keep it

CBenoit:
@awakecoding No problem! I changed accordingly. The environment variable will be checked first, then the configuration key in the config file (`WebApp.StaticRootPath`), and if nothing is specified, it falls back to a `webapp` folder in the same folder as the executable.

thenextman (approved):

PR #664, opened 2024-01-24
feat(dwg): improved webapp error messages

kristahouse:
Implement UI for displaying info & error messages to user. 
Implement error UI for connection form. 
Note: more work to be done to conform to the figma design doc. This is first pass.

bug in the removing tabs/components.
bug in autologin when session expired.


awakecoding:
I just tried it with both auth disabled and enabled. With auth enabled, it looks like it's trying to auto-login with an empty user and password as it loads the login screen, causing an error in the server logs:

`2024-01-24T14:43:10.814412Z ERROR listener{port=7171}:http{client=[::1]:43189}:request{method=POST path=/jet/webapp/app-token}: devolutions_gateway::http: error=401 Unauthorized at devolutions-gateway\src\api\webapp.rs:164:28: user not found`

I guess it's not straightforward for the frontend to detect if auth is disabled or not? I just want to avoid log error pollution with the ERROR level, so if we can just ignore it in the backend that could work as well @CBenoit

For the error handling, I tried connecting with an invalid password:

image

The web client in DVLS is able to return more detailed authentication error information, I don't know what is calling but we should try exposing more details, especially error codes which may be available. A lack of a good error description can always be salvaged by an error code which we can use to diagnose issues.

For server-initiated disconnections, like session bumping, I got this:

image

Error codes are normally available for server-initiated disconnections, we have them here in FreeRDP:
https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/errinfo.c
https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/errinfo.h

The list is available from the protocols docs: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/a21a1bd9-2303-49c1-90ec-3932435c248c

Again, no need to go over the entire list, just the common ones for meaningful textual description, but it would be important to bubble up the error info error code. Some error codes can be handled explicitly not to show an "error" as they are server-initiated disconnections that are truly user-initiated, like ERRINFO_LOGOFF_BY_USER (0x0000000C) "The disconnection was initiated by the user logging off his or her session on the server."

All error codes should be in hexadecimal notation like (0x0000000C), it's easier to find them in the docs after.

awakecoding (approved):

awakecoding:
This being said, I think the changes are good, so the comments are really for the next pass of improvements in a future pull request

PR #637, opened 2024-01-10
feat(dgw): Network scan port discovery

irvingoujAtDevolution:
TCP based port discovery, this is a draft, waiting for https://github.com/Devolutions/devolutions-gateway/pull/636 to be merged first

PR #665, opened 2024-01-24
fix(dgw) : improvement in network-scanner-net

irvingoujAtDevolution:
The tests were always running fine on my machine, trying to figure out why it does not work in CI.

Edit 0: looks like it works in [this run](https://github.com/Devolutions/devolutions-gateway/actions/runs/7642687656), try remove tracing

Edit 1: looks like it is working again, re-run to make sure it really really works.

Edit 2:CI test is working again with this change, I'm confident this will fix the test


CBenoit:
@irvingoujAtDevolution Do you know what was the problem exactly?

CBenoit:
Looks like this is not fixing the problem. 4 runs out of 6 are never ending:
image

crates/network-scanner-net/Cargo.toml
```diff
@@ -9,10 +9,12 @@ publish = false
 [dependencies]
 anyhow = "1.0.79"
 crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
+futures = "0.3.30"
 parking_lot = "0.12.1"
-polling = "3.3.2"
+polling = {git = "https://github.com/irvingoujAtDevolution/polling.git"}
```
irvingoujAtDevolution, line 14: a updated version to polling, allow us to kown weather tcp connection has failed or not.

irvingoujAtDevolution:
It should really work this time, I have run it locally 100 times, all of them stopped within the time frame.
image

crates/network-scanner-net/Cargo.toml
irvingoujAtDevolution: waiting to be merged into polling main branch https://github.com/smol-rs/polling/pull/185

irvingoujAtDevolution:
> @irvingoujAtDevolution It might be the case that the problem is still not solved unfortunately :/ The CI has been running for more than 30 minutes on the windows job.

What? it has a

> @irvingoujAtDevolution It might be the case that the problem is still not solved unfortunately :/ The CI has been running for more than 30 minutes on the windows job.

Sorry I might looked at the wrong place, but it looks like it is passing in 3 seconds
image

CBenoit (changes requested):
Still very unstable. Also fails on Linux sometimes now.

image

irvingoujAtDevolution:
I offcially temproryily give up on fixing the test, I can have about 10% the chance fail the tests tracing_subscriber turned off, 0% if I have the tracing_subscriber turned on. 

I am confident that it will work fine with tcp, see intense_tcp.rs

I added ignore test for all the tests for now, will turn back and fix it in the future.

The improvement I obtained alone the way is good, now the async runtime drop correctly,  The first event information after calling add is registered correctly, the tcp connection established or fails correctly.

@CBenoit Thank you very much for the help, I'll try figure it out later

crates/network-scanner-net/Cargo.toml
irvingoujAtDevolution: It is merged, I will update it once they publish a new version

CBenoit:
Sorry, I reviewed the code in the other PR: https://github.com/Devolutions/devolutions-gateway/pull/637
Ideally make sure the base is properly set to the branch it is based on and keep the draft status.

irvingoujAtDevolution:
> Sorry, I reviewed the code in the other PR: #637 Ideally make sure the base is properly set to the branch it is based on and keep the draft status.

Updated according to the reviews

CBenoit:
I updated to point on a commit from the official polling repo, it’s less likely to break in the future.

CBenoit (approved):
Thank you!

PR #668, opened 2024-01-25
feat(dgw): webapp - implement the new session toolbar UI design.

kristahouse:
- Implemented the new session toolbar UI as per the figma design doc.  
- cleaned up and removed unused styles


CBenoit (approved):
The new toolbar is looking nice! Thank you :tada:

PR #671, opened 2024-01-29
feat(dgw): webapp: added dropdown for last recently used hostnames.

kristahouse:
- added p-autocomplete for the connection form. currently stores hostname entries forever. we can change this to be session based or implement an expiry. 

- fixed: Logout user when you try to create a session token and returns a 401.

- removed commented out code.

- improved some UI elements.


CBenoit (approved):
LGTM :rocket:

PR #678, opened 2024-01-30
feat(webapp): add full screen mode for monitor screen

kristahouse:
Added feature that allows for the session to go into full screen mode of the monitor. 

The session toolbar is initially hidden in full screen mode. It reappears when the user hovers the mouse at the top edge of the screen.


CBenoit (approved):
LGTM!

CBenoit:
Note: I think we could use a "webapp" scope from now on. It will be useful to see what is only affecting the web application in the README.

PR #680, opened 2024-02-01
refactor(dgw): fix clippy warnings in network scan

CBenoit:
This way, there is less indirections in our data structure.

What do you think? @irvingoujAtDevolution

### DIFF
--- a/crates/network-scanner-net/src/runtime.rs
+++ b/crates/network-scanner-net/src/runtime.rs
@@ -90,10 +90,11 @@ impl Socket2Runtime {
         register_receiver: Receiver<RegisterEvent>,
         event_sender: Sender<Event>,
     ) -> anyhow::Result<()> {
-        // we make is_terminated Arc<AtomicBool> and poller Arc<Poller> so that we can clone them and move them into the thread
-        // we cannot hold a Arc<Socket2Runtime> in the thread, because it will create a cycle reference and the runtime will never be dropped.
-        let is_terminated = self.is_terminated.clone();
+        // To prevent an Arc cycle with the Arc<Socket2Runtime>, we added additional indirection.
+        // Otherwise, the reference in the new thread would prevent the runtime from being dropped and shutdown.
         let poller = self.poller.clone();
+        let is_terminated = self.is_terminated.clone();
+
         std::thread::Builder::new()
             .name("[raw-socket]:io-event-loop".to_string())
             .spawn(move || {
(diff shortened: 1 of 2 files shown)


crates/network-scanner-net/src/runtime.rs
```diff
@@ -106,14 +106,18 @@ impl Socket2Runtime {
...
+                    if runtime.is_terminated.load(Ordering::Acquire) {
                         break;
                     }
 
                     tracing::debug!("polling events");
-                    if let Err(e) = poller.wait(&mut events, None) {
+                    if let Err(e) = runtime.poller.wait(&mut events, None) {
                         tracing::error!(error = ?e, "failed to poll events");
-                        is_terminated.store(true, Ordering::SeqCst);
+                        runtime.is_terminated.store(true, Ordering::SeqCst);
```
irvingoujAtDevolution, line 120: would it panic if runtime is dropped here? as poller.wait() will stop blocking because of notify in the drop function of runtime?

crates/network-scanner-net/src/runtime.rs
```diff
@@ -106,14 +106,18 @@ impl Socket2Runtime {
                 let mut events_happened = HashMap::new();
 
                 loop {
-                    if is_terminated.load(Ordering::Acquire) {
+                    let Some(runtime) = runtime.upgrade() else {
```
irvingoujAtDevolution, line 109: I suspect this will work, as upgrates still [increases the Arc count ](https://github.com/rust-lang/rust/blob/11f32b73e0dc9287e305b5b9980d24aecdc8c17f/library/alloc/src/sync.rs#L2781).

Once it turns into an Arc, the drop function will not be called, and hence the notify will never happen

crates/network-scanner-net/src/runtime.rs
```diff
@@ -106,14 +106,18 @@ impl Socket2Runtime {
...
+                        break;
+                    };
+
+                    if runtime.is_terminated.load(Ordering::Acquire) {
                         break;
                     }
 
                     tracing::debug!("polling events");
-                    if let Err(e) = poller.wait(&mut events, None) {
+                    if let Err(e) = runtime.poller.wait(&mut events, None) {
```
irvingoujAtDevolution, line 118: and it will block here forever

CBenoit:
@irvingoujAtDevolution You’re right. I’ll update the PR to only fix the warnings and revert the rest.

irvingoujAtDevolution (approved):
Thank you!

PR #683, opened 2024-02-05
feat(network scan): A working scanner with ping, port discovery, broadcast, and netbios

irvingoujAtDevolution:
This PR contains code necessary for new functionalities, changes,improvement and refactoring of the network scan crates. Please see network-scan/examples for usages.

@CBenoit Let me know if smaller PR would be preferable, The main changes include
1. uses DashMap to imporve the preformance of event loop
2. add netbios,ping,port discovery and broadcast
3. integrate everything to make a working scanner


crates/network-scanner-net/src/runtime.rs
```diff
@@ -81,130 +79,99 @@ impl Socket2Runtime {
...
         &self,
         register_receiver: Receiver<RegisterEvent>,
-        event_sender: Sender<Event>,
+        event_history: Arc<DashSet<EventWrapper>>,
     ) -> anyhow::Result<()> {
-        // To prevent an Arc cycle with the Arc<Socket2Runtime>, we added additional indirection.
-        // Otherwise, the reference in the new thread would prevent the runtime from being dropped and shutdown.
-        let poller = self.poller.clone();
+        // we make is_terminated Arc<AtomicBool> and poller Arc<Poller> so that we can clone them and move them into the thread
+        // we cannot hold a Arc<Socket2Runtime> in the thread, because it will create a cycle reference and the runtime will never be dropped.
```
CBenoit, line 92: style: I encourage writing proper sentences starting with an uppercase and ending with a period.

crates/network-scanner/src/broadcast/block.rs
CBenoit, line 1: naming: `blocking`

crates/network-scanner/src/task_utils.rs
```diff
@@ -0,0 +1,153 @@
...
+pub fn clone5<A, B, C, D, E>(a: &A, b: &B, c: &C, d: &D, e: &E) -> (A, B, C, D, E)
+where
+    A: Clone,
+    B: Clone,
+    C: Clone,
+    D: Clone,
+    E: Clone,
+{
+    (a.clone(), b.clone(), c.clone(), d.clone(), e.clone())
+}
```
CBenoit, line 146: nitpick: I don’t think this is very idiomatic. We typically just use `.clone()` directly explicitly. I don’t really want to adopt this style if we’re not going to be consistent.

CBenoit (commented):
Nice work!

> Let me know if smaller PR would be preferable

It would indeed be very much welcomed as smaller PRs. Probably it’s too much work for you to change that now so we can try to handle this the way it is now. I’ll not go through every change though.

>     1. uses DashMap to imporve the preformance of event loop

Did you perform any form of benchmarking before making this move?
I would be interested in seeing some numbers before we add this dependency.
(Following on the above: this alone could be a dedicated PR.)

I’ll continue reviewing tomorrow.

crates/network-scanner/src/task_utils.rs
irvingoujAtDevolution: I understand I hesitated when I wrote this code, but it really helps with readability, there are just too many to clone when we need to use these variables in different tasks. using an extra 5 lines is really bad for readability, considering there are already a lot of indentations and logic flooding in the tasks.

I can add #[inline] tag so it would compile to the same assembly code

crates/network-scanner/src/broadcast/block.rs
irvingoujAtDevolution: updated

crates/network-scanner/src/task_utils.rs
irvingoujAtDevolution: updated

irvingoujAtDevolution:
> Nice work!
> 
> > Let me know if smaller PR would be preferable
> 
> It would indeed be very much welcomed as smaller PRs. Probably it’s too much work for you to change that now so we can try to handle this the way it is now. I’ll not go through every change though.
> 
> > ```
> > 1. uses DashMap to imporve the preformance of event loop
> > ```
> 
> Did you perform any form of benchmarking before making this move? I would be interested in seeing some numbers before we add this dependency. (Following on the above: this alone could be a dedicated PR.)
> 
> I’ll continue reviewing tomorrow.

I'll try give a benchmark, but notably,even if the performance is slightly worse, the logic is much more strightforward and now lock-free.

crates/network-scanner-net/Cargo.toml
```diff
@@ -8,9 +8,10 @@ publish = false
 [dependencies]
 anyhow = "1.0.79"
 crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
+dashmap = "5.5.3"
 futures = "0.3.30"
 parking_lot = "0.12.1"
-polling = { git = "https://github.com/smol-rs/polling", rev = "62430fd56e668559d08ca7071ab13a0e116ba515" }
+polling = { git = "https://github.com/irvingoujAtDevolution/polling.git", branch = "fix-linux-tcp-fail" }
```
irvingoujAtDevolution, line 14: Wasn't aware the problem of Linux side untile the CI reported me error. :(. 

This is currently necessary for linux TCP to work, see line 237 for reason

crates/network-scanner-net/src/runtime.rs
```diff
@@ -40,7 +41,7 @@ impl Drop for Socket2Runtime {
     }
 }
 
-const QUEUE_CAPACITY: usize = 1024;
+const QUEUE_CAPACITY: usize = 8024;
```
irvingoujAtDevolution, line 44: Maybe too big?

crates/network-scanner-net/src/socket.rs
```diff
@@ -203,15 +219,32 @@ impl<'a> Future for ConnectFuture<'a> {
...
+                    "connection failed",
+                )));
+            }
 
-        if !events.is_empty() {
-            tracing::trace!(?events, "connect success");
-            return std::task::Poll::Ready(Ok(()));
+            // This is a special case, this happens when using epoll to wait for a unconnected TCP socket.
+            // We clearly needs to call connect function, so we break the loop and call connect.
+            #[cfg(target_os = "linux")]
```
irvingoujAtDevolution, line 237: Linux and Windows have different behavior, Linux shots event to you even before you call connect, we need to identify if the event is "Connect failed" or "Connect successful" or something else (out and hup). Polling does not expose plateform dependent args, so we have to make a banch specific to our needs here. Will disguss with Polling's owner later.

crates/network-scanner-net/src/test.rs
```diff
@@ -15,7 +15,6 @@ use tokio::{
 use crate::socket::AsyncRawSocket;
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
-#[ignore = "TODO"]
 async fn multiple_udp() -> anyhow::Result<()> {
```
irvingoujAtDevolution, line 18: Looks like all tests have passed. 😆 
image

crates/network-scanner-proto/Cargo.toml
```diff
@@ -1,12 +1,15 @@
...
-authors = ["Devolutions Inc. <infos@devolutions.net>","Jeremy Wall <jeremy@marzhillstudios.com>"]
+authors = [
+    "Devolutions Inc. <infos@devolutions.net>",
+    "Jeremy Wall <jeremy@marzhillstudios.com>",
+    "Jon Grimes <jonkgrimes@gmail.com>"
+]
 edition = "2021"
 publish = false
-license = "Apache-2.0"
+license = "MIT, Apache-2.0"
```
irvingoujAtDevolution, line 11: Is this enough? the newly added code are under MIT

crates/network-scanner/src/ping.rs
```diff
@@ -19,58 +85,16 @@ pub fn blocking_ping(ip: Ipv4Addr) -> anyhow::Result<()> {
...
         .recv_from(&mut buffer)
         .with_context(|| format!("failed to receive packet from {}", ip))?;
 
-    // SAFETY: `recv_from` returns the number of bytes written into the buffer, so the `size` first
-    // elements are in an initialized state.
-    let filled = unsafe { assume_init(&buffer[..size]) };
-
-    let packet = icmp_v4::Icmpv4Packet::parse(filled).context("cannot parse icmp packet")?;
-
-    if let icmp_v4::Icmpv4Message::EchoReply { payload, .. } = packet.message {
```
irvingoujAtDevolution, line 39: don't care about what the response is, we only care about there's a computer on the other end.

crates/network-scanner/src/task_utils.rs
```diff
@@ -0,0 +1,108 @@
...
+type HandlesReceiver = crossbeam::channel::Receiver<tokio::task::JoinHandle<anyhow::Result<()>>>;
+type HandlesSender = crossbeam::channel::Sender<tokio::task::JoinHandle<anyhow::Result<()>>>;
+#[derive(Debug)]
+pub(crate) struct TaskExecutionRunner {
+    pub(crate) context: TaskExecutionContext,
+    pub(crate) handles_receiver: HandlesReceiver,
+    pub(crate) handles_sender: HandlesSender,
+}
+
+impl TaskExecutionContext {
```
irvingoujAtDevolution, line 47: This is created purely for the sake of readability, it was super hard of keep track of things without it.

crates/network-scanner/src/task_utils.rs
```diff
@@ -0,0 +1,108 @@
...
+        T: FnOnce(TaskExecutionContext) -> F + Send + 'static,
+        F: Future<Output = anyhow::Result<()>> + Send + 'static,
+    {
+        let context = self.context.clone();
+        let handle = tokio::task::spawn(async move { task(context).await });
+        let _ = self.handles_sender.send(handle);
+    }
+
+    pub(crate) fn new(scanner: NetworkScanner) -> anyhow::Result<Self> {
+        let (handles_sender, handles_receiver) = crossbeam::channel::unbounded();
```
irvingoujAtDevolution, line 101: needs to be unbounded, if this is blocked, then all async task are blocked indefinately.

crates/network-scanner-net/src/runtime.rs
```diff
@@ -81,130 +80,105 @@ impl Socket2Runtime {
...
                         is_terminated.store(true, Ordering::SeqCst);
                         break;
                     };
+
                     for event in events.iter() {
                         tracing::trace!(?event, "event happened");
-                        events_happened.insert(event.key, event);
+                        // This is different from just insert, as the event wrapper will have the same hash, it actually does not replace the old one.
+                        // by removing the old one first, we can make sure the new one is inserted.
+                        event_history.remove(&event.into());
```
irvingoujAtDevolution, line 121: The underlying data structure is a hash map, and the value as set is the key to the map. Which will not be updated if the hash is the same. 
We need to replace the value in the set because we need the latest event with it's system specific event arg, in particular, for linux's OUT and HUP.
Since the underlying system specific event are not exposed, we cannot hash it, so instead, we do it this way to make sure it's the latest event that is actually in the cache,

crates/network-scanner-net/src/test.rs
```diff
@@ -179,50 +175,28 @@ async fn work_with_tokio_tcp() -> anyhow::Result<()> {
...
 
-            tracing::info!("good_socket: {:?}", good_socket);
             tracing::info!("bad_socket: {:?}", bad_socket);
 
             let unused_port = 12345;
 
-            let available_addr = addr;
             let non_available_addr = SocketAddr::from(([127, 0, 0, 1], unused_port));
 
-            let handle =
```
irvingoujAtDevolution, line 202: Maybe should add this back?

crates/network-scanner-net/Cargo.toml
CBenoit: Can you link to the issue / pull request you created on their side?

crates/network-scanner-net/src/runtime.rs
CBenoit: Any reason why you changed?

crates/network-scanner-net/src/test.rs
CBenoit: Nice!!

crates/network-scanner-proto/Cargo.toml
CBenoit: I think it’s fine! But you should write it like this:

```suggestion
license = "MIT/Apache-2.0"
```

crates/network-scanner/src/ping.rs
CBenoit: Fair enough!

crates/network-scanner/src/task_utils.rs
CBenoit: Please add a comment explaining the rationale directly in the code, we’ll eventually forget otherwise

crates/network-scanner-net/src/test.rs
CBenoit: If we are already testing the same thing elsewhere, it’s okay to not I guess? As you prefer!

crates/network-scanner-net/src/test.rs
CBenoit, line 1: suggestion: Can be a separate PR, but I suggest moving this file as an integration test instead

CBenoit (commented):
> I'll try give a benchmark, but notably,even if the performance is slightly worse, the logic is much more strightforward and now lock-free.

I may remember incorrectly, but I don’t think `dashmap` is implemented in a [lock-free manner](https://en.wikipedia.org/wiki/Non-blocking_algorithm)? (IIRC it’s a sharded hashmap).

To be honest, I don’t think that removing a bunch of `.lock()` calls is that much compelling and my main concern is more about bloating the executable and increasing the compile time for something that we don’t necessarily need, without mentioning that we are also increasing our supply chain attack surface.
I’m not necessarily against `dashmap` if it’s worth it and we actually need it, but I would have preferred a more principled approach.
That’s why this kind of change is typically much more welcomed as a dedicated PR: we can easily compare without too much noise. Now it’s going to be painful to go back if we decide there is not enough value to it.

> Plus, it gives our the potential ability to implement method like `async fn writeable(&self)` or `async fn readable(&self)` in the future, by checking if cetarin event exist in the event cache.

I’m not sure to follow. Are you referring to interior mutability? We also have that with `Mutex<HashMap<T>>` or `RwLock<HashMap<T>>`, and `parking_lot` are not returning `Result` so it’s only just one more `.lock()` to write. (No even `.lock().unwrap()`.)

Again, maybe we want it, I’ll think through it again.

crates/network-scanner-net/src/runtime.rs
irvingoujAtDevolution: yes, it was causing errors on registering the events.

crates/network-scanner-net/Cargo.toml
irvingoujAtDevolution: the problem on polling is that they are trying to have a platform indepent interface for all polling mechanisms, so I guess they would particularly against have something like this. I have a better idea in mind to how to implement this so it is more acceptable for both us and them, I'll put on a update shortly.

crates/network-scanner-net/src/test.rs
irvingoujAtDevolution: you are absolutely correct! inner mutability can be achinve using mutex hash map as well. I'll give a benchmarks for both implementation seprately.

crates/network-scanner-proto/README.md
```diff
@@ -2,4 +2,6 @@
 This crates contains network packet struct definitions and parsers.
 
 ## Acknowledgement 
-This project includes portions of [icmp-socket](https://github.com/zaphar/icmp-socket), held by Jeremy Wall, 2022.
\ No newline at end of file
+This project includes portions of
+1. [icmp-socket](https://github.com/zaphar/icmp-socket), held by Jeremy Wall, 2022.
+2. [nbtscanner]  (https://github.com/jonkgrimes/nbtscanner/blob/main/src/nbt_packet.rs) held by Jon Grimes,2018
```
CBenoit, line 7:
```suggestion
2. [nbtscanner](https://github.com/jonkgrimes/nbtscanner/blob/main/src/nbt_packet.rs) held by Jon Grimes,2018
```

crates/network-scanner/src/task_utils.rs
irvingoujAtDevolution: added

crates/network-scanner-net/Cargo.toml
irvingoujAtDevolution: https://github.com/smol-rs/polling/actions/runs/7819053683/job/21330660851?pr=189

crates/network-scanner-net/src/runtime.rs
```diff
@@ -40,7 +41,7 @@ impl Drop for Socket2Runtime {
     }
 }
 
-const QUEUE_CAPACITY: usize = 1024;
+const QUEUE_CAPACITY: usize = 8024;
 impl Socket2Runtime {
     /// Create a new runtime with a queue capacity, default is 1024.
```
CBenoit, line 46: The default has been changed

crates/network-scanner-net/Cargo.toml
```diff
@@ -10,7 +10,7 @@ anyhow = "1.0.79"
 crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
 futures = "0.3.30"
 parking_lot = "0.12.1"
-polling = { git = "https://github.com/irvingoujAtDevolution/polling.git", branch = "fix-linux-tcp-fail" }
+polling = { git = "https://github.com/irvingoujAtDevolution/polling.git", branch = "fix-detect-connection-fail" }
```
CBenoit: Before we can merge, we need to move this into a repository living in the Devolutions organization. You can either give the ownership of the fork to Devolutions or ask devops to clone the repo for us and push your modification there.

crates/network-scanner-net/src/runtime.rs
irvingoujAtDevolution: updated

Cargo.lock
CBenoit, line 1: Can you make sure the Cargo.lock is not containing unrequired changes? I still see `dashmap`
You can reset it to the revision on master and rerun `cargo check` for instance

crates/network-scanner/Cargo.toml
```diff
@@ -8,12 +8,20 @@ publish = false
...
+serde = { version = "1.0.196", features = ["serde_derive"] }
 socket2 = "0.5.5"
-tokio = { version = "1.36.0", features = ["io-util"] }
+thiserror = "1.0.56"
+tokio = { version = "1.35.1" , features = ["rt"]}
 tracing = "0.1.40"
 
 [dev-dependencies]
-tokio = { version = "1.36.0", features = ["rt","macros","rt-multi-thread"] }
+tokio = { version = "1.35.1", features = ["rt","macros","rt-multi-thread","tracing"] }
```
CBenoit, line 26: issue: Probably a merge conflict which went wrong, but we are now using tokio 1.36.0, not 1.35.1

crates/network-scanner/Cargo.toml
```diff
@@ -8,12 +8,20 @@ publish = false
...
 anyhow = "1.0.79"
+crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
+dashmap = "5.5.3"
+derive_builder = "0.13.0"
+futures = { version = "0.3.30" }
+lazy_static = "1.4.0"
+network-interface = "1.1.1"
 network-scanner-net ={ path = "../network-scanner-net" }
 network-scanner-proto ={ path = "../network-scanner-proto" }
+serde = { version = "1.0.196", features = ["serde_derive"] }
```
CBenoit, line 19: question: Why do we need serde in this crate?

crates/network-scanner/Cargo.toml
```diff
@@ -8,12 +8,20 @@ publish = false
 
 [dependencies]
 anyhow = "1.0.79"
+crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
+dashmap = "5.5.3"
+derive_builder = "0.13.0"
+futures = { version = "0.3.30" }
+lazy_static = "1.4.0"
```
CBenoit, line 15: question: Unused?

crates/network-scanner/Cargo.toml
```diff
@@ -8,12 +8,20 @@ publish = false
 
 [dependencies]
 anyhow = "1.0.79"
+crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
+dashmap = "5.5.3"
+derive_builder = "0.13.0"
+futures = { version = "0.3.30" }
```
CBenoit, line 14: issue: Unused dependency.

crates/network-scanner/Cargo.toml
```diff
@@ -8,12 +8,20 @@ publish = false
 
 [dependencies]
 anyhow = "1.0.79"
+crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
+dashmap = "5.5.3"
+derive_builder = "0.13.0"
```
CBenoit, line 13: issue: We are already using typed-builder in the rest of the codebase, I would prefer we stick to this

crates/network-scanner/Cargo.toml
```diff
@@ -8,12 +8,20 @@ publish = false
 
 [dependencies]
 anyhow = "1.0.79"
+crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
+dashmap = "5.5.3"
```
CBenoit, line 12: issue: Now unused

crates/network-scanner/src/task_utils.rs
```diff
@@ -0,0 +1,174 @@
+use std::sync::atomic::{AtomicBool, AtomicUsize};
+use std::{net::IpAddr, sync::Arc, time::Duration};
+
+use dashmap::DashMap;
+
+use futures::Future;
```
CBenoit, line 6: issue: The `Future` trait is now part of `std`
```suggestion
use std::future::Future;
```

crates/network-scanner/src/task_utils.rs
```diff
@@ -0,0 +1,174 @@
...
+use crate::{
+    ip_utils::{get_subnets, Subnet},
+    scanner::NetworkScanner,
+};
+
+pub(crate) type IpSender = tokio::sync::mpsc::Sender<(IpAddr, Option<String>)>;
+pub(crate) type IpReceiver = tokio::sync::mpsc::Receiver<(IpAddr, Option<String>)>;
+pub(crate) type PortSender = tokio::sync::mpsc::Sender<(IpAddr, Option<String>, u16)>;
+pub(crate) type PortReceiver = tokio::sync::mpsc::Receiver<(IpAddr, Option<String>, u16)>;
+#[derive(Debug, Clone)]
```
CBenoit, line 19: nitpick: Spacing

crates/network-scanner/Cargo.toml
irvingoujAtDevolution: this is another crates, where it used to store scanned IPs, it is unused in network-scan-net

crates/network-scanner/src/task_utils.rs
irvingoujAtDevolution: updated

CBenoit (commented):
We need to clean up the dependencies before we merge. Incidentally, the dependency on `futures` is unrequired not only for `network-scanner` but also for `network-scanner-net`

crates/network-scanner/Cargo.toml
CBenoit: We probably don’t need dashmap in `network-scanner` either

crates/network-scanner/Cargo.toml
CBenoit: By the way, the standard library now has OnceLock which can be always be used instead of lazy_static: https://doc.rust-lang.org/std/sync/struct.OnceLock.html

crates/network-scanner/Cargo.toml
irvingoujAtDevolution: it was easier for the main devolutions bin to use as query parameter, removed for now

crates/network-scanner/Cargo.toml
irvingoujAtDevolution: was trying to use it to have singleton runtime/task_manager, removed

crates/network-scanner/Cargo.toml

crates/network-scanner/Cargo.toml
CBenoit: https://users.rust-lang.org/t/if-let-else-if-let-else-better-way-to-structure-code/104393/10?u=cbenoit

crates/network-scanner/Cargo.toml

CBenoit (commented):
There is also a conflict to fix with master, but otherwise I’m thinking about merging today

crates/network-scanner/Cargo.toml
irvingoujAtDevolution: we might? we actually have a lot concurrent request

CBenoit: I’ll be honest here, dashmap is probably not worth anything unless we hit at least 10 000 000 operations per seconds, and I highly doubt we’ll ever because we are I/O-bound
This kind of data structure is a useful building block for CPU-bound algorithms

irvingoujAtDevolution: udpated

crates/network-scanner/src/task_utils.rs
```diff
@@ -25,7 +24,8 @@ pub(crate) struct TaskExecutionContext {
     pub port_sender: PortSender,
     pub port_receiver: Arc<Mutex<PortReceiver>>,
 
-    pub ip_cache: Arc<DashMap<IpAddr, Option<String>>>,
+    pub ip_cache: Arc<parking_lot::Mutex<HashMap<IpAddr, Option<String>>>>,
```
CBenoit, line 27: suggestion: You could use https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html (if we have more reads that writes)

irvingoujAtDevolution:
> We need to clean up the dependencies before we merge. Incidentally, the dependency on `futures` is unrequired not only for `network-scanner` but also for `network-scanner-net`

updated, removed all unsued dependencies

irvingoujAtDevolution:
> There is also a conflict to fix with master, but otherwise I’m thinking about merging today

updated, thank you!

crates/network-scanner/src/task_utils.rs
irvingoujAtDevolution: updated

crates/network-scanner/Cargo.toml

Cargo.lock
irvingoujAtDevolution: updated

crates/network-scanner-net/Cargo.toml
```diff
@@ -9,7 +9,7 @@ publish = false
 anyhow = "1.0.79"
 crossbeam = { version = "0.8.4", features = ["crossbeam-channel"] }
 parking_lot = "0.12.1"
-polling = { git = "https://github.com/devolutions/polling.git", branch = "fix-detect-connection-fail" }
+polling = { git = "https://github.com/devolutions/polling.git", rev = "c04e8ee40415cad551fe044457270f4a2d7c491d" }
```
CBenoit, line 12: Let’s hardcode the commit rather than the branch

CBenoit (approved):
Let’s merge. Thank you for this work!
In the future, try to send single-purpose PRs, please.
It’s okay even if the PR is very small, it’ll be quick to review.

PR #686, opened 2024-02-08
fix(webapp): bug fixes for removing connections & added full screen size feature.

kristahouse:
- added full screen mode.
- fixed screen size data sent to IronRDP component
- fixed screen size when default setting or the full screen setting is used in input form.
- fixed when token expires and session data was not removed.


CBenoit (approved):
LGTM!

PR #687, opened 2024-02-09
fix(webapp): terminate session connections on logout.

kristahouse:
-Bug fix: session connections were not closed upon logout. 
-removed dead code


CBenoit (approved):
Thank you!!

PR #688, opened 2024-02-11
fix(dgw): make NETSCAN tokens short-lived

irvingoujAtDevolution:
As per previous Slack discussion, netowrk scan token should be used only once.


CBenoit (approved):
Thank you!

PR #689, opened 2024-02-11
feat(dgw): network scan HTTP API

irvingoujAtDevolution:
Added basic api websocket endpoint for network scan

### DIFF
--- a/devolutions-gateway/Cargo.toml
+++ b/devolutions-gateway/Cargo.toml
@@ -94,6 +94,7 @@ packet = { git = "https://github.com/fdubois1/rust-packet.git" }
 
 # For KDC proxy
 portpicker = "0.1"
+network-scanner = { version = "0.0.0", path = "../crates/network-scanner" }
 
 [target.'cfg(windows)'.dependencies]
 rustls-cng = "0.3"
--- a/devolutions-gateway/Cargo.toml
+++ b/devolutions-gateway/Cargo.toml
@@ -23,6 +23,7 @@ ironrdp-pdu = { version = "0.1", git = "https://github.com/Devolutions/IronRDP",
 ironrdp-rdcleanpath = { version = "0.1", git = "https://github.com/Devolutions/IronRDP", rev = "4844e77b7f65024d85ba74b1824013eda6eb32b2" }
 ceviche = "0.5"
 picky-krb = "0.8"
+network-scanner = { version = "0.0.0", path = "../crates/network-scanner" }
 
 # Serialization
 serde = "1.0"
@@ -94,7 +95,6 @@ packet = { git = "https://github.com/fdubois1/rust-packet.git" }
 
 # For KDC proxy
 portpicker = "0.1"
-network-scanner = { version = "0.0.0", path = "../crates/network-scanner" }
 
 [target.'cfg(windows)'.dependencies]
 rustls-cng = "0.3"
--- a/devolutions-gateway/src/api/network_scan.rs
+++ b/devolutions-gateway/src/api/network_scan.rs
@@ -55,7 +55,7 @@ pub async fn handler(
                 }
             }
         }
-        tracing::info!("Network scan finished");
+        info!("Network scan finished");
         stream.stop();
     });
     Ok(res)
--- a/devolutions-gateway/src/api/network_scan.rs
+++ b/devolutions-gateway/src/api/network_scan.rs
@@ -1,0 +1,123 @@
+use std::net::IpAddr;
+
+use axum::{
+    extract::{ws::Message, State, WebSocketUpgrade},
+    response::Response,
+};
+
+use network_scanner::scanner::NetworkScannerParams;
+use serde::Serialize;
+
+use crate::{http::HttpError, DgwState};
+
+pub async fn handler(
+    _: State<DgwState>,
+    ws: WebSocketUpgrade,
+    query_params: axum::extract::Query<NetworkScanQueryParams>,
+) -> Result<Response, HttpError> {
+    let scanner_params: NetworkScannerParams = query_params.0.into();
+
+    let scanner = network_scanner::scanner::NetworkScanner::new(scanner_params).map_err(|e| {
+        tracing::error!("Failed to create network scanner: {:?}", e);
+        HttpError::internal().build(e)
+    })?;
+
+    let res = ws.on_upgrade(move |mut websocket| async move {
+        let stream = scanner.start().expect("Failed to start network scanner");
+        tracing::info!("Network scan started");
+        loop {
+            tokio::select! {
+                result = stream.recv() => {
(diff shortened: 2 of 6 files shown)


devolutions-gateway/Cargo.toml
```diff
@@ -94,6 +94,7 @@ packet = { git = "https://github.com/fdubois1/rust-packet.git" }
 
 # For KDC proxy
 portpicker = "0.1"
+network-scanner = { version = "0.0.0", path = "../crates/network-scanner" }
```
CBenoit, line 97: nitpick: This is unrelated to the KDC proxy, you can move this to the "in-house" section instead

devolutions-gateway/src/api/network_scan.rs
```diff
@@ -0,0 +1,123 @@
...
+                        break;
+                    };
+
+                    if let Ok(Message::Close(_)) = msg {
+                        break;
+                    }
+                }
+            }
+        }
+        tracing::info!("Network scan finished");
```
CBenoit, line 58: nitpick: tracing macros are all imported by default (see the #[macro_use] in lib.rs). Same at the other places.

```suggestion
        info!("Network scan finished");
```

devolutions-gateway/src/api/network_scan.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct NetworkScanQueryParams {
+    pub ping_interval: Option<u64>,     // in milliseconds,default 200
+    pub ping_timeout: Option<u64>,      // in milliseconds,default 500
+    pub broadcast_timeout: Option<u64>, // in milliseconds,default 1000
+    pub port_scan_timeout: Option<u64>, // in milliseconds,default 1000
+    pub netbios_timeout: Option<u64>,   // in milliseconds,default 1000
+    pub netbios_interval: Option<u64>,  // in milliseconds,default 200
+    pub max_wait: Option<u64>,          // max_wait for entire scan duration in milliseconds, suggested!
```
CBenoit, line 72: nitpick: Use doc strings

```rust
    // Ping interval in milliseconds (default is 200)
    pub ping_interval: Option<u64>,
```

devolutions-gateway/src/api/network_scan.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+const FAMOUS_PORTS: [u16; 10] = [22, 23, 80, 443, 389, 636, 3389, 5900, 5985, 5986];
```
CBenoit, line 75: naming: `COMMON_PORTS`

devolutions-gateway/src/api/network_scan.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+#[derive(Debug, Serialize)]
+pub struct NetworkScanResponse {
+    pub ip: String,
+    pub hostname: Option<String>,
+    #[serde(rename = "type")]
+    pub type_: String,
+}
```
CBenoit, line 97: suggestion: For `ip`, you can use the `IpAddr` type directly instead of `String`. `Serialize` is implemented on it.

naming: `ty` is a common abbreviation to avoid the alien trailing `_`

BUT, question: what about calling this field `protocol` or `application_protocol` (we use the word "application protocol" abbreviated to "ap" in our tokens)

suggestion: Maybe we can reuse the `Protocol` type found in the `token` module, instead of `String`.
This way, we also keep the same stringified representation at both places (e.g.: "ssh" instead of "SSH" and the "winrm-http-pwsh" / "winrm-https-pwsh" combination instead of WinRM)

devolutions-gateway/src/api/network_scan.rs
```diff
@@ -0,0 +1,123 @@
...
+    response::Response,
+};
+
+use network_scanner::scanner::NetworkScannerParams;
+use serde::Serialize;
+
+use crate::{http::HttpError, DgwState};
+
+pub async fn handler(
+    _: State<DgwState>,
```
CBenoit, line 14: question: Do we absolutely need to have this parameter? Seems unrequired, but maybe axum needs it for compilation.

devolutions-gateway/src/api/network_scan.rs
```diff
@@ -0,0 +1,123 @@
...
+    ws: WebSocketUpgrade,
+    query_params: axum::extract::Query<NetworkScanQueryParams>,
+) -> Result<Response, HttpError> {
```
CBenoit, line 17: issue: Seems like we are not checking for the NETSCAN token. See the `get_sessions` route for an example using an extractor.

devolutions-gateway/src/api/network_scan.rs
```diff
@@ -0,0 +1,123 @@
...
+    let scanner_params: NetworkScannerParams = query_params.0.into();
+
+    let scanner = network_scanner::scanner::NetworkScanner::new(scanner_params).map_err(|e| {
+        tracing::error!("Failed to create network scanner: {:?}", e);
+        HttpError::internal().build(e)
+    })?;
+
+    let res = ws.on_upgrade(move |mut websocket| async move {
+        let stream = scanner.start().expect("Failed to start network scanner");
```
CBenoit, line 26: issue: `expect` will panics on failure. If you can’t return an error, log an error instead.

CBenoit (commented):
Thank you!

devolutions-gateway/Cargo.toml
irvingoujAtDevolution: updated

devolutions-gateway/src/api/network_scan.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/api/network_scan.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/api/network_scan.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/api/network_scan.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/api/network_scan.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/api/network_scan.rs
irvingoujAtDevolution: updated, we don't need it, thanks!

devolutions-gateway/src/api/network_scan.rs
irvingoujAtDevolution: updated

PR #692, opened 2024-02-12
fix(webapp): fix selection of hostname rdp form

kristahouse:
- small fix for setting the hostname in the rdp form when the form is returned from ctrl-alt-del sign out, a disconnect via the Start Menu, or an error in a connection.


CBenoit (approved):
LGTM!

PR #693, opened 2024-02-12
feat(dgw): mDNS network scan support

irvingoujAtDevolution:
Add mdns to network scan, added some extra helper method for better readbility and maintainbility.

Open for review but needs to wait for the #689 to be merged first

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,9 +29,9 @@ dependencies = [
 
 [[package]]
 name = "aes"
-version = "0.8.3"
+version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac1f845298e95f983ff1944b728ae08b8cebab80d684f0a832ed0fc74dfa27e2"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
@@ -533,9 +533,9 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chrono"
-version = "0.4.33"
+version = "0.4.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9f13690e35a5e4ace198e7beea2895d29f3a9cc55015fcebe6336bd2010af9eb"
+checksum = "5bc015644b92d5890fab7489e49d21f879d5c990186827d42ec511919404f38b"
 dependencies = [
  "android-tzdata",
  "iana-time-zone",
@@ -597,9 +597,9 @@ dependencies = [
 
 [[package]]
 name = "crc32fast"
-version = "1.3.2"
+version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b540bd8bc810d3885c6ea91e2018302f68baba2129ab3e88f32389ee9370880d"
+checksum = "b3855a8a784b474f333699ef2bbca9db2c4a1f6d9088a90a2d25b1eb53111eaa"
 dependencies = [
  "cfg-if",
 ]
@@ -1468,9 +1468,9 @@ dependencies = [
 
 [[package]]
 name = "hermit-abi"
-version = "0.3.5"
+version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d0c62115964e08cb8039170eb33c1d0e2388a256930279edca206fff675f82c3"
+checksum = "bd5256b483761cd23699d0da46cc6fd2ee3be420bbe6d020ae4a091e70b7e9fd"
 
 [[package]]
 name = "hex-literal"
@@ -1736,9 +1736,9 @@ dependencies = [
 
 [[package]]
 name = "indexmap"
-version = "2.2.2"
+version = "2.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "824b2ae422412366ba479e8111fd301f7b5faece8149317bb81925979a53f520"
+checksum = "233cf39063f058ea2caae4091bf4a3ef70a653afbc026f5c4a4135d114e3c177"
 dependencies = [
  "equivalent",
  "hashbrown",
@@ -3375,9 +3375,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-pki-types"
-version = "1.2.0"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0a716eb65e3158e90e17cd93d855216e27bde02745ab842f2cab4a39dba1bacf"
+checksum = "048a63e5b3ac996d78d402940b5fa47973d2d080c6c6fffa1d0f19c4445310b7"
 
 [[package]]
 name = "rustls-webpki"
@@ -4195,9 +4195,9 @@ dependencies = [
 
 [[package]]
 name = "toml_edit"
-version = "0.22.4"
+version = "0.22.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c9ffdf896f8daaabf9b66ba8e77ea1ed5ed0f72821b398aba62352e95062951"
+checksum = "99e68c159e8f5ba8a28c4eb7b0c0c190d77bb479047ca713270048145a9ad28a"
 dependencies = [
  "indexmap",
  "serde",
@@ -4987,9 +4987,9 @@ checksum = "dff9641d1cd4be8d1a070daf9e3773c5f67e78b4d9d42263020c057706765c04"
 
 [[package]]
 name = "winnow"
-version = "0.5.39"
+version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5389a154b01683d28c77f8f68f49dea75f0a4da32557a58f68ee51ebba472d29"
+checksum = "6b1dbce9e90e5404c5a52ed82b1d13fc8cfbdad85033b6f57546ffd1265f8451"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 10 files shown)


crates/network-scanner/src/mdns.rs
```diff
@@ -0,0 +1,86 @@
...
+    let receiver = service_deamon.browse(META_QUERY)?;
+    let service_deamon_clone = service_deamon.clone();
+
+    let (result_sender, result_receiver) = tokio::sync::mpsc::channel(255);
+
+    task_manager
+        .with_timeout(entire_duration)
+        .when_finished(move || {
+            tracing::debug!("mdns meta query finished");
+            while let Err(e) = service_deamon_clone.stop_browse(META_QUERY) {
```
irvingoujAtDevolution, line 25: using retry as indicated in it's doc
```
/// Stops searching for a specific service type.
    ///
    /// When an error is returned, the caller should retry only when
    /// the error is `Error::Again`, otherwise should log and move on.
    pub fn stop_browse(&self, ty_domain: &str) -> Result<()> {
        self.send_cmd(Command::StopBrowse(ty_domain.to_string()))
    }
```

pauldumais (approved):
Wow this is awesome, LGTM

Cargo.lock
```diff
@@ -1978,6 +1999,19 @@ dependencies = [
...
+name = "mdns-sd"
+version = "0.10.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1961985d2b8f6764565cfe77916f7926aabb6b377cfb5ddf191f7e34136b547f"
+dependencies = [
+ "flume",
+ "if-addrs",
+ "log",
+ "polling 2.8.0",
+ "socket2 0.4.10",
```
CBenoit, line 2012: issue (non-blocking): This crate is used outdated dependencies, and this causes dependency duplication. Do you think you could send a PR for updating them soonish? It’s okay to wait until you are done with implementing the feature first. If you don’t do it now, can you create a Jira ticket to not forget? Thank you.

crates/network-scanner/src/lib.rs
```diff
@@ -23,6 +24,9 @@ pub enum ScannerError {
     #[error("Network interface does not have a netmask")]
     InterfaceDoesNotHaveNetmask(Addr),
 
+    #[error("Network interface does not have a broadcast address")]
+    MdnsError(#[from] mdns_sd::Error),
```
CBenoit, line 28: question: Is the error message accurate? I don’t know mdns that much, but couldn’t the library return an error for many other reasons unrelated to the broadcast address?

crates/network-scanner/src/mdns.rs
```diff
@@ -0,0 +1,86 @@
...
+                if let mdns_sd::Error::Again = e {
+                    continue;
+                }
+                tracing::error!("mdns stop browse error: {}", e);
+            }
```
CBenoit, line 30: issue: The loop logic seems erroneous. My understanding is that we should _break_ when the variant is not `Again`, and otherwise the `continue` keyword is not even required since this is a `while` loop.

nitpick: Use tracing structured logging, especially emit an `error` field for errors.

```suggestion
            while let Err(e) = service_deamon_clone.stop_browse(META_QUERY) {
                match e {
                    mdns_sd::Error::Again => {
                        tracing::trace!("mdns stop_browse transient error, trying again");
                    }
                    fatal => {
                        tracing::error!(error = %fatal, "mdns stop_browse fatal error");
                        break;
                    }
                }
            }
```

crates/network-scanner/examples/mdns.rs
```diff
@@ -0,0 +1,26 @@
...
+use network_scanner::{mdns, task_utils::TaskManager};
+
+#[tokio::main]
+pub async fn main() -> anyhow::Result<()> {
+    tracing_subscriber::fmt::SubscriberBuilder::default()
+        .with_max_level(tracing::Level::INFO)
+        .with_thread_names(true)
+        .init();
+
+    let deamon = mdns_sd::ServiceDaemon::new()?;
```
CBenoit, line 12: suggestion: Could you hide the `mdns_sd` dependency behind our own crate? I.e.: make sure that the user code of our `network-scanner` crate does not need to depend on `mdns_sd` directly. This means that the example should not call `mdns_sd` directly.

rationale: Easier to use API, and less dependencies to declare (and forget) in consumer crate (i.e.: `devolutions-gateway`).

crates/network-scanner/src/mdns.rs
```diff
@@ -0,0 +1,86 @@
...
+                task_manager
+                    .with_timeout(single_query_duration)
+                    .when_finished(move || {
+                        tracing::debug!("mdns query finished for {}", fullname_clone);
+                        while let Err(e) = service_deamon_clone.stop_browse(&fullname_clone) {
+                            if let mdns_sd::Error::Again = e {
+                                continue;
+                            }
+                            tracing::error!("mdns stop browse error: {}", e);
+                        }
```
CBenoit, line 62: issue: Same problem as mentioned previously.

crates/network-scanner/src/scanner.rs
```diff
@@ -299,11 +341,13 @@ impl Display for ScanMethod {
...
-    pub max_wait_time: u64,     // max_wait for entire scan duration in milliseconds, suggested!
+    pub ping_interval: u64,             // in milliseconds
+    pub ping_timeout: u64,              // in milliseconds
+    pub broadcast_timeout: u64,         // in milliseconds
+    pub port_scan_timeout: u64,         // in milliseconds
+    pub netbios_timeout: u64,           // in milliseconds
+    pub netbios_interval: u64,          // in milliseconds
+    pub mdns_meta_query_timeout: u64,   // in milliseconds
+    pub mdns_single_query_timeout: u64, // in milliseconds
+    pub max_wait_time: u64,             // max_wait for entire scan duration in milliseconds, suggested!
```
CBenoit, line 352: thought: We should probably use the `Duration` type instead. Should be addressed separately.

CBenoit (changes requested):
Nice!

CBenoit:
> Open for review but needs to wait for the https://github.com/Devolutions/devolutions-gateway/pull/689 to be merged first

Converted to draft to make it obvious.

Cargo.lock
irvingoujAtDevolution: https://github.com/keepsimple1/mdns-sd/pull/172

irvingoujAtDevolution: created PR to mdsn-sd

crates/network-scanner/src/mdns.rs
irvingoujAtDevolution: updated

crates/network-scanner/examples/mdns.rs
irvingoujAtDevolution: updated

crates/network-scanner/src/lib.rs

crates/network-scanner/src/lib.rs
```diff
@@ -23,6 +24,9 @@ pub enum ScannerError {
     #[error("Network interface does not have a netmask")]
     InterfaceDoesNotHaveNetmask(Addr),
 
+    #[error("Mdns scan error: {0}")]
```
CBenoit, line 27:
```suggestion
    #[error("mDNS scan error: {0}")]
```

crates/network-scanner/src/mdns.rs
```diff
@@ -0,0 +1,110 @@
...
+                            tracing::debug!(sub_service_event=?response);
+                            if let ServiceEvent::ServiceResolved(info) = response {
+                                let server = info.get_fullname();
+                                let port = info.get_port();
+                                let ip = info.get_addresses();
+
+                                for ip in ip {
+                                    let ip = *ip;
+                                    let server = server.to_string();
+                                    let _ = result_sender.send((ip, Some(server), port)).await;
```
CBenoit, line 100: issue: If `send` fails, we should stop the scanning by breaking the loop, because it means the receiver is closed and whatever result we get from now will be ignored.

Note that you’ll need to [break the **outer** loop by using a label](https://doc.rust-lang.org/rust-by-example/flow_control/loop/nested.html#nesting-and-labels).

devolutions-gateway/src/api/mod.rs
```diff
@@ -21,7 +21,7 @@ pub fn make_router<S>(state: crate::DgwState) -> axum::Router<S> {
         .nest("/jet/jrec", jrec::make_router(state.clone()))
         .nest("/jet/config", config::make_router(state.clone()))
         .nest("/jet/session", session::make_router(state.clone()))
-        .nest("/jet/sessions", sessions::make_router(state.clone()))
+        .nest("/jet/sessions", session::make_router(state.clone()))
```
CBenoit, line 24: issue: Any reason for this? This looks very problematic (creating the wrong router).

CBenoit (commented):
Thank you! Looks mostly good. I spotted two other issues which should be fixed, and then we can merge.

devolutions-gateway/src/api/mod.rs
irvingoujAtDevolution: this is a mistake, thank you for point it out!

irvingoujAtDevolution: updated

crates/network-scanner/src/mdns.rs
irvingoujAtDevolution: updated

crates/network-scanner/src/mdns.rs
```diff
@@ -0,0 +1,111 @@
...
+                                    break 'outer;
```
CBenoit, line 101: issue: You probably meant this:
```suggestion
                                    if let Err(_) = result_sender.send((ip, Some(server), port)).await {
                                        break 'outer;
                                    }
```

crates/network-scanner/src/mdns.rs
irvingoujAtDevolution: updated

CBenoit (approved):
LGTM! :rocket:

irvingoujAtDevolution:
> LGTM! 🚀

Merci!

PR #695, opened 2024-02-13
feat(installer): localization, refactoring and flow changes

thenextman:
### UI Flow
The Windows installer UI flow is tweaked; the listeners dialog is now presented _before_ the access URI configuration. This is more logical because it lets us to take the default HTTP listener port for the access URI (while also allowing a power user to continue customizing). It's now possible to configure an HTTPS listener behind an HTTP access URI which doesn't make much sense, but trying to choose an HTTP access URI has a clear callout that this is not recommended for production environments.

### Localization
The installer is now fully localizable: all UI strings come from localization files.

I pulled the localizations into a .json file for each culture, this is easier to parse and work with that xml/wxl. A custom T4 text template processes the localizations as follows:

- The primary culture is en-US
- Each non-primary localization is diff'd against the primary, and a "DevolutionsGateway_{culture}_missing.json" file is generated, containing any missing localizations (for passing to the translator)
- Any missing strings for that culture are filled by their values from the primary culture
- Individual .wxl files are generated for each culture
- A `Strings.cs` file is generated with the constant IDs and comment hints for all the primary culture strings. An extension method on `MsiRuntime` allows for straightforward, error free localization of dynamic strings in .cs sources.

### Other
Small tweaks and refactoring throughout. Notably, the generated installer properties (`GatewayProperties`) are now strongly-typed.


CBenoit (approved):
Happy to see this!! :rocket:

PR #696, opened 2024-02-14
feat(installer): installer revision based on feedback

thenextman:
Fixes an important installer bug (the customization UI should only be shown on a first install, not on upgrades).

Additionally, incorporated feedback items:

- The Access URI hostname will default to the FQDN of the machine if it's domain-joined (rather than just the NetBIOS name)
- Added an "Enter search text here" cue banner on the system certificate selection page
- Default the Access URI protocol to whatever was chosen for the HTTP listener. The user can still customize.
- If configuring as a "companion" install (i.e. no ngrok, no standalone web application) provide a hint linking to the Devolutions Server/Hub documentation on the encryption keys page ("Find your public key for [Devolutions Server](https://docs.devolutions.net/server/dgw/server-configuration/) or [Devolutions Hub](https://docs.devolutions.net/hub/dgw/hub-business-configuration/install-manually/)")
- Provide a help callout when configuring an external certificate, explaining the supported file formats

### DIFF
--- a/package/WindowsManaged/Dialogs/AccessUriDialog.cs
+++ b/package/WindowsManaged/Dialogs/AccessUriDialog.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Net;
 using DevolutionsGateway.Actions;
 using DevolutionsGateway.Dialogs;
 using DevolutionsGateway.Helpers;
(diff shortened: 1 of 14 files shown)


awakecoding (approved):

CBenoit (approved):
Nice improvements!

PR #697, opened 2024-02-14
fix(dgw): rename type for network scan to ap(application protocol)

irvingoujAtDevolution:
rename type ty to ap, stay consistent with rest of the gateway

### DIFF
--- a/devolutions-gateway/src/api/network_scan.rs
+++ b/devolutions-gateway/src/api/network_scan.rs
@@ -105,14 +105,14 @@ impl From<NetworkScanQueryParams> for NetworkScannerParams {
 pub struct NetworkScanResponse {
     pub ip: IpAddr,
     pub hostname: Option<String>,
-    pub ap: ApplicationProtocol,
+    pub protocol: ApplicationProtocol,
 }
 
 impl NetworkScanResponse {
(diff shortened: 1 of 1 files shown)


devolutions-gateway/src/api/network_scan.rs
```diff
@@ -105,15 +105,14 @@ impl From<NetworkScanQueryParams> for NetworkScannerParams {
 pub struct NetworkScanResponse {
     pub ip: IpAddr,
     pub hostname: Option<String>,
-    #[serde(rename = "type")]
-    pub ty: ApplicationProtocol,
+    pub ap: ApplicationProtocol,
```
CBenoit, line 108: Copying my comment from Slack for visibility:

I was really suggesting to use `protocol` or `application_protocol`, not `ap` which is used as an abbreviation in the token claim jet_ap  (because JWT claims are abbreviated by convention).
Here, it’s okay to not use the abbreviation since it’s not a token claim.

I think that `protocol` is a good compromise if you prefer to avoid the longer `application_protocol` naming.

irvingoujAtDevolution: use `protocol` instead, `application_protocol` is a bit to verbose

CBenoit (approved):
Thank you!

PR #698, opened 2024-02-14
feat(dgw): add Windows and Linux functions to extract machine's network adapter interfaces

irvingoujAtDevolution:
As requested by Paul, this could be useful later


crates/network-scanner/src/interfaces/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+pub use windows::get_network_interfaces;
+
+#[cfg(target_os = "linux")]
+pub use linux::get_network_interfaces;
+
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
irvingoujAtDevolution, line 16: serialize to be used later in gateway

CBenoit: I would prefer we don’t reuse this struct as a "API model" (think [DTO](https://en.wikipedia.org/wiki/Data_transfer_object)), and more generally that we don’t define this kind of "system boundary API" too far away from the actual system boundary, in this case, the HTTP endpoint of the Devolutions Gateway.

Instead, I suggest you treat this struct as an internal/business model for the `network-scanner` crate, and then later define a struct specifically to serve as a "API model" directly in the `devolutions_gateway::api::network_scan` module.

This way, you don’t have to think about the system boundaries of the Devolutions Gateway when working on this crate. The [rules at the boundaries are different](https://www.tedinski.com/2018/02/06/system-boundaries.html). Let’s leave serialization concerns where they belong.

crates/network-scanner/src/interfaces/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+pub struct NetworkInterface {
+    name: String,
+    description: Option<String>,
+    mac_address: Vec<[u8; 6]>,
```
CBenoit, line 20: issue: There are two main formats for mac addresses. EUI-48 (6 bytes) and EUI-64 (8 bytes). This structure is ignoring the latter.

crates/network-scanner/src/interfaces/windows.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+impl From<ipconfig::Adapter> for NetworkInterface {
+    fn from(adapter: ipconfig::Adapter) -> Self {
+        NetworkInterface {
+            name: adapter.adapter_name().to_string(),
+            description: Some(adapter.description().to_string()),
+            mac_address: adapter
+                .physical_address()
+                .map(|mac| vec![[mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]]])
```
CBenoit, line 21: issue: This is ignoring the existence of the 8-byte mac addresses. I believe that’s why the library is not returning a `[u8; 6]` directly.

crates/network-scanner/src/interfaces/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    ipv4_address: Vec<Ipv4Addr>,
+    ipv6_address: Vec<Ipv6Addr>,
```
CBenoit, line 22: question: What is the benefit of using two different collections instead of a single `Vec<IpAddr>`?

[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/network-scanner/src/interfaces/windows.rs
```diff
@@ -0,0 +1,45 @@
...
+                .ip_addresses()
+                .iter()
+                .filter_map(|ip| match ip {
+                    IpAddr::V6(ipv6) => Some(*ipv6),
+                    _ => None,
+                })
+                .collect(),
+            prefixes: adapter.prefixes().to_vec(),
+            operational_status: adapter.oper_status() == ipconfig::OperStatus::IfOperStatusUp,
+            default_gateway: adapter.gateways().to_vec(),
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/network-scanner/src/interfaces/mod.rs
irvingoujAtDevolution: updated

crates/network-scanner/src/interfaces/mod.rs
irvingoujAtDevolution: updated, you are right!

crates/network-scanner/src/interfaces/windows.rs
irvingoujAtDevolution: updated!

crates/network-scanner/src/interfaces/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    prefixes: Vec<(IpAddr, u32)>,
+    operational_status: bool,
+    default_gateway: Vec<IpAddr>,
```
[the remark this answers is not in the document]
irvingoujAtDevolution: yes, it is!

crates/network-scanner/src/interfaces/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+    dns_servers: Vec<IpAddr>,
+}
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated!

crates/network-scanner/src/interfaces/mod.rs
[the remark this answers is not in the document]
irvingoujAtDevolution: yes it is, but not common, one example is [MAC address spoofing](https://en.wikipedia.org/wiki/MAC_spoofing)

crates/network-scanner/src/interfaces/mod.rs
irvingoujAtDevolution: understand, updated!

crates/network-scanner/src/interfaces/mod.rs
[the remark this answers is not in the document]
CBenoit: Correct me if I'm wrong, but if you spoof your MAC address, you are "changing" the MAC address of the network interface. You are not adding an additional MAC address. As a result, you still only have a single MAC address, not two. Of course, a single device can have many MAC addresses. One per network interface (be it virtual or physical). But each network interface doesn't have more than one MAC address at a time, even if it can be changed at will. Can you double check this?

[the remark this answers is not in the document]
irvingoujAtDevolution: you are right, one network interface at the same time can only have one mac address, will update.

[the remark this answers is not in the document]
irvingoujAtDevolution: updated

crates/network-scanner/Cargo.toml
```diff
@@ -17,10 +17,19 @@ network-scanner-proto ={ path = "../network-scanner-proto" }
...
-tokio = { version = "1.36.0", features = ["rt", "sync", "time"] }
+tokio = { version = "1.36.0", features = ["rt", "sync", "time", "fs"] }
 tracing = "0.1.40"
 typed-builder = "0.18.1"
 
+[target.'cfg(target_os = "windows")'.dependencies]
+ipconfig = "0.3.2"
+
+[target.'cfg(target_os = "linux")'.dependencies]
+futures = "0.3.30"
```
CBenoit, line 28: suggestion: You can use `futures-util` instead. This is a smaller dependency.

crates/network-scanner/src/interfaces/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+            _ => Err(anyhow::anyhow!("invalid MAC address length")),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct NetworkInterface {
+    pub name: String,
+    pub description: Option<String>,
+    pub mac_addresses: Option<MacAddr>,
```
CBenoit, line 44: naming: `mac_address` without the plural, since there is a single MAC address.

crates/network-scanner/Cargo.toml

crates/network-scanner/src/interfaces/mod.rs
irvingoujAtDevolution: updated

crates/network-scanner/Cargo.toml
```diff
@@ -17,10 +17,19 @@ network-scanner-proto ={ path = "../network-scanner-proto" }
...
 typed-builder = "0.18.1"
 
+[target.'cfg(target_os = "windows")'.dependencies]
+ipconfig = "0.3.2"
+
+[target.'cfg(target_os = "linux")'.dependencies]
+futures-util = "0.3.30"
+netlink-packet-route = "0.19.0"
+rtnetlink = "0.14.1"
+
```
CBenoit, line 31:
```suggestion
```

CBenoit (approved):

PR #701, opened 2024-02-14
fix(dgw): add ARD to common ports to scan

irvingoujAtDevolution:
Need to fix the fact that the service indentification is currently based on port. For mdns, the service type is given, so no need to do port to service mapping anymore. but that is a TODO in the future. 
For this PR, we just add the port 3283 to common ports.

edit: #702 allows service type obtained by mdns to be passed to the caller


CBenoit (approved):
LGTM

PR #705, opened 2024-02-16
feat(dgw): generate netscan token for webapp

pauldumais:
Added the ability to generate netscan tokens for the webapp.
I also removed the "destination" field from the returned truple as with the netscan it's not relevant, hopefully that is ok, it was only used for logging.

### DIFF
--- a/devolutions-gateway/src/api/webapp.rs
+++ b/devolutions-gateway/src/api/webapp.rs
@@ -330,7 +330,7 @@ pub(crate) async fn sign_session_token(
     let now = time::OffsetDateTime::now_utc().unix_timestamp();
     let exp = now + i64::try_from(lifetime).map_err(HttpError::internal().err())?;
 
-    let (claims, content_type) = match req.content_type {
+    let (claims, content_type, destination) = match req.content_type {
         SessionTokenContentType::Association {
             protocol,
             destination,
@@ -359,6 +359,7 @@ pub(crate) async fn sign_session_token(
             })
             .map_err(HttpError::internal().with_msg("ASSOCIATION claims").err())?,
             ContentType::Association,
+            Some(destination),
         ),
 
         SessionTokenContentType::Jmux {
@@ -384,6 +385,7 @@ pub(crate) async fn sign_session_token(
             })
             .map_err(HttpError::internal().with_msg("JMUX claims").err())?,
             ContentType::Jmux,
+            Some(destination),
         ),
 
         SessionTokenContentType::Kdc { krb_realm, krb_kdc } => (
@@ -402,6 +404,7 @@ pub(crate) async fn sign_session_token(
             })
             .map_err(HttpError::internal().with_msg("KDC claims").err())?,
             ContentType::Kdc,
+            Some(krb_kdc),
         ),
 
         SessionTokenContentType::NetScan {} => (
@@ -423,6 +426,7 @@ pub(crate) async fn sign_session_token(
             })
             .map_err(HttpError::internal().with_msg("Netscan claims").err())?,
             ContentType::NetScan,
+            None,
         ),
     };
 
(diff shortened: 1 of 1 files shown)


devolutions-gateway/src/api/webapp.rs
```diff
@@ -248,6 +248,7 @@ pub(crate) enum SessionTokenContentType {
         /// Unique ID for this session
         session_id: Uuid,
     },
+    NetScan {},
```
CBenoit, line 251: style: Define unit-like variant when no fields are required (idiom)
```suggestion
    NetScan,
```

devolutions-gateway/src/api/webapp.rs
```diff
@@ -413,7 +434,6 @@ pub(crate) async fn sign_session_token(
 
     info!(
         user = web_app_token.sub,
-        %destination,
```
CBenoit, line 416: issue: We need to log the destination for auditing purposes.

CBenoit (commented):
I’m not sure that the `sign_session_token` route is the best place for generating NETSCAN tokens. As implied by the name, "session" tokens are used to open RDP, VNC, etc remote sessions. This only includes ASSOCIATION, JMUX and KDC.
Should we really consider NETSCAN tokens as session tokens too? Or is it best to define a new route? More on this later.

The other question I have is: do we really want to allow anyone with a WEBAPP token to perform network scanning? It wasn’t planned to have network scanning in the standalone web application for the V1, and we didn’t discuss adding permissions yet. I suggest waiting for the next release.

> I also removed the "destination" field from the returned tuple as with the netscan it's not relevant, hopefully that is ok, it was only used for logging.

Indeed, it’s not relevant for NETSCAN tokens, but for other token types we need to keep the destination in the logs for auditing purpose.

This may be one other reason to use a separate route.
In the future we may also want to generate SCOPE tokens in order to perform a few administrative tasks such as opening the (future) recording player or listing and killing running sessions. I was thinking of adding an endpoint dedicated to these more powerful tokens. (But nothing has been designed yet.)

If we are okay with generating NETSCAN tokens using the `sign-session-token` route for the time being, you can optionally log the destination like this:

```diff
index c0a54b6..2f2def7 100644
--- a/devolutions-gateway/src/api/webapp.rs
+++ b/devolutions-gateway/src/api/webapp.rs
@@ -330,7 +330,7 @@ pub(crate) async fn sign_session_token(
     let now = time::OffsetDateTime::now_utc().unix_timestamp();
     let exp = now + i64::try_from(lifetime).map_err(HttpError::internal().err())?;

-    let (claims, content_type) = match req.content_type {
+    let (claims, content_type, destination) = match req.content_type {
         SessionTokenContentType::Association {
             protocol,
             destination,
@@ -359,6 +359,7 @@ pub(crate) async fn sign_session_token(
             })
             .map_err(HttpError::internal().with_msg("ASSOCIATION claims").err())?,
             ContentType::Association,
+            Some(destination),
         ),

         SessionTokenContentType::Jmux {
@@ -384,6 +385,7 @@ pub(crate) async fn sign_session_token(
             })
             .map_err(HttpError::internal().with_msg("JMUX claims").err())?,
             ContentType::Jmux,
+            Some(destination),
         ),

         SessionTokenContentType::Kdc { krb_realm, krb_kdc } => (
@@ -402,6 +404,7 @@ pub(crate) async fn sign_session_token(
             })
             .map_err(HttpError::internal().with_msg("KDC claims").err())?,
             ContentType::Kdc,
+            Some(krb_kdc),
         ),

         SessionTokenContentType::NetScan {} => (
@@ -423,6 +426,7 @@ pub(crate) async fn sign_session_token(
             })
             .map_err(HttpError::internal().with_msg("Netscan claims").err())?,
             ContentType::NetScan,
+            None,
         ),
     };

@@ -432,12 +436,22 @@ pub(crate) async fn sign_session_token(
         .encode(provisioner_key)
         .map_err(HttpError::internal().with_msg("sign session token").err())?;

-    info!(
-        user = web_app_token.sub,
-        lifetime,
-        %content_type,
-        "Granted a session token"
-    );
+    if let Some(destination) = destination {
+        info!(
+            user = web_app_token.sub,
+            lifetime,
+            %content_type,
+            %destination,
+            "Granted a session token"
+        );
+    } else {
+        info!(
+            user = web_app_token.sub,
+            lifetime,
+            %content_type,
+            "Granted a session token"
+        );
+    }

     let cache_control = TypedHeader(headers::CacheControl::new().with_no_cache().with_no_store());
```

devolutions-gateway/src/api/webapp.rs
```diff
@@ -401,7 +404,29 @@ pub(crate) async fn sign_session_token(
             })
             .map_err(HttpError::internal().with_msg("KDC claims").err())?,
             ContentType::Kdc,
-            krb_kdc,
+            Some(krb_kdc),
+        ),
+
+        SessionTokenContentType::NetScan {} => (
```
CBenoit, line 410:
```suggestion
        SessionTokenContentType::NetScan => (
```

CBenoit (approved):
LGTM

PR #707, opened 2024-02-16
build(dgw): eliminate openssl link dependency on Linux

awakecoding:
OpenSSL was still a link dependency brought in by reqwest in Linux. We build for Linux on Ubuntu 20.04 and expect it to be forward-compatible, but this is not possible with OpenSSL, as the library is different and incompatible, especially since OpenSSL 1.1 is now in its EOL and Ubuntu 22.04 has moved on to OpenSSL 3.

Our last .deb package release somehow declares a dependency on libssl, but I'm not sure where it's coming from. Maybe we already removed it from the latest master branch and somehow re-introduced the OpenSSL link dependency by mistake. In any case, I at least updated the declared distro name to "focal" (20.04) since it was still saying "xenial" (16.04).

@CBenoit one last thing I want to make sure is that by switching to rustls for reqwest that we still validate against the system CA bundle. I recall it was now an option for rustls, but can you make sure we've enabled it properly?

### DIFF
--- a/devolutions-gateway/Cargo.toml
+++ b/devolutions-gateway/Cargo.toml
@@ -64,7 +64,7 @@ tracing-appender = "0.2"
 # Async, futures…
 tokio = { version = "1.36", features = ["signal", "net", "io-util", "time", "rt", "rt-multi-thread", "sync", "macros", "parking_lot", "fs"] }
 tokio-rustls = { version = "0.24", features = ["dangerous_configuration", "tls12"] }
-reqwest = { version = "0.11", features = ["json"] } # TODO: directly use hyper in subscriber module
+reqwest = { version = "0.11", default-features = false, features = ["rustls-tls", "json"] } # TODO: directly use hyper in subscriber module
 futures = "0.3"
 async-trait = "0.1"
 tower = "0.4"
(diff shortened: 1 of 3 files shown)


CBenoit (approved):
Good catch!

devolutions-gateway/Cargo.toml
```diff
@@ -64,7 +64,7 @@ tracing-appender = "0.2"
 # Async, futures…
 tokio = { version = "1.36", features = ["signal", "net", "io-util", "time", "rt", "rt-multi-thread", "sync", "macros", "parking_lot", "fs"] }
 tokio-rustls = { version = "0.24", features = ["dangerous_configuration", "tls12"] }
-reqwest = { version = "0.11", default-features = false, features = ["rustls-tls", "json"] } # TODO: directly use hyper in subscriber module
+reqwest = { version = "0.11", default-features = false, features = ["rustls-tls-native-roots", "json"] } # TODO: directly use hyper in subscriber module
```
CBenoit, line 67: > One last thing I want to make sure is that by switching to rustls for reqwest that we still validate against the system CA bundle. I recall it was now an option for rustls, but can you make sure we've enabled it properly?

We need to enable the `rustls-tls-native-roots`. I pushed a patch for that, it should be good to go!

PR #706, opened 2024-02-16
feat(dgw): add endpoint for network interface config

irvingoujAtDevolution:
waiting for #698 to be merged first.

Added api endpoint for ipconfig, reuse network token type. It is not necessary to create a new token type specifically for this.

PR #709, opened 2024-02-19
fix(dgw): /jet/network-scan -> /jet/net/scan

CBenoit:
As mentioned here: https://github.com/Devolutions/devolutions-gateway/pull/706#discussion_r1493775706

If the intention is to have multiple endpoints related to network exploration, let’s nest everything under `/jet/net/`:

- `GET /jet/net/scan`
- `GET /jet/net/config`
- …

I want to make this change before we cut the release because it’s harder to change later because of our forward compatibility requirements.


pauldumais (commented):
This has already been in the RDMW beta for a few weeks, but I think we can make the change, as there are more beta builds coming.  @CBenoit Let me know when this gets merged and I can make the corresponding change in RDMW.

pauldumais (approved):

PR #714, opened 2024-02-21
feat(installer): additional localization strings

thenextman:
Simply update the French localization with missing strings


CBenoit (approved):
LGTM!

PR #715, opened 2024-02-21
fix(dgw): polish scanner, fix issues

irvingoujAtDevolution:
fixes the folloing issue
1. this should stop error message from mdns_sd
2. scanner should be stopped correctly now, let me know if on your side @CBenoit 

Issue: DGW-131


crates/network-scanner/examples/scan.rs
```diff
@@ -33,15 +34,21 @@ fn main() -> anyhow::Result<()> {
         let stream = scanner.start()?;
         let stream_clone = stream.clone();
         let now = std::time::Instant::now();
-        while let Ok(Some(res)) = stream_clone
-            .recv_timeout(Duration::from_secs(120))
+
+        tokio::task::spawn_blocking(move || {
```
irvingoujAtDevolution, line 38: type anything to stop the scanning

crates/network-scanner/src/mdns.rs
```diff
@@ -19,10 +19,34 @@ impl MdnsDaemon {
...
+
+    pub fn stop(&self) {
+        let receiver = match self.service_daemon.shutdown() {
+            Ok(receiver) => receiver,
+            Err(e) => {
+                // if e is try again, we should try again, but only once
+                let result = if matches!(e, mdns_sd::Error::Again) {
+                    self.service_daemon.shutdown()
+                } else {
+                    tracing::error!(error = %e, "Failed to shutdown service daemon");
```
CBenoit, line 31: The macro is in scope
```suggestion
                    error!(error = %e, "Failed to shutdown service daemon");
```

crates/network-scanner/src/mdns.rs
```diff
@@ -19,10 +19,34 @@ impl MdnsDaemon {
...
+                    Err(e)
+                };
+
+                let Ok(receiver) = result else {
+                    warn!("Failed to shutdown service daemon");
+                    return;
+                };
```
CBenoit, line 38: The error message is not logged

crates/network-scanner/src/mdns.rs
```diff
@@ -19,10 +19,34 @@ impl MdnsDaemon {
...
+
+                receiver
+            }
+        };
+
+        let _ = receiver.recv_timeout(std::time::Duration::from_millis(100));
```
CBenoit, line 44: Can you add a comment explaining the purpose of this?

crates/network-scanner/examples/scan.rs
CBenoit: I recommend this instead: https://docs.rs/tokio/latest/tokio/signal/fn.ctrl_c.html

CBenoit:
It’s now closing but not in a clean way. Everything should be good if you add this line:

```diff
index 440d2f5..4b58f17 100644
--- a/devolutions-gateway/src/api/net.rs
+++ b/devolutions-gateway/src/api/net.rs
@@ -74,6 +74,8 @@ pub async fn handle_network_scan(
         info!("Network scan finished");

         stream.stop();
+
+        let _ = websocket.close().await;
     });

     Ok(res)
```

irvingoujAtDevolution:
> It’s now closing but not in a clean way. Everything should be good if you add this line:
> 
> ```diff
> index 440d2f5..4b58f17 100644
> --- a/devolutions-gateway/src/api/net.rs
> +++ b/devolutions-gateway/src/api/net.rs
> @@ -74,6 +74,8 @@ pub async fn handle_network_scan(
>          info!("Network scan finished");
> 
>          stream.stop();
> +
> +        let _ = websocket.close().await;
>      });
> 
>      Ok(res)
> ```

I added this at where the stream receives none

crates/network-scanner/examples/scan.rs
irvingoujAtDevolution: updated

crates/network-scanner/src/mdns.rs
irvingoujAtDevolution: updated

crates/network-scanner/src/mdns.rs
irvingoujAtDevolution: added

CBenoit:
> I added this at where the stream receives none

Why not just keep this:

```rust
stream.stop();
let _ = websocket.close().await;
```

At the end? This seems less error-prone. Is it causing any problem?

devolutions-gateway/src/api/net.rs
```diff
@@ -44,6 +44,12 @@ pub async fn handle_network_scan(
             tokio::select! {
                 result = stream.recv() => {
                     let Some(entry) = result else {
+                        let _ = websocket
+                            .send(Message::Close(Some(axum::extract::ws::CloseFrame {
+                                code: axum::extract::ws::close_code::NORMAL,
+                                reason: std::borrow::Cow::from("network scan finished successfully"),
```
CBenoit, line 50: This is nice, but I don’t think we ever use or even print the reason message on client side. So I would rather keep a straightforward code flow.

irvingoujAtDevolution:
> > I added this at where the stream receives none
> 
> Why not just keep this at the end?
> 
> ```rust
> stream.stop();
> 
> let _ = websocket.close().await;
> ```
> 
> This looks less error-prone to me. Is it causing any problem?

Not really, it‘s just the logic to me. looks more precise, if the select ends because of the second branch, where the client sends close message, it feels not right to send back a message on a closed connection. Though, it practially makes little difference.

crates/network-scanner/Cargo.toml
```diff
@@ -31,5 +31,5 @@ netlink-packet-route = "0.19.0"
 rtnetlink = "0.14.1"
 
 [dev-dependencies]
-tokio = { version = "1.36.0", features = ["rt","macros","rt-multi-thread","tracing"] }
+tokio = { version = "1.36.0", features = ["rt","macros","rt-multi-thread","tracing","signal"] }
```
CBenoit, line 34:
```suggestion
tokio = { version = "1.36.0", features = ["rt", "macros", "rt-multi-thread", "tracing", "signal"] }
```

crates/network-scanner/src/mdns.rs
```diff
@@ -19,10 +19,35 @@ impl MdnsDaemon {
...
+        };
+
+        // Receive the last event (Shutdown), preventing the receiver from being dropped, avoid logging an error from the sender side(the mdns crate)
+        let _ = receiver.recv_timeout(std::time::Duration::from_millis(100));
+    }
 }
 
-const SERVICE_TYPES_INTERESTED: [ServiceType; 11] = [
-    ServiceType::Ard,
+// Ard is the same as Vnc
```
CBenoit, line 49:
```suggestion
// ARD is a variant of the RFB (VNC) protocol
```

crates/network-scanner/src/mdns.rs
```diff
@@ -129,9 +158,8 @@ impl TryFrom<&str> for ServiceType {
             "_telnet._tcp" => Ok(ServiceType::Telnet),
             "_ldap._tcp" => Ok(ServiceType::Ldap),
             "_ldaps._tcp" => Ok(ServiceType::Ldaps),
-            // https://jonathanmumm.com/tech-it/mdns-bonjour-bible-common-service-strings-for-various-vendors/
-            // OSX Screen Sharing
-            "_rfb._tcp" => Ok(ServiceType::Ard),
+            // Ard is a special type of Vnc
```
CBenoit, line 161:
```suggestion
            // ARD is a variant of RFB (VNC) protocol.
```

CBenoit (approved):

PR #718, opened 2024-02-22
fix(dgw): better support for ngrok free plan

CBenoit:
Our installer is allowing the 0.0.0.0/0 CIDR by default because premium plans need the IP restrictions to be configured or just all external traffic. However this doesn’t play well with the free plan. This patch is using a dirty trick to detect the free plan and ignores the IP restriction configuration when it is detected.

Issue: DGW-134

### DIFF
--- a/devolutions-gateway/src/ngrok.rs
+++ b/devolutions-gateway/src/ngrok.rs
@@ -47,7 +47,8 @@ impl NgrokSession {
         Ok(Self { inner: session })
     }
 
-    pub fn configure_endpoint(&self, name: &str, conf: &NgrokTunnelConf) -> NgrokTunnel {
+    // FIXME: Make this function non-async again when the ngrok UX issue is fixed.
+    pub async fn configure_endpoint(&self, name: &str, conf: &NgrokTunnelConf) -> NgrokTunnel {
         use ngrok::config::Scheme;
 
         match conf {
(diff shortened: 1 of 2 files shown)


devolutions-gateway/src/ngrok.rs
```diff
@@ -102,6 +105,25 @@ impl NgrokSession {
...
+                    // https://ngrok.com/docs/errors/err_ngrok_9017/
+                    // "Your account is not authorized to use ip restrictions."
+                    Err(ngrok::session::RpcError::Response(e)) if e.error_code.as_deref() == Some("ERR_NGROK_9017") => {
+                        info!("Detected a ngrok free plan subscribtion. IP restriction rules are disabled.");
+
+                        // Revert the builder to before applying the CIDR rules.
+                        builder = before_cidrs;
+                    }
+                    _ => {}
+                }
```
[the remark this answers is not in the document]
awakecoding: that should do it for now - we should open an issue on ngrok-rs asking for a better way to handle this though

devolutions-gateway/src/ngrok.rs
```diff
@@ -102,6 +105,25 @@ impl NgrokSession {
                     .iter()
                     .fold(builder, |builder, cidr| builder.deny_cidr(cidr));
 
+                // HACK: Find the subcribtion plan. This uses ngrok-rs internal API, so it’s not great.
```
thenextman, line 108: subcribtion > "subscription" (2x)

CBenoit: Good catch. Addressed by https://github.com/Devolutions/devolutions-gateway/pull/720

PR #726, opened 2024-02-27
DGW-140-Standalone-Passwords-are-shown-in-clear-text-in-the-DevTool-Console

kristahouse:
- Removed all the console.log entries in the .ts code, including the one that shows connection parameters.
- Turned off the debugwasm & verbose for IronRDP and IronVNC.

### DIFF
--- a/webapp/src/client/app/modules/web-client/ard/web-client-ard.component.html
+++ b/webapp/src/client/app/modules/web-client/ard/web-client-ard.component.html
@@ -67,6 +67,6 @@
     </div>
   </div>
 
-  <iron-remote-gui-vnc #ironGuiElementVnc debugwasm="OFF" verbose="false" scale="fit" flexcenter="true"></iron-remote-gui-vnc>
+  <iron-remote-gui-vnc #ironGuiElementVnc scale="fit" flexcenter="true"></iron-remote-gui-vnc>
 
 </div>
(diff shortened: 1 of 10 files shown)


webapp/src/client/app/modules/web-client/ard/web-client-ard.component.html
```diff
@@ -67,6 +67,6 @@
     </div>
   </div>
 
-  <iron-remote-gui-vnc #ironGuiElementVnc debugwasm="INFO" verbose="true" scale="fit" flexcenter="true"></iron-remote-gui-vnc>
+  <iron-remote-gui-vnc #ironGuiElementVnc debugwasm="OFF" verbose="false" scale="fit" flexcenter="true"></iron-remote-gui-vnc>
```
ngirot-devolutions, line 70: Parameters is not already OFF and false by default ?
If it's the case you can just remove them from the declaration.

webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.html
```diff
@@ -67,6 +67,6 @@
     </div>
   </div>
 
-  <iron-remote-gui-vnc #ironGuiElementVnc debugwasm="INFO" verbose="true" scale="fit" flexcenter="true"></iron-remote-gui-vnc>
+  <iron-remote-gui-vnc #ironGuiElementVnc scale="fit" flexcenter="true"></iron-remote-gui-vnc>
```
CBenoit, line 70: Good enough for now, but I would like an option to enable the logging back in the UI in the future. It’s useful when troubleshooting problems deeper in our implementations.

CBenoit (approved):
LGTM!

PR #727, opened 2024-02-27
DGW-143-Standalone-Web-App-UI-Issues

kristahouse:
Made changes to Web Form:
- fixed the host autocomplete dropdown from being cut off in it's container. 
This also fixed the issue with scroll bars appearing.

- Added the tooltips for Telnet and SSH dropdown.

- Also fixed in the same code, an un-reported issue with the protocol dropdown showing the wrong tooltip when there was an error displayed above the web form.

Changes in the sidebar:
- When a long host name is used it was cut off. Added the tooltip ellipsis to manage these cases.


CBenoit (approved):
LGTM! Thank you!

PR #728, opened 2024-02-28
fix(webapp): update SSH and Telnet packages

kristahouse:
Update Irving's SSH & Telnet packages that handles use cases when the hostname is incorrect, losing the connection to Gateway and closing a session elegantly.


CBenoit (approved):
LGTM!

PR #729, opened 2024-02-28
doc: updated fontscdn link

dion-gionet:


kristahouse (changes requested):
Both files give me a http 400 error.

dion-gionet:
@kristahouse It was due to a mistake on my end which should be fixed now, however https://cdnfonts.devolutions.net/Roboto/static/Roboto.css seems to give me a different issue now, I'll look into it

dion-gionet:
@kristahouse fixed now!

kristahouse (approved):
LGTM!

PR #735, opened 2024-02-29
fix(webapp): authentication list state is not preserved on error

kristahouse:
Small update to preserve the authentication state upon error.


CBenoit (approved):

CBenoit:
image

FYI, we can link the commit and the ticket by adding this footer in the commit description.
It’s good to have it for the changelog.

PR #736, opened 2024-02-29
fix(webapp): the “fullscreen setting” is applied to VNC sessions even though the option is not shown there

kristahouse:
Fixed the Web form to submit the only the required controls for each protocol type.

Also increased the size of the Web form container.


webapp/src/client/app/modules/web-client/form/web-client-form.component.scss
```diff
@@ -44,7 +44,7 @@
 
 .box {
   background-color: var(--box-background-color);
-  width: 377px;
+  width: 447px;
```
CBenoit, line 47: note: Even if this change is fairly small, it should be a separate PR/commit with an appropriate description so that it can be properly tracked and logged in our changelog.

CBenoit (approved):
LGTM!

PR #739, opened 2024-03-01
fix(webapp): improve the error catching for VNC and ARD

kristahouse:
Issue: DGW-157

### DIFF
--- a/webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
+++ b/webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
@@ -380,7 +380,13 @@ export class WebClientVncComponent extends WebClientBaseComponent implements  On
   }
 
   private getMessage(errorData: UserIronRdpError | string): string {
-    const errorKind: UserIronRdpErrorKind = this.getErrorKind(errorData);
+    let errorKind: UserIronRdpErrorKind = UserIronRdpErrorKind.General;
+
+    if (typeof errorData === 'string') {
+      console.error(errorData);
+    } else {
+      errorKind = errorData.kind();
+    }
 
     //For translation 'UnknownError'
     //For translation 'ConnectionErrorPleaseVerifyYourConnectionSettings'
@@ -397,33 +403,5 @@ export class WebClientVncComponent extends WebClientBaseComponent implements  On
       default:
         return 'Connection error: Please verify your connection settings.';
     }
-  }
-
-  private getErrorKind(errorData: UserIronRdpError | string): UserIronRdpErrorKind {
-    if (typeof errorData === 'string') {
-      return this.parseErrorString(errorData);
-    }
-
-    if (errorData?.kind() !== UserIronRdpErrorKind.General) {
-      return errorData.kind();
-    }
-
-    console.error('Error Kind:', errorData.kind());
-    const backtrace: string = errorData?.backtrace();
-    return this.parseErrorString(backtrace);
-  }
-
-  private parseErrorString(error: string): UserIronRdpErrorKind {
-    console.error(error);
-
-    if (error.indexOf('authentication rejected') >= 0) {
-      return UserIronRdpErrorKind.WrongPassword;
-    }
-
-    if (error.indexOf('connection has been rejected') >= 0) {
-      return UserIronRdpErrorKind.LogonFailure;
-    }
-
-    return UserIronRdpErrorKind.General;
   }
 }
--- a/webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
+++ b/webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
@@ -355,7 +355,7 @@ export class WebClientVncComponent extends WebClientBaseComponent implements  On
   }
 
   private notifyUser(eventType: SessionEventType, errorData: UserIronRdpError | string): void {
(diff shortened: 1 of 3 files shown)


webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
```diff
@@ -400,4 +398,32 @@ export class WebClientVncComponent extends WebClientBaseComponent implements  On
...
+    if (error.indexOf('authentication rejected') >= 0) {
+      return UserIronRdpErrorKind.WrongPassword;
+    }
+
+    if (error.indexOf('connection has been rejected') >= 0) {
+      return UserIronRdpErrorKind.LogonFailure;
+    }
+
+    return UserIronRdpErrorKind.General;
+  }
```
CBenoit, line 428: question: I understand the intention is to parse the backtrace of the error returned by the VNC module. Is that correct? I don’t think this should be a concern for the Devolutions Gateway web UI. The WASM module should return the correct error kind and the consumer code just roll with it. Parsing the error string is very fragile, but I think that in this case we don’t have a choice since that’s the only information returned by the server (no standardized status codes). Nevertheless, the string parsing logic should live in the WASM module.

cc @thenextman , how are we proceeding in FreeVNC? Do you have a list of common strings sent by the major VNC servers that we could reuse?

webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
```diff
@@ -400,4 +398,32 @@ export class WebClientVncComponent extends WebClientBaseComponent implements  On
         return 'Connection error: Please verify your connection settings.';
     }
   }
+
+  private getErrorKind(errorData: UserIronRdpError | string): UserIronRdpErrorKind {
+    if (typeof errorData === 'string') {
+      return this.parseErrorString(errorData);
+    }
```
CBenoit, line 405: question: Does it make sense to parse strings not returned by the WASM module? (I’m not sure of the provenance of the "string errors", but that’s not returned by the WASM module at least.)

webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
```diff
@@ -374,16 +374,14 @@ export class WebClientVncComponent extends WebClientBaseComponent implements  On
...
-    } else {
-      this.clientError = this.getMessage(error.kind());
-    }
+    this.clientError = this.getMessage(error);
 
     this.webSessionService.updateWebSessionIcon(this.webSessionId, WebClientVncComponent.DVL_WARNING_ICON);
   }
 
-  private getMessage(errorKind: UserIronRdpErrorKind): string {
+  private getMessage(errorData: UserIronRdpError | string): string {
```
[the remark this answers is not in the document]
kristahouse: The SessionEvent interface in iron-remove-gui & iron-remove-gui-vnc component has data declared as UserIronRdpError | string

webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
kristahouse: I absolutely understand it is not ideal. The reasoning in parsing the backtrace is a last attempt to give the user any type of an informative error message. 

At the moment, most errors from the components are returning error kind: General. We need the error kind to match up with what the backtrace is giving the webapp. Those parsing phrases above are real case examples of when the error kind is returned as General.

thenextman: @CBenoit No we don't even try. Actually the error handling in FreeVNC is super basic; it's somewhat like the above: allow the consumer to differentiate between a connection failure and an authentication failure. For any more detailed diagnostic in RDM it's necessary to read logs. 

On non-Windows RDM, even RDP error handling is basically non-existent ("an error occurred, read your log file!"). It's a pain point but the related teams don't really see fixing that as a priority 😞

webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
CBenoit: It seems the web component may return strings then, but you don’t have to parse these. It’s likely for web component-specific errors.

webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
CBenoit: @kristahouse I’ll include the parsing in the WASM module directly so you don’t have to do it here.

kristahouse: @CBenoit Perfect. Thank you!

webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
kristahouse: @CBenoit Ok. I will not parse them. Thank you!

CBenoit:
@kristahouse FYI: the parsing logic is now handled by the WASM modules since this PR: https://github.com/Devolutions/devolutions-gateway/pull/744

kristahouse:
Removed parsing logic from web app.

CBenoit (approved):
LGTM!

PR #750, opened 2024-03-06
fix(webapp): shutdown not called when closing from left menu

kristahouse:
DGW-167


CBenoit (approved):
LGTM!

CBenoit:
FYI, for the issue to be properly linked (especially in the changelog), an `Issue` footer should be added as part of the commit. Writing the issue ID alone is not enough.
I modified the squash commit accordingly:
image

PR #753, opened 2024-03-07
fix(installer): properly write ARP InstallLocation on fresh installs

thenextman:
In the following conditions:

- Fresh installation
- Installing without UI (quiet or silent)
- Without passing a custom `INSTALLDIR` to msiexec

The action that sets the ARP (Add/Remove Programs) install location will use the default value for `INSTALLDIR`. However, it's sequenced too early: a default can't be resolved until after __CostFinalize__. In this case, we end up writing an empty string to the install location field.


ForestVincent:
Related DVLS console issue : https://devolutions.atlassian.net/browse/DPS-10342

awakecoding (approved):

PR #757, opened 2024-03-08
feat(webapp): check if a new version is available

irvingoujAtDevolution:
Added tooltip. Changed the interceptor logic, now, it will not add auth header for all request to third party url


CBenoit:
I modified the title of the PR to demonstrate how it should look like in the interest of generating a good changelog:

> feat(webapp): check if a new version is available

- It’s a new feature
- For the web application
- What is the feature in a few words, the user don’t need to know implementation details

If more details should be mentioned, they can be written down in the body of the commit. This will be part of the changelog too.
Try to not include all the intermediate commit messages, because I’ll have to remove this from the changelog manually later.
GitHub automatically includes them when squashing a PR with many commits:
image

This should look like this:
image

CBenoit (approved):
The patch looks good to me, but maybe @kristahouse can have a deeper look?
My comments above are mostly on the form. Feel free to merge, but try to follow the convention to simply the future releases. Thank you!

irvingoujAtDevolution:
> The patch looks good to me, but maybe @kristahouse can have a deeper look? My comments above are mostly on the form. Feel free to merge, but try to follow the convention to ease the future releases. Thank you!

Will do, I used github cli that automatically fills up the title, will pay extra attention on it next time.

kristahouse:
Just to follow up. The patch LGTM! Thanks @irvingoujAtDevolution!

PR #768, opened 2024-03-14
ci(dgw): add arm64 build for Linux

thenextman:
Add an arm64 build for Devolutions Gateway on Linux. Currently the binary is not packaged in a .deb nor is a container created. This will follow after the next round of updates to .deb packaging.


CBenoit (approved):
LGTM! I’ll use the `ci` type directly instead of `fix`. This differentiates from actual features / bug fixes part of Devolutions Gateway / Jetsocat.

PR #769, opened 2024-03-14
feat(webapp): Netscan UI

irvingoujAtDevolution:
[DGW-119](https://devolutions.atlassian.net/browse/DGW-119)

Added Net Scan UI for gateway, please see the attached video.

One thing to notice is that, the UI could be better, and it is debatable on how we should present it. This will kept as draft until we agree on how we should present our user this feature on the UI.

https://github.com/Devolutions/devolutions-gateway/assets/139169536/7e0af6df-1522-4729-9f5f-f2e210380206

[DGW-119]: https://devolutions.atlassian.net/browse/DGW-119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -375,6 +375,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
+name = "base64"
+version = "0.22.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


irvingoujAtDevolution:
> > One thing to notice is that, the UI could be better, and it is debatable on how we should present it. This will kept as draft until we agree on how we should present our user this feature on the UI.
> 
> We should submit this to the UX team.

UX team will send us a design by next week.

PR #783, opened 2024-03-22
ci: fix artifact idempotency

thenextman:
Fixes issues with immutable artifacts when running the overall "create new release" workflow. Additionally, adds a "dry-run" option to that workflow and does some small cleanup (inputs from both `workflow_dispatch` and `workflow_call` are now both accessible through the `inputs` context).


CBenoit (approved):
Nice!

awakecoding (approved):

PR #789, opened 2024-04-01
ci: fix packaging of web client

thenextman:
The package workflow has a sequencing error: we need to `tar` the web client _after_ repackaging the Windows Installer; otherwise the tarball gets packaged instead of the directory.


CBenoit (approved):
LGTM!

PR #796, opened 2024-04-03
feat(webapp): add ssh key authentication

irvingoujAtDevolution:
demo video:

https://github.com/Devolutions/devolutions-gateway/assets/139169536/6dae002b-2b21-4c1e-8316-61a8bc95ef1a

I tried to use PrimeNG file upload component, but it does not suit our need.
Please let me know if by using extraData attribute a good idea when creating new session.

### DIFF
--- a/webapp/src/client/app/modules/web-client/form/form-controls/file-control/file-control.component.html
+++ b/webapp/src/client/app/modules/web-client/form/form-controls/file-control/file-control.component.html
@@ -1,0 +1,16 @@
+<div>
+    <label for="SSH Privatekey">Private Key</label>
+    <p-fileUpload name="ssh-private-key" maxFileSize="10000" (onSelect)="onSelect($event)" customUpload="true"
+        [multiple]="false" [showUploadButton]="false" [showCancelButton]="false">
+        <ng-template file pTemplate="file">
+            <div class="file-container">
+                <div class="file-name">{{uploadedFile.name}}</div>
+                <div class="file-size">{{uploadedFile.size / 1024}} KB</div>
+            </div>
+        </ng-template>
+    </p-fileUpload>
+    <div class="gateway-form-input password-container">
+        <div class="form-helper-text" *ngIf="!isValidFile()">
+            {{getErrorMessage()}}
+        </div>
+</div>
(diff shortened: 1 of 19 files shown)


webapp/src/client/app/shared/services/web-form.service.ts
```diff
@@ -73,4 +86,16 @@ export class WebFormService extends BaseComponent {
...
+
+  public canConnect(form: FormGroup): boolean {
+    return form.valid && this.canConnectExtraCallback();
+  }
+
+  canConnectIfAlsoTrue(callback: () => boolean): void {
+    this.canConnectExtraCallback = () => callback();
+  }
+
+  resetCanConnectCallback(){
```
kristahouse, line 98: I'll be picky here and ask to add a space after the parathesis 🧐 Everything else looks great!
`resetCanConnectCallback() {`

kristahouse:
Code looks good. Nice job @irvingoujAtDevolution! 

I will add Paul's comments so we can keep @CBenoit in the loop:

Following the UI in DVLS for the ssh key: We usually display the key in a text box, after the user uploads the key. The text box could also allow the user to type the key manually or paste it in quickly.

CBenoit:
> I will add Paul's comments so we can keep @CBenoit in the loop:
> 
> Following the UI in DVLS for the ssh key: We usually display the key in a text box, after the user uploads the key. The text box could also allow the user to type the key manually or paste it in quickly.

Thank you! This is a sensible UX. Code-wise, I have nothing more to add!

irvingoujAtDevolution:
https://github.com/Devolutions/devolutions-gateway/assets/139169536/0e0f1c05-521a-4049-9284-db2913e6049a

updated as described

webapp/src/client/app/modules/web-client/form/form-controls/file-control/file-control.component.html
```diff
@@ -0,0 +1,46 @@
...
+            <span>
+                <p-button
+                    type="button"
+                    [label]="'clear'"
+                    (onClick)="clearPublicKeyData()"
+                >
+                </p-button>
+                <p-button
+                    type="button"
+                    [label]="'upload'"
```
CBenoit, line 25:
```suggestion
                    [label]="'Upload'"
```

irvingoujAtDevolution:
> Out of curiosity, is it still possible to drag and drop the file?

No, it is not yet possible.

CBenoit:
Do you think you could check with Paul or Krista if this should be implemented?

irvingoujAtDevolution:
> Do you think you could check with Paul or Krista if this should be implemented?

sure, that's a good idea.

irvingoujAtDevolution:
Uploading GW final versio private key.mp4…

PR #805, opened 2024-04-09
ci: restore deb package OneDrive upload in release

thenextman:


thenextman:
@CBenoit #799 was the problem, it wasn't just renaming the .deb but also moving it to the "upload" directory 🤦

CBenoit (approved):
Oops!! 🤦‍♂️ 
Thank you for the fix!

PR #807, opened 2024-04-15
perf(dgw): use a buffer of 1k bytes for ARD VNC sessions

RRRadicalEdward:
_ARD_ uses _MVS_ video codec which doesn't like buffering, and we need to have the buffer as minimal as possible.

Also, this commit adds new `copy_bidirectional` transport that is forked [the one from tokio](https://docs.rs/tokio/latest/tokio/io/fn.copy_bidirectional.html). It's forked because the original function doesn't allow overriding the buffer size (8K is used by default). There is [an issue](https://github.com/tokio-rs/tokio/issues/6454) on tokio side for it. We will be able to replace our fork with the upstream easily when it's ready.

Releated to https://github.com/Devolutions/IronVNC/issues/338.


RRRadicalEdward:
I needed to fork DevolutionsGateway because I don't have the right to push directly to the repo.

RRRadicalEdward:
It's easy to add `copy_bidirectional` with the buffer size in _tokio_ directly, but we haven't agreed on what the naming should be. If you have any thoughts, I would appreciate 😄 - https://github.com/tokio-rs/tokio/issues/6454.

crates/transport/src/copy_bidirectional.rs
```diff
@@ -0,0 +1,123 @@
+//! Fork of https://github.com/tokio-rs/tokio/blob/master/tokio/src/io/util/copy.rs to allow us set
+//! variable length `CopyBuffer` size instead of default 8k.
```
CBenoit, line 2:
```suggestion
//! Vendored code from https://github.com/tokio-rs/tokio/blob/master/tokio/src/io/util/copy.rs
//! It is modified to allow us setting the `CopyBuffer` size instead of hardcoding 8k.
```

crates/transport/src/copy_buffer.rs
CBenoit, line 1: suggestion: To reduce the number of "vendored" files, can you keep that in the same file as `copy_bidirectional`?

RRRadicalEdward: Yes, sure

CBenoit (approved):
LGTM!! Thank you!

CBenoit:
For changelog:

Apple ARD uses the so-called MVS video codec.
It is a tricky codec: Apple didn't implement proper congestion control, so it's basically just TCP controlling the flow (not by much).
Our MVS implementation for the web client is obviously not as fast as the native one, and can’t keep up when there are too much data in transit.
To reduce the amount of data in transit, we reduced the size of the copy buffer when using web socket forwarding endpoint and if the application protocol of the session is set to ARD.

Issue: DGW-138

RRRadicalEdward:
> For changelog:

> Apple ARD uses the so-called MVS video codec.
> It is a tricky codec: Apple didn't implement proper congestion control, so it's basically just TCP controlling the flow (not by much).
> Our MVS implementation for the web client is obviously not as fast as the native one, and can’t keep up when there are > too much data in transit.
> To reduce the amount of data in transit, we reduced the size of the copy buffer when using web socket forwarding endpoint and if the application protocol of the session is set to ARD.

> Issue: DGW-138

Perfect 👌

crates/transport/src/copy_bidirectional.rs
```diff
@@ -0,0 +1,249 @@
+//! Vendored code from https://github.com/tokio-rs/tokio/blob/1f6fc55917f971791d76dc91cce795e656c0e0d3/tokio/src/io/util/copy.rs
```
CBenoit, line 1:
```suggestion
//! Vendored code from:
//! - https://github.com/tokio-rs/tokio/blob/1f6fc55917f971791d76dc91cce795e656c0e0d3/tokio/src/io/util/copy.rs
//! - https://github.com/tokio-rs/tokio/blob/1f6fc55917f971791d76dc91cce795e656c0e0d3/tokio/src/io/util/copy_bidirectional.rs
```

CBenoit:
I’m sorry. It appears we can’t merge PRs from forks because of the new CI job which requires secrets (devolutions-gateway-web-ui). I’ll merge the code with this PR instead: https://github.com/Devolutions/devolutions-gateway/pull/809

PR #809, opened 2024-04-15
perf(dgw): use a buffer of 1k bytes for ARD VNC sessions

CBenoit:
Apple ARD uses the so-called MVS video codec.
It is a tricky codec: Apple didn't implement proper congestion control, so it's basically just TCP controlling the flow (not by much).
Our MVS implementation for the web client is obviously not as fast as the native one, and can’t keep up when there are too much data in transit.
To reduce the amount of data in transit, we reduced the size of the copy buffer when using web socket forwarding endpoint and if the application protocol of the session is set to ARD.

Issue: DGW-138


pacmancoder (approved):
LGTM!

RRRadicalEdward:
`copy_bidirectional_with_sizes` was added today: https://github.com/tokio-rs/tokio/pull/6500. 
We will be able to use in the next release. Stay tuned 😁

PR #813, opened 2024-04-16
feat(dgw): add analytics to gateway standalone

irvingoujAtDevolution:
Enable analytics on Gateway Standalone with OpenSearch.

Place to pay attention to:
There is really no elegant solution to use system environment variables in angular. These env vars are hardcoded  in the environment file as suggested by Angular documentation. We could argue that it will be visible to our user in production anyway so it is not a bit deal if we expose it here.

Updated some functions that was originally not used.

### DIFF
--- a/webapp/src/client/app/shared/services/analytic.service.ts
+++ b/webapp/src/client/app/shared/services/analytic.service.ts
@@ -23,7 +23,7 @@ export class AnalyticService {
     });
   }
 
-  public sendOpenEvent(connectionType: ProtocolString): ConnectionIndentifier {
+  public sendOpenEvent(connectionType: ProtocolString): ConnectionIdentifier {
     this.sendEvent({
       connectionType: connectionType,
     });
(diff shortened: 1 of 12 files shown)


webapp/src/client/app/shared/services/analytic.service.ts
```diff
@@ -0,0 +1,147 @@
...
+        .catch((error) => {
+          if (isDevMode()) {
+            console.error('Error sending event', error);
+          }
+        });
+    });
+  }
+}
+
+export interface ConnectionIndentifier {
```
pauldumais, line 123: Typo "Identifier" remove the 2nd letter "n".

irvingoujAtDevolution: updated!

kristahouse (approved):
LGTM

pauldumais (approved):

pauldumais (approved):
LGTM

CBenoit:
note: The scope for this PR/commit should be `webapp` instead of `dgw` (should be adjusted when redacting the release notes, cc @thenextman)

PR #814, opened 2024-04-17
chore(openapi): regenerate clients

thenextman:
The key change (apart from bumping the client versions) is to [scope the .NET client to just .NET 6](https://github.com/Devolutions/devolutions-gateway/compare/update-open-api?expand=1#diff-babdabf2f45c7d80d96bf7be45141c678d139781bf5dd2e1f50c653d4e44c082R13) (removing the netstandard 2.1 target).

Targeting netstandard at all doesn't make much sense, since we don't care to support the old .NET core or Xamarin et al. Importantly, however, it triggers the code generation to give us a [broader default exception factory](https://github.com/Devolutions/devolutions-gateway/compare/update-open-api?expand=1#diff-0b87b424239868a357eec793b18723144347ac15a981a95f136ad7ffcbea5c77R62). 

This will cause non-HTTP errors to be raised as exceptions, instead of returned as a null response.


awakecoding (approved):

thenextman:
> Did you modify the generated files manually, without modifying the template? The changes will be overridden next time we generate the client code if we don’t modify the templates themselves

No, I just changed the target framework in the config.json. Removing the netstandard21 target was enough to make it generate a better exception handler (in the template, there's a condition around netstandard - I don't know why - but since the netstandard target is redundant, this works)

thenextman:
@CBenoit https://github.com/OpenAPITools/openapi-generator/blob/807250a4300f3a4fed65f990e7055b34047222e1/modules/openapi-generator/src/main/resources/csharp/Configuration.mustache#L59

CBenoit:
Oh, that’s great then! Thank you!

PR #816, opened 2024-04-18
fix(dgw): error code on service startup failure

CBenoit:
Instead of panicking when failing to start the service, we instead attempt to log the error to the log file and return an error code.

Issue: DGW-174


devolutions-gateway/src/main.rs
```diff
@@ -125,22 +125,36 @@ fn service_controller() -> Controller {
 
 enum GatewayServiceEvent {}
 
+const BAD_CONFIG_ERR_CODE: u32 = 1;
+const START_FAILED_ERR_CODE: u32 = 2;
```
CBenoit, line 129: These are the two status code we now return when failing on startup. When the config is invalid, the logger is never initialized, so we can’t log the exact error, but we know roughly were to look. It’s also possible to run the executable manually in the terminal to get the detailed error. For the other error, the logger will typically be ready (expect if the error is us being unable to setup the logger).

thenextman (approved):
Thanks!

PR #818, opened 2024-04-18
ci: skip web client and packaging in forks

thenextman:
The web client can't be built in a fork, because it requires access to secrets. This also prevents packaging working.

This PR disables those features in different repos. For now that will allow the core CI to run but preclude building changes to the web client, .deb or Windows installer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -279,6 +279,7 @@ jobs:
     name: devolutions-gateway-web-ui
     runs-on: ubuntu-latest
     needs: preflight
+    if: github.repository_owner == 'Devolutions'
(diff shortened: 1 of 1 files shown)


awakecoding (approved):

awakecoding:
small suggestion: use [github.repository_owner](https://docs.github.com/en/actions/learn-github-actions/contexts) to compare with "Devolutions" instead of the full github ref

thenextman:
> small suggestion: use [github.repository_owner](https://docs.github.com/en/actions/learn-github-actions/contexts) to compare with "Devolutions" instead of the full github ref

Done!

PR #832, opened 2024-04-30
feat: initial implementation for WebM player for recordings

nicolesabourin1:
Adds a video and xterm player at the `GET /jet/jrec/play` endpoint which supports multiple videos and builds the page dynamically based on the type of recording.

Issue: DGW-110


webapp/player/index.html
```diff
@@ -0,0 +1,215 @@
+<html>
+  <head>
+      <meta name="viewport" content="width=device-width">
+      <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/xterm-player@latest/dist/css/xterm-player.min.css" />
+      <script src="https://cdn.jsdelivr.net/npm/xterm@4.4.0/lib/xterm.min.js"></script>
```
awakecoding, line 5: @nicolesabourin1 I'm not sure how we currently handle the dependencies for the rest of our web frontend, but we'd need to build and include external dependencies alongside the rest of the frontend such that it can work without Internet access. No external CDN delivery can be used for this, as this causes problems for GDPR compliance in addition to breaking things when used without Internet access

webapp/player/index.html
```diff
@@ -0,0 +1,215 @@
...
+            const arrayBuffer = req.response;
+            if (arrayBuffer) {
+              const byteArray = new Uint8Array(arrayBuffer);
+              onLoad(byteArray);
+            }
+          };
+          req.send(null);
+        }
+
+        function convertTRPtoCast(fileArray) {
```
awakecoding, line 153: we have our answer, Paul baked in a live TRP to .cast converter directly in the Javascript of the player, so it'll be able to play the older TRP file format as well :)

awakecoding (changes requested):

webapp/player/index.html
CBenoit: My understanding is that we should use the `@xterm/xterm` npm package: https://www.npmjs.com/package/@xterm/xterm

Can you confirm? @kristahouse

kristahouse: Yes, you can use ES6 module syntax to import modules in a plain JavaScript setup within an HTML page.

pauldumais: The problem here is we need a way to host the static web files inside the gateway web server for the xterm js and css files, just like we did for the webapp static files, but in this case the webapp may not be enabled,  @CBenoit will need to make an endpoint for the static files.

CBenoit: @pauldumais Just like for the webapp, it’s possible to serve static files via the `GET /jet/jrec/play` endpoint. It serves all the files found in the `player/` folder. Is there anything which should be changed? Do we have different requirements for the player?

pauldumais: @CBenoit awesome, I just added the xterm player files locally to the folder

awakecoding (approved):

PR #844, opened 2024-05-07
fix(webapp):fix ui align with titile or ip is too long

irvingoujAtDevolution:
image

when on hover, a scroll bar horizontally will pop up (if the string is long enough), allow user to view the entire ip address or username


pauldumais (approved):
LGTM

PR #845, opened 2024-05-07
fix(webapp): fix netscan result duplicate and performance improvement

irvingoujAtDevolution:
Before: Netscan request is fired whenever this netscan component is mounted
After: Netscan request only one time per each web session. All results are cached in memory.


pauldumais (approved):
LGTM

PR #846, opened 2024-05-07
fix(webapp): fix version compare logic

irvingoujAtDevolution:
Before:  show version update icon when version  is different, the logic is that user would not be able to install a version we haven't released.

After: strictly check the version, only show icon if the latest version is greater than the current version

Issue: DGW-182


CBenoit (approved):
LGTM!

PR #847, opened 2024-05-08
fix(webapp): fix force rescan button fix force rescan button failure introduced by previous PR in net-scan component now force rescan works correctly

irvingoujAtDevolution:
issue: DGW-185


CBenoit (approved):
Nice! Thank you!

PR #851, opened 2024-05-14
DGW-164-Webapp-Suspicious-e-logs-in-web-application

kristahouse:
Downgraded the Primeng package as the console log was introduced in version 16.5.0.

Upgrading to the next Primeng version would cause a cascade of updates for the majority of the angular dependencies. This includes upgrading the rxjs package, however, in doing so would cause a conflict in the web packages for IronRDP & IronVNC. We should make future plans to perform this update.

Issue: DGW-164


CBenoit (approved):
Thank you for figuring out this!

PR #856, opened 2024-05-14
feat(webapp): allow ssh connection to use encrypted ssh key

irvingoujAtDevolution:
Add SSH encrypted private key authentication method.


CBenoit (approved):
Nice improvement!

PR #859, opened 2024-05-15
fix(pwsh): fix reading .pem files from PowerShell runspace

thenextman:
When we call [Encoding.GetString(byte[])](https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.getstring?view=net-8.0#system-text-encoding-getstring(system-byte())) from a C# PowerShell runspace (e.g. DPS Console), we are getting the following error:

>The following exception occurred while retrieving member "GetString": "GenericArguments[0], 'System.ReadOnlySpan`1[System.Byte]', on 'System.Func`2[T,TResult]' violates the constraint of type 'T'."

I don't know why. This results in an empty `$PemData` which obviously can't be parsed.

The workaround is to read the certificate and private key file directly as UTF8, rather than getting the UTF8 representation of the data.

### DIFF
--- a/powershell/DevolutionsGateway/Private/CertificateHelper.ps1
+++ b/powershell/DevolutionsGateway/Private/CertificateHelper.ps1
@@ -230,28 +230,25 @@ function Get-PemCertificate
             $PrivateKey = $expanded.PrivateKey.ToPem().ToRepr()
         }
     } else {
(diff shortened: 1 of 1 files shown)


CBenoit (approved):
Subtile! Thank you!

awakecoding (approved):

PR #862, opened 2024-05-15
fix(webapp): tooltip for the selected protocol is always set to RDP

kristahouse:
Issue: DGW-187

### DIFF
--- a/webapp/src/client/app/modules/web-client/form/web-client-form.component.ts
+++ b/webapp/src/client/app/modules/web-client/form/web-client-form.component.ts
@@ -139,6 +139,7 @@ export class WebClientFormComponent
             }
           });
 
+          this.updateProtocolTooltip(protocol);
           this.formService.detectFormChanges(this.cdr);
         },
         (error) =>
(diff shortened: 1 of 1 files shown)


webapp/src/client/app/modules/web-client/form/web-client-form.component.ts
```diff
@@ -150,10 +151,8 @@ export class WebClientFormComponent
     this.buildFormGroup(this.inputFormData)
       .pipe(
         takeUntil(this.destroyed$),
-        switchMap((formGroup) => {
+        switchMap((formGroup: FormGroup<any>) => {
```
irvingoujAtDevolution, line 154: Is it possible to use unknown or explicit type here?

kristahouse: I used `<any>` because the form is dynamic. 

I could probably improve the type safety by creating an interface for the form model with all our properties as optional. Thoughts?

```
interface ConnectionFormModel {
  username?: string;
  password?: string;
  ...
}

...

switchMap((formGroup: FormGroup<ConnectionFormModel >) => {
```

irvingoujAtDevolution: This is a bit beyond the scope of this PR. I think we can use the type `unknown` for the scope of this function. As we are not accessing it's properties here.

irvingoujAtDevolution: `unknown` is slightly more preferable over 'any'
https://typescript-eslint.io/rules/no-explicit-any/#:~:text=TypeScript's%20unknown%20is%20almost%20always,t%20yet%20known%20or%20representable

kristahouse: Sounds like a good plan. I'll make an update for this pr tomorrow. 😃

irvingoujAtDevolution (approved):

irvingoujAtDevolution:
Thank you @kristahouse, sorry for nitpicking!

PR #871, opened 2024-05-23
fix(dgw): rework network interface DTO definition

irvingoujAtDevolution:
Rework network interface DTO definition for network scan crates and endpoint. As described in Jira ticket.

Issue: DGW-187


CBenoit (approved):
LGTM

PR #888, opened 2024-06-18
feat(installer): add a basic configuration check

thenextman:
This PR adds a basic configuration check to the installer. We attempt to assess effective file permissions and report any issues to the user.

> [!NOTE]  
> I'd consider these changes quite safe; there is no behaviour change in the rest of the installer; the biggest risk is that we have false positives (inadvertently report non-issues to the user).

> [!NOTE]  
> This PR also makes "Custom" authentication the default option when enabling the standalone web application.

This is extremely tricky to make correct: paths may or may not exist (and it's hard to know if the path doesn't exist, or we simply don't have access), querying Windows access control lists is convoluted - especially when dealing with nested files and folders and inheritance, and we may be dealing with shared network locations that can be hard to read permissions from (share permissions versus file permissions).

These changes make a "best effort" to report issues to the user. We take the simplest approach: impersonate the relevant user (NETWORK SERVICE) and attempt to open the directories and files we're interested in, with the proper permission mask. If issues are detected, the user will see a "View configuration issues" link on the last page of the installer (the location can also be found in the log file, for unattended installs - but invalid configurations will not cause the installation to fail). Clicking the link opens a basic HTML table detailing any issues.

Screenshot 2024-06-17 at 22 27 16

Screenshot 2024-06-17 at 22 27 27

After making corrections, the user can re-run the installer ("Repair") to regenerate the report.

There is a risk of false positives; one example is that log files or recordings might exist on a network share. A failure to access could indicate missing permissions, but it could also be a conflict with another Gateway instance running in a farm and accessing the file(s) at the same moment.

Furthermore, for log files and recordings, we only attempt to access a subset (the 10 oldest) of the interesting files. While logs should be limited by rotation, there could be many hundreds of recordings and we simply don't want to try and open all of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,6 @@ Stable options are:
 
         * **Proto** (_String_): MUST be set to `http`.
         * **Domain** (_String_): The domain to request, as registered in the ngrok dashboard.
-        * **Metadata** (_String_): Arbitrary user-defined metadata that will appear in the ngrok service API when listing tunnel sessions.
         * **CircuitBreaker** (_Ratio_): Reject requests when 5XX responses exceed this ratio.
         * **Compression** (_Boolean_): Enable gzip compression for HTTP responses.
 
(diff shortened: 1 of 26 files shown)


CBenoit (approved):
Thank you!

PR #889, opened 2024-06-18
feat(agent): Devolutions Gateway service updater

pacmancoder:
This PR implements Gateway updater logic in `devolutions-agent`.

This update mechanism is triggered by changing `ProgramData/Devolutions/Agent/update.json` file, setting the required version to update the gateway.

Example `update.json`:
```json
{
    "Gateway": {
        "TargetVersion": "2024.2.1"
    }
}
```

#### Features:
- Feature quite generic, so additional products could be updated later with the same mechanism by extending it
- Updater checks for already installed products on the system via MSI's upgrade code, so no spurious same-version MSI invocation is possible
- Requested Gateway version's MSI is gathered from [devolutions.net ](https://devolutions.net/productinfo.htm)
- `update.json` is created on first service start, and correct DACL is set to allow `NT AUTHORITY/Network Service` (which is used by Gateway) to edit this file
- Downloaded MSI hash is validated
- MSI signature and certificate thumbprint are validated
- All downloaded MSIs, Installation logs temp files are marked for removal after use

PR #890, opened 2024-06-18
feat(gateway): add API to trigger Devolutions Gateway update

pacmancoder:
Gateway side of updater feature started at https://github.com/Devolutions/devolutions-gateway/pull/889

The update could be triggered with the following request:
```
GET /jet/update?version=2024.2.1
```


devolutions-gateway/src/api/update.rs
```diff
@@ -0,0 +1,63 @@
...
+        gateway: Some(ProductUpdateInfo { target_version }),
+    };
+
+    serde_json::to_string(&update_json)
+        .ok()
+        .and_then(|serialized| std::fs::write(updater_file_path, serialized).ok())
+        .ok_or_else(|| {
+            error!("Failed to write new Gateway version to `update.json`");
+            HttpErrorBuilder::new(StatusCode::INTERNAL_SERVER_ERROR).msg("Agent updater service is unavailable")
+        })?;
```
CBenoit, line 60: suggestion: Let’s avoid to kill the source errors. It’s harder to understand what is the underlying problem later.

This is how I recommend to handle this kind of code:
```suggestion
    let update_json = serde_json::to_string(&update_json).map_err(
        HttpError::internal()
            .with_msg("failed to serialize the update manifest")
            .err(),
    )?;

    std::fs::write(updater_file_path, update_json).map_err(
        HttpError::internal()
            .with_msg("failed to write the new `update.json` manifest on disk")
            .err(),
    )?;
```

Note that you don’t need the `error!` statement, errors are logged by one of our middleware upper in the layer.

devolutions-gateway/src/api/update.rs
```diff
@@ -0,0 +1,63 @@
...
+    _scope: UpdateScope,
+) -> Result<Json<UpdateResponse>, HttpError> {
+    let target_version = query.version;
+
+    let updater_file_path = get_updater_file_path();
+
+    if !updater_file_path.exists() {
+        error!("Failed to start Gateway update, `update.json` does not exist (should be created by Devolutions Agent)");
+
+        return Err(HttpErrorBuilder::new(StatusCode::SERVICE_UNAVAILABLE).msg("Agent updater service is unavailable"));
```
CBenoit, line 47: suggestion: It’s enough to just return the error, it will be logged.

```suggestion
        return Err(HttpErrorBuilder::new(StatusCode::SERVICE_UNAVAILABLE).msg("Agent updater service is not installed"));
```

devolutions-gateway/src/api/mod.rs
```diff
@@ -27,7 +28,8 @@ pub fn make_router<S>(state: crate::DgwState) -> axum::Router<S> {
         .route("/jet/rdp", axum::routing::get(rdp::handler))
         .nest("/jet/fwd", fwd::make_router(state.clone()))
         .nest("/jet/webapp", webapp::make_router(state.clone()))
-        .nest("/jet/net", net::make_router(state.clone()));
+        .nest("/jet/net", net::make_router(state.clone()))
+        .route("/jet/update", axum::routing::get(update::start_update));
```
CBenoit, line 32: question: I think this should not be `GET`, but `POST` or `PATCH`. `GET` should be used when fetching some data, not when updating it.

devolutions-gateway/src/api/update.rs
```diff
@@ -0,0 +1,63 @@
...
+        (status = 200, description = "Update request has been processed successfully", body = UpdateResponse),
+        (status = 400, description = "Bad request"),
+        (status = 401, description = "Invalid or missing authorization token"),
+        (status = 403, description = "Insufficient permissions"),
+        (status = 500, description = "Agent updater service is malfunctioning"),
+        (status = 503, description = "Agent updater service is unavailable"),
+    ),
+    security(("scope_token" = ["gateway.update"])),
+))]
+pub(super) async fn start_update(
```
CBenoit, line 36: note: The naming suggests that the update procedure will actually be started by the function, but in fact all we do is updating the `update.json`. I think this could be clarified in the name. It would also be nice to document the logic a little bit.

CBenoit:
@pacmancoder I think you need to rebase on master

pacmancoder:
@CBenoit yep, I'll rebase today and add follow-up comment fixes from https://github.com/Devolutions/devolutions-gateway/pull/889#event-13229471813

CBenoit:
@pacmancoder Oh I didn’t see your message. I just sent follow PRs for these.

pacmancoder:
@CBenoit rebased and made refactoring

devolutions-gateway/openapi/gateway-api.yaml
CBenoit, line 1: For now, leave openapi untouched as-is

CBenoit (commented):
It seems that you need to run cargo format

devolutions-gateway/openapi/gateway-api.yaml
pacmancoder: Hm. should I temporarily comment out `utoipa` attributes in that case?

CBenoit: It’s fine to leave that as-is, let’s just not generate the files yet
It seems a bit buggy

CBenoit (approved):
LGTM!

CBenoit:
@pacmancoder I didn’t notice when merging but FYI the commit scope should be `dgw`.

PR #898, opened 2024-06-21
refactor(agent): updater.disable → updater.enabled

CBenoit:
For consistency between our config files.

### DIFF
--- a/devolutions-agent/src/config.rs
+++ b/devolutions-agent/src/config.rs
@@ -122,12 +122,16 @@ pub fn load_conf_file_or_generate_new() -> anyhow::Result<dto::ConfFile> {
 pub mod dto {
     use super::*;
 
+    pub const fn default_bool<const V: bool>() -> bool {
+        V
+    }
+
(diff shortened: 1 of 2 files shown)


devolutions-agent/src/config.rs
```diff
@@ -122,12 +122,16 @@ pub fn load_conf_file_or_generate_new() -> anyhow::Result<dto::ConfFile> {
...
+        V
+    }
+
     #[derive(PartialEq, Eq, Debug, Default, Clone, Serialize, Deserialize)]
     #[serde(rename_all = "PascalCase")]
     pub struct UpdaterConf {
         /// Enable updater module (enabled by default)
-        #[serde(default)]
-        pub disable: bool,
+        #[serde(default = "default_bool::<true>")]
```
pacmancoder, line 133: AFIK side allows the following syntax without a separate function:
```rust
#[serde(default_value  = "true")]
```

pacmancoder (approved):

PR #899, opened 2024-06-21
fix(dgw): preserve DGW access URI base in recording player

irvingoujAtDevolution:
This needs to be published for side by side installed Gateway to work on recording players.


fdubois1 (approved):
LGTM

PR #912, opened 2024-07-08
feat(agent): initial NOW-PROTO implementation

pacmancoder:
This PR implements `NOW-PROTO` protocol according to the current draft specification.

Additionally, I made some fixes/changes to the specification in https://github.com/Devolutions/devolutions-gateway/pull/913 , please review these changes first before reviewing this PR.

- Pdu parsing/error handling were copied from IronRDP, but code was and simplified/functionality reduced in some places.
- `#[no_std]` is supported with `alloc`, crate itself do not have any dependencies on other workspace members, therefore it could be used as a standalone library for other projects besides devolutions-agent
- bitfields/enums are not strict when parsing, while providing access to unknown flags/variant -- this should improve backwards-compatibility for the protocol in future.
- Some structs are currently missing due to non-specified strucutre in specification (e.g. CMD exec command and System info 
PDUs)
- All structures are covered by simple unit tests, but it would be great to add fuzzing in future

Blocked by #913

### DIFF
--- a/crates/now-proto-pdu/src/cursor.rs
+++ b/crates/now-proto-pdu/src/cursor.rs
@@ -1,0 +1,310 @@
+use crate::PduResult;
+
+#[derive(Debug)]
+pub struct ReadCursor<'a> {
+    inner: &'a [u8],
+    pos: usize,
+}
+
+impl<'a> ReadCursor<'a> {
+    pub const fn new(bytes: &'a [u8]) -> Self {
+        Self { inner: bytes, pos: 0 }
+    }
+
+    pub const fn len(&self) -> usize {
+        self.inner.len() - self.pos
+    }
+
+    pub const fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    pub const fn eof(&self) -> bool {
+        self.is_empty()
+    }
+
+    pub fn remaining(&self) -> &[u8] {
+        &self.inner[self.pos..]
+    }
+
+    pub const fn inner(&self) -> &[u8] {
+        self.inner
+    }
+
+    pub const fn pos(&self) -> usize {
+        self.pos
+    }
+
+    pub fn read_array<const N: usize>(&mut self) -> [u8; N] {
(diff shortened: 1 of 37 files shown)


crates/now-proto-pdu/src/macros.rs
```diff
@@ -0,0 +1,184 @@
+//! Helper macros for PDU encoding and decoding
```
pacmancoder, line 1: Mostly code from IronRDP

crates/now-proto-pdu/src/cursor.rs
```diff
@@ -0,0 +1,310 @@
+use crate::PduResult;
```
pacmancoder, line 1: Mostly code from IronRDP

crates/now-proto-pdu/src/error.rs
```diff
@@ -0,0 +1,93 @@
+//! This module provides [`PduError`] and [`PduErrorKind`] types based on
```
pacmancoder, line 1: Mostly code from IronRDP

CBenoit:
Do you think we could reuse the traits and types from IronRDP directly by adding a dependency on it? We already have the dependency in devolutions-agent: https://github.com/Devolutions/devolutions-gateway/blob/f091443c529494bcaa9d87942356c3c26760e372/devolutions-agent/Cargo.toml#L31-L38
I also want to move the core traits from ironrdp-pdu to a new crate, ironrdp-core. When it’s done, we can get rid of the superfluous PDUs of RDP :slightly_smiling_face:

pacmancoder:
>Do you think we could reuse the traits and types from IronRDP directly by adding a dependency on it? We already have the dependency in devolutions-agent

Well, I simplified some types (e.g. error handling types) for now-proto implementation, but overall, if we could separate out `ironrdp-core` (naming: maybe `ironpdu-core` ?) that would be great!

CBenoit (approved):
Great job! Looking pretty good, I don’t see any outstanding issue. Let’s merge and move on to incremental PRs!

PR #922, opened 2024-07-16
fix(dgw): recording player now scales with the size of the window

nicolesabourin1:
- CSS changes, the recording player now scales with the size of the window

Issue: DGW-198


pauldumais (approved):
LGTM

CBenoit (commented):
Thank you!

PR #923, opened 2024-07-18
ci: prepare for aws-lc-rs

CBenoit:
This patch is preparing the CI for newer rustls which is using aws-lc-rs as the default crypto backend.

- Install a newer version of gcc for Linux. aws-lc-rs is rejecting some versions of gcc, including the one packaged by default with Ubuntu 20.04. It’s important because of this bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189
- Install NASM when building with release / production profiles.
- Run tests using the dev profile. The tlk.ps1 script only allowed release and production profiles, but there is no reason to run the tests using these profiles.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,9 @@ jobs:
 
         - name: Tests
           shell: pwsh
-          run: ./ci/tlk.ps1 test -Platform ${{ matrix.os }} -Architecture ${{ matrix.arch }} -CargoProfile ${{ needs.preflight.outputs.rust-profile }}
+          env:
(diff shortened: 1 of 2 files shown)


awakecoding:
Won't switching the Linux runner to Ubuntu 22.04 potentially cause the aws-lc-rs C components to be built for it instead of our current baseline of Ubuntu 20.04?

CBenoit:
You're right. I'll install a newer toolchain on 20.04 instead

CBenoit:
@awakecoding This should be good now!

awakecoding (approved):

PR #937, opened 2024-07-25
fix(dgw): fix recording player parsing problem

irvingoujAtDevolution:
1. Remove usage of btoa, it fails on different charset
2. Add 1 millisecond if the time of previous event is the same at the next event. Otherwise the player will throw an error

Issue: DPS-11197


pauldumais (approved):
LGTM

PR #938, opened 2024-07-25
chore(webapp): add Biome as linter for webapp

irvingoujAtDevolution:
Added Biome as formatter and linter for Webapp.
This PR only contains change in dependency and configuration files.

### DIFF
--- a/webapp/biome.json
+++ b/webapp/biome.json
@@ -1,0 +1,12 @@
+{
+	"$schema": "https://biomejs.dev/schemas/1.8.3/schema.json",
+	"organizeImports": {
+		"enabled": true
+	},
+	"linter": {
+		"enabled": true,
+		"rules": {
+			"recommended": true
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


webapp/biome.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "https://biomejs.dev/schemas/1.8.2/schema.json",
+  "organizeImports":{
+    "enabled": true
+  },
+  "linter": {
+    "rules": {
+      "style": {
+        "useTemplate": "off",
+        "useImportType": "off"
```
irvingoujAtDevolution, line 10: Import as type prevent Angular dependency injection from working

webapp/biome.json
```diff
@@ -0,0 +1,24 @@
...
+      }
+    }
+  },
+  "formatter": {
```
irvingoujAtDevolution, line 14: the format preferred by DVLS team

CBenoit (approved):
Nice! Looks good to me. Are you intending to fix the lints? If you do, we could start to enforce the checks in the CI.

PR #942, opened 2024-07-26
feat(pedm): add base PEDM implementation

kbouchard-dev:
CI changes:
- ~~Environment variable `DAGENT_PEDM_PFX` should be defined as an absolute path to a code signing certificate to sign the `Appx` package for the right click context menu (Windows 11 requires this to avoid going into the 'Show more options' option)~~
- ~~Environment variable `DAGENT_PEDM_PFX_PASSWORD` should be defined as the password of the certificate. **This is bad**, I would like to discuss other ways of doing it but I lack experience in CI/CD secret handling.~~
- ~~CI/CD .yml files will need to be updated.~~ Done

OpenAPI:
- The OpenAPI spec for the PEDM is generated with [`aide`](https://github.com/tamasfe/aide/). I noticed the repo uses another library and so if we should use only one I'd be glad to change it.
- Previously the script to generate the OpenAPI specs used the PowerShell `>` operator, which will write files in UTF16, which is a bit cumbersome to use. I replaced it with `| Out-File -Encoding UTF8`

Enabling the feature:
- ~~For the moment, you will need to toggle Devolutions PEDM in the installer as well as modify the file `C:\ProgramData\Devolutions\Agent\agent.json` to include `"Pedm":{"Enabled":true}` in the root object.~~
  - ~~To do this, you will need Administrator privileges~~
- Toggle 'PEDM' in the installer.

Known issues:
- ~~The virtual account feature is a bit unstable. Over RDP, applications sometimes entirely fail to show UI, and not with RDP, the applications load an old themed UI. This is probably due to the virtual account not having enough permissions in the actual user's desktop and session. Furthermore, there are probably issues linked to permissions and logging in to third parties because of this different user.~~ Fixed
  - See: https://www.reddit.com/r/Intune/comments/16z2uue/epm_network_share/
  - Worth investigating if this feature is needed
- Currently a process requesting changes to the PEDM configuration with an elevated token will always succeed. This also includes tokens created by the PEDM, and so an elevation can lead to full elevation everywhere. A possible way to fix this issue is "tagging" each token created by the PEDM and rejecting them when trying to modify the settings. This would be done though a token's attributes. However, we must verify that the attributes cannot be changed.
- The PEDM does not currently support elevating to medium integrity or elevating to high integrity. Right now, all elevations are in high integrity.
- The PEDM does not currently support changing behaviors for child processes. This is possible but not implemented yet.
- The `Appinfo` hook is in the code, but not enabled.


pacmancoder:
@thenextman could you please review the wix installer changes in this PR please?

thenextman:
> @thenextman could you please review the wix installer changes in this PR please?

I went through this installer and everything looks good, nice job 🤙 

My only concern is that we seem to be shipping a .NET 8 application but there's no provision for checking that dependency at install time. It does mean that if the runtime is not installed, the user will be prompted to install that the first time it's run which perhaps isn't a great experience (as we've seen with RDM).

I don't see it needing to be fixed as part of this PR, but it should be addressed before this goes into the user hands (whether that be "properly" by applying the dependency via the installer, like RDM by doing it inside the installer but not using Windows Installer features, simply blocking the install until the dependency is there, or something else).

kbouchard-dev:
> > @thenextman could you please review the wix installer changes in this PR please?
> 
> I went through this installer and everything looks good, nice job 🤙
> 
> My only concern is that we seem to be shipping a .NET 8 application but there's no provision for checking that dependency at install time. It does mean that if the runtime is not installed, the user will be prompted to install that the first time it's run which perhaps isn't a great experience (as we've seen with RDM).
> 
> I don't see it needing to be fixed as part of this PR, but it should be addressed before this goes into the user hands (whether that be "properly" by applying the dependency via the installer, like RDM by doing it inside the installer but not using Windows Installer features, simply blocking the install until the dependency is there, or something else).
> 
> Note that this becomes especially important if this will be deployed by MDM (e.g. InTune). The .NET dependency cannot be easily installed by the MDM and should be handled on our side.

Thanks and good catch. As there's nothing too fancy going on I'll see if it's possible to downgrade to .NET Framework as to make it compatible by default.

thenextman:
> > > @thenextman could you please review the wix installer changes in this PR please?
> > 
> > 
> > I went through this installer and everything looks good, nice job 🤙
> > My only concern is that we seem to be shipping a .NET 8 application but there's no provision for checking that dependency at install time. It does mean that if the runtime is not installed, the user will be prompted to install that the first time it's run which perhaps isn't a great experience (as we've seen with RDM).
> > I don't see it needing to be fixed as part of this PR, but it should be addressed before this goes into the user hands (whether that be "properly" by applying the dependency via the installer, like RDM by doing it inside the installer but not using Windows Installer features, simply blocking the install until the dependency is there, or something else).
> > Note that this becomes especially important if this will be deployed by MDM (e.g. InTune). The .NET dependency cannot be easily installed by the MDM and should be handled on our side.
> 
> Thanks and good catch. As there's nothing too fancy going on I'll see if it's possible to downgrade to .NET Framework as to make it compatible by default. For now it would work with RDM, but the error dialog can be shown when a user is not allowed to elevate and it's not guaranteed RDM is installed for that popup.

@kbouchard-dev That would be awesome (downgrade to .NET Framework) if possible. It also avoids issues with multi arch and simplifies the build.

kbouchard-dev:
For the outdated `tower` and `hyper` dependencies in `devolutions-pedm-shared`, this should be fixed with the 7.8.0 release of OpenAPI Generator: https://github.com/OpenAPITools/openapi-generator/pull/19115

pacmancoder:
Just a heads-up, I am currently in the process of reviewing this PR, optimistically, I'll be able to finish it tomorrow

crates/devolutions-pedm/openapi/generate_all.ps1
```diff
@@ -0,0 +1,19 @@
+#!/bin/env pwsh
+
+$ErrorActionPreference = "Stop"
+
+Push-Location -Path $PSScriptRoot
+
+try {
+	../../../tools/generate-openapi/generate.ps1 pedm | Out-Host
```
pacmancoder, line 8: I believe its better to write path relative to script root ($PSScriptRoot variable) isntead of relative path (`../../..`)
cc @awakecoding

crates/devolutions-pedm/src/config.rs
```diff
@@ -0,0 +1,46 @@
+use std::{
+    path::{Path, PathBuf},
+    sync::OnceLock,
+};
+
+use camino::Utf8PathBuf;
+use win_api_wrappers::{raw::Win32::Foundation::LUID, win::Module};
+
+pub static PEDM_DESKTOP_RELPATH: &'static str = r"desktop/DevolutionsPedmDesktop.exe"; // TODO change this
```
pacmancoder, line 9: Is this TODO should be removed or renaming should occur?

crates/devolutions-pedm/src/elevator/local_admin_elevator.rs
```diff
@@ -0,0 +1,75 @@
...
+            sid: Sid::from_well_known(WinLocalAccountAndAdministratorSid, None)?,
+            attributes: (SE_GROUP_ENABLED | SE_GROUP_ENABLED_BY_DEFAULT | SE_GROUP_MANDATORY) as _,
+        });
+
+        groups.0.push(SidAndAttributes {
+            sid: owner_sid.clone(),
+            attributes: (SE_GROUP_ENABLED | SE_GROUP_ENABLED_BY_DEFAULT | SE_GROUP_MANDATORY | SE_GROUP_OWNER) as _,
+        });
+
+        // groups.0.push(SidAndAttributes {
```
pacmancoder, line 48: leftover comment?

crates/devolutions-pedm/src/error.rs
```diff
@@ -0,0 +1,93 @@
...
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use tracing::error;
+use win_api_wrappers::raw::Win32::Foundation::{
+    ERROR_ACCESS_DISABLED_BY_POLICY, ERROR_CANCELLED, ERROR_INVALID_PARAMETER, E_UNEXPECTED,
+};
+
+#[derive(Deserialize, Serialize, Debug, JsonSchema)]
+#[serde(rename_all = "PascalCase")]
+pub enum Error {
```
pacmancoder, line 18: maybe lets use `thiserror` here?

crates/devolutions-pedm/src/utils.rs
```diff
@@ -0,0 +1,161 @@
...
+    Ok(process_information.unwrap())
+}
+
+#[derive(Default)]
+pub struct MultiHasher {
+    sha1: Sha1,
+    sha256: Sha256,
+}
+
+impl MultiHasher {
```
pacmancoder, line 70: Why do we actually need multi hash? is it required for WinAPI calls or for the other reason?

crates/win-api-wrappers/src/error.rs
```diff
@@ -0,0 +1,124 @@
...
+    Win32::{
+        Foundation::{E_POINTER, WIN32_ERROR},
+        System::Rpc::RPC_STATUS,
+    },
+};
+
+use crate::undoc::LSA_SID_NAME_MAPPING_OPERATION_ERROR;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Error {
```
pacmancoder, line 18: Might be good to use `thiserror` here

crates/win-api-wrappers/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+mod error;
+pub use error::Error;
+
+pub mod win;
+
+#[allow(non_camel_case_types, non_snake_case, non_upper_case_globals)]
```
pacmancoder, line 6: Lets add here comment about reason for silencing lints with `allow`

crates/win-api-wrappers/src/rpc.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    pub fn impersonate_client<F>(&self, f: F) -> Result<()>
+    where
+        F: FnOnce() -> Result<()>,
+    {
+        unsafe { RpcImpersonateClient(Some(self.0)) }.ok()?;
+
+        let r = f();
+
+        unsafe { RpcRevertToSelfEx(Some(self.0)) }.ok()?;
```
pacmancoder, line 91: It would be great to wrap this  in some kind of wrapper, which calls RpcRevertToSelfEx on destructor

crates/win-api-wrappers/src/win.rs
```diff
@@ -0,0 +1,3683 @@
+use std::{
```
pacmancoder, line 1: Lets split this file to multiple small modules, single with >3K lines of code is very hard to follow 😢

package/AgentWindowsManaged/Actions/AgentActions.cs
```diff
@@ -194,6 +194,43 @@ internal static class AgentActions
...
+        Impersonate = false,
+        Execute = Execute.deferred,
+        Return = Return.check,
+        Step = Step.InstallFiles,
+        When = When.After,
+        Sequence = Sequence.InstallExecuteSequence,
+        Condition = Condition.BeingUninstalled,
+    };
+
+    // For some reason, when running as NT AUTHORITY\SYSTEM on Windows 11,
```
pacmancoder, line 214: Indeed sounds like a very strange behavior

pacmancoder (commented):
I gave the PR a first review round, it is a very impressive chunk of work!

ci/tlk.ps1
```diff
@@ -329,65 +383,97 @@ class TlkRecipe
...
+                $MakeAppxOutput = & 'MakeAppx.exe' 'pack' '/f' "${CargoPackagePath}/mapping.txt" '/p' "./devolutions-pedm-contextmenu.msix" '/nv' '/o'
+                if (!$?) {
+                    throw "MakeAppx package creation failed: ${MakeAppxOutput}"
+                }
+
+                if (-Not ($(Test-Path $Env:DAGENT_PEDM_PFX) -And $Env:DAGENT_PEDM_PFX_PASSWORD)) {
+                    throw 'DAGENT_PEDM_PFX / DAGENT_PEDM_PFX_PASSWORD was not found in the environment'
+                }
+
+                $SignToolOutput = & 'SignTool.exe' 'sign' '/fd' 'SHA256' '/a' '/f' "${Env:DAGENT_PEDM_PFX}" '/p' "${Env:DAGENT_PEDM_PFX_PASSWORD}" (Get-Item "./devolutions-pedm-contextmenu.msix").FullName
```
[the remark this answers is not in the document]
thenextman: Yes, signing needs to be done in the CI workflow. You can invoke devops to get the proper certificates/secrets made available. 

It's ok to allow signing in tlk for dev/test purposes with a local certificate, but not for releases.

Signing is normally only done in the package workflow since privileged access to the signing information requires the workflow run to be authorized. If it's important that CI artifacts are also signed, we'd need to consider options here (perhaps sign in CI with a dev / low privilege certificate, and then resign in packaging)

ci/tlk.ps1
```diff
@@ -329,65 +383,97 @@ class TlkRecipe
...
+                }
+
+                Pop-Location
             }
 
-            & $StripExecutable $SrcExecutablePath | Out-Host
+            Pop-Location
         }
 
-        if ($DestinationExecutable) {
```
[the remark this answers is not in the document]
thenextman: Minor correction; it's the package workflow that will call tlk. The release workflow just pushes artifacts / creates the release. Bur yes, someone on architecture-maintainers will need to authorize a workflow run.

crates/devolutions-pedm-contextmenu/devolutions-pedm-contextmenu-shell/src/lib.rs
```diff
@@ -136,14 +136,22 @@ fn find_main_explorer(session: u32) -> Option<u32> {
...
-            || proc.token(TOKEN_QUERY).ok()?.session_id().ok()? != session
+            && proc.token(TOKEN_QUERY).ok()?.session_id().ok()? == session
+            && proc
+                .peb()
+                .ok()?
+                .user_process_parameters()
+                .ok()?
+                .command_line
+                .to_command_line()
+                .eq_ignore_ascii_case("C:\\Windows\\Explorer.EXE"))
```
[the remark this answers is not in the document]
thenextman: Viewing this completely out of context, but are we assuming Windows is installed on the C: drive?

crates/devolutions-pedm-shared/devolutions-pedm-client-http/.gitignore
```diff
@@ -0,0 +1,3 @@
+/target/
```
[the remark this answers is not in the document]
kbouchard-dev: This is autogenerated by OpenAPI Generator. We could add a .gitignore for files it generates. I'll see if there are configurations to change its behavior.

crates/devolutions-pedm/src/utils.rs
kbouchard-dev: Not really. It's more of a convenience feature. If the user wants to add a hash check they can select either sha256 or sha1. There's sha1 since it's popular for certificate fingerprints. This is just a helper that computes all the supported hashes as one.

crates/devolutions-pedm-contextmenu/devolutions-pedm-contextmenu-shell/src/lib.rs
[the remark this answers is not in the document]
pacmancoder: @thenextman huh, is it possible to install to non-C drive? 😮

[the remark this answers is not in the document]
thenextman: @pacmancoder It's possible that there isn't even a C drive

One way to know would be to call `SHGetKnownFolderPath` with `FOLDERID_Windows` (in C, at least :))

ci/tlk.ps1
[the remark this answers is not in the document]
CBenoit: I started a dry run on this branch to catch potential errors.

crates/devolutions-pedm-hook/src/hook.rs
```diff
@@ -0,0 +1,146 @@
+use std::path::Path;
+
+use devolutions_pedm_shared::client::{
+    self,
+    models::{LaunchPayload, StartupInfoDto},
+};
+use tracing::error;
+use win_api_wrappers::{
```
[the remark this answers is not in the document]
CBenoit: @kbouchard-dev You can automatically format using this style by running rustfmt with the nightly toolchain.

```bash
cargo +nightly fmt
```

crates/devolutions-pedm-shared/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+schemars = { version = "0.8.21", optional = true }
+windows-registry = { version = "0.2.0", optional = true }
+devolutions-pedm-client-http = { path = "./devolutions-pedm-client-http", optional = true }
+tower = { version = "0.3", optional = true } # old version required for OpenAPI generator
+pin-project = { version = "1.1.5", optional = true }
+
+[features]
+pedm_client = ["hyper", "serde_json", "tokio", "winapi", "policy", "serde", "devolutions-pedm-client-http", "tower", "pin-project", "anyhow"]
+policy = ["regex", "schemars", "serde", "glob"]
+build = []
```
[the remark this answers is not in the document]
pacmancoder: Ah, I am sorry, my comment was a little bit confusing, I wanted to say that instead of empty `build` feature set, there should be an empty `default` feature set, that is more in-line with rest of our code

ci/tlk.ps1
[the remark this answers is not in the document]
CBenoit: @kbouchard-dev 
The CI appears broken from start. Just in case, can you rebase on master and push again?

crates/devolutions-pedm-shared/Cargo.toml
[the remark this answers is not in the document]
CBenoit: The `build` feature is actually for feature-gating a module called `build`, so my understanding is that we can’t replace it with `default` :thinking:

crates/devolutions-pedm-shared/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "devolutions-pedm-shared"
+version = "2024.8.2"
```
CBenoit, line 3: suggestion: Internal crates that we don’t publish (with `publish = false`), are assigned the version `0.0.0` because it doesn’t really matter and we don’t want to keep updating that number at each release.

crates/devolutions-pedm-shared/devolutions-pedm-client-http/.gitignore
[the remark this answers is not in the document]
CBenoit: You can add a .openapi-generator-ignore file for ignoring files generated by the generator that we don’t need: https://github.com/Devolutions/devolutions-gateway/blob/master/devolutions-gateway/openapi/dotnet-client/.openapi-generator-ignore

CBenoit:
I’m not sure why, but the CI appears to not be running on your PR. I tried to remove the draft status in case that’s the cause (it’s probably not a draft anymore anyway).

crates/devolutions-pedm/openapi/generate_all.ps1
CBenoit: It’s fine, because there is a `Push-Location -Path $PSScriptRoot` at the top of the file, ensuring we are always relative to the file.

crates/devolutions-pedm-shared/devolutions-pedm-client-http/src/models/error.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl std::fmt::Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            Self::AccessDenied => write!(f, "AccessDenied"),
+            Self::NotFound => write!(f, "NotFound"),
+            Self::InvalidParameter => write!(f, "InvalidParameter"),
+            Self::InternalError => write!(f, "InternalError"),
+            Self::Cancelled => write!(f, "Cancelled"),
+        }
```
CBenoit, line 37: suggestion: If you only want to output the name of the variant instead of a pretty message, you can simply to this:
```suggestion
        write!(f, "{self:?}")
```

crates/devolutions-pedm/src/error.rs
CBenoit: In this case, I believe the benefit to be too minor to be worth it. This error is useful for deriving the `Display` trait with nice user-facing errors, but we don’t have such error messages here. See [this comment](https://github.com/Devolutions/devolutions-gateway/pull/942/files/0deddb6d97397eba1462f66fc044bff5f7d2e605#r1704158166).

ci/tlk.ps1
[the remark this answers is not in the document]
kbouchard-dev: Just rebased

kbouchard-dev:
CI bug looks like the Agent not being gated behind Windows properly. Will look into it.

CBenoit (commented):
Here is the course of action I suggest:

- Document the unsafe blocks, explaining preconditions and why the usage is correct.
- Document and checkout important knowledge directly in committed README / modules files. This includes things like known issues, future plans, and / or a general overview of the module. I leave you deciding what is relevant, but for instance: what are the two elevation methods, and why we don’t yet use one of them (virtual account elevator).
- Enable the workspace lints: https://github.com/Devolutions/devolutions-gateway/blob/873217c8042167bf378613f812b3db2a399d256a/jetsocat/Cargo.toml#L9-L10
  - If that’s too much work, just enable the "safer unsafe" lints listed here: https://github.com/Devolutions/devolutions-gateway/blob/873217c8042167bf378613f812b3db2a399d256a/Cargo.toml#L31-L59
  - We can help you enabling remaining lints and fixing them in follow up PRs.
- Dry run of `Create New Release` on the branch to see if the workflow is still fine.

I think the PR can be merged once all of this is addressed, and we can follow up on it in separate PRs to help Kyle.

I’ll be on vacation starting tomorrow, can you take care of landing the PR? @pacmancoder

pacmancoder:
@kbouchard-dev ping me when PR will be available for re-review 👀

.github/workflows/package.yml
```diff
@@ -130,17 +145,17 @@ jobs:
...
         project: [ jetsocat, devolutions-gateway, devolutions-agent ]
-        os: [ windows, macos, linux ]
+        os: [ windows, linux ]
         include:
           - os: windows
             runner: windows-2022
-          - os: macos
-            runner: macos-latest
+          #- os: macos
+          #  runner: macos-latest
```
CBenoit, line 158: nitpick: Feel free to remove unused code.

CBenoit:
Package workflow is failed with this error:

image

.github/workflows/package.yml
kbouchard-dev: This was because the MacOS runner in test mode didn't have proper keys for signing. Will update it now now that the agent packages correctly.

rbstp (approved):
CI portion looks good

PR #944, opened 2024-07-29
refactor: enable additional lints

CBenoit:
I took the list that we used for IronRDP and removed a few that I believed didn’t apply.

Can you review this? @pacmancoder

### DIFF
--- a/crates/network-scanner-net/examples/basic_ping.rs
+++ b/crates/network-scanner-net/examples/basic_ping.rs
@@ -1,3 +1,5 @@
+#![allow(unused_crate_dependencies)]
+
 use std::mem::MaybeUninit;
 use std::net::SocketAddr;
 
--- a/crates/network-scanner/src/broadcast/blocking.rs
+++ b/crates/network-scanner/src/broadcast/blocking.rs
@@ -34,6 +34,7 @@ impl Iterator for BorcastBlockStream {
             return None;
         }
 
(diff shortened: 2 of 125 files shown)


crates/network-scanner-net/examples/basic_ping.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused_crate_dependencies)]
```
pacmancoder, line 1: Maybe let's create a separate issue to address removing these kinds of lints later?

crates/network-scanner/src/broadcast/blocking.rs
```diff
@@ -34,6 +34,7 @@ impl Iterator for BorcastBlockStream {
             return None;
         }
 
+        // SAFETY: TODO: explain safety.
```
pacmancoder, line 37: Let's add this to follow-up issue

pacmancoder (approved):
LGTM overall, although I think we need to address unused crate dependencies and TODOs in `SAFETY` comments in the follow-up issue/PR

crates/network-scanner-net/examples/basic_ping.rs
CBenoit: Agreed, my intention is to send another PR to rework the workspace. I’ll follow up on this soonish.

PR #955, opened 2024-07-31
ci(dgw): package XMF native library

CBenoit:


.github/workflows/package.yml
```diff
@@ -240,6 +240,13 @@ jobs:
           $Destination = Join-Path "webapp" "client"
           gh run download ${{ needs.preflight.outputs.run }} -n webapp-client -D "$Destination"
 
+      - name: Download Cadeau
+        shell: pwsh
+        run: |
+          $Platform = @{'windows'='win'; 'linux'='linux'}[${{ matrix.os }}]
+          $Arch = @{'x86_64'='x64';'arm64'='arm64'}[${{ matrix.arch }}]
+          ./ci/download-cadeau.ps1 -Platform $Platform -Architecture $Arch
```
CBenoit, line 248: FIXME: I think it’s not working

package/WindowsManaged/Program.cs
```diff
@@ -100,6 +100,30 @@ private static string DevolutionsWebClientPath
         }
     }
 
+    private static string LibXmfPath
+    {
+        get
+        {
+            string path = Environment.GetEnvironmentVariable("DGATEWAY_LIB_XMF_PATH");
+
+            if (string.IsNullOrEmpty(path) || !Directory.Exists(path))
```
thenextman, line 109: This should be `File.Exists`, not `Directory.Exists`. Probably the reason for your CI error.

package/WindowsManaged/Program.cs
```diff
@@ -259,6 +283,10 @@ static void Main()
                             StopOn = SvcEvent.InstallUninstall,
                         },
                     },
+                    new (LibXmfPath)
+                    {
+                        TargetFileName = Includes.EXECUTABLE_NAME,
```
thenextman, line 288: The constant value is "DevolutionsGateway.exe", so this will not work; since we don't need to reference this elsewhere in the installer you could just put the value directly e.g. ` = "xmf.dll"`

thenextman:
Looks good, I added a couple comments of things that won't work (and one of them probably explains your CI error)

package/WindowsManaged/Program.cs
CBenoit: Nice catch! Copy paste gone wrong! 😂

CBenoit:
Thank you! I’ll give it a try and merge if everything seems alright :slightly_smiling_face:

CBenoit:
I confirmed the DLL is properly packaged and extracted at the right place:

image

PR #957, opened 2024-08-01
fix(dgw)!: jet_rec claim is now a string

CBenoit:
Possible values are:

- `none`: No policy to enforce (recording is optional)
- `external`: An external application (e.g.: RDM) must push the recording stream via a separate websocket connection
- `proxy`: Session must be recorded directly at Devolutions Gateway level (not implemented yet)


awakecoding:
How about "stream" instead of "external"?

awakecoding (approved):

CBenoit:
> How about "stream" instead of "external"?

Sounds good, I’ll change that.

PR #958, opened 2024-08-01
fix(dotnet-utils)!: update jet_rec serialization

CBenoit:


awakecoding:
Same thing: change external to stream

CBenoit:
Got you! Updated.

PR #974, opened 2024-08-14
fix(jetsocat,dgw): allow bigger JMUX messages

CBenoit:
This updates the decoder part of the JMUX codec.

The intention is to increase the maximum message size for JMUX message in the future.
By increasing the maximum size from 4k to 16k, an improvement of about 77.7% can be achieved when measuring using iperf.

With 4k:

> 0.0000-19.4523 sec  26.6 GBytes  11.7 Gbits/sec

With 16k:

> 0.0000-13.8540 sec  33.6 GBytes  20.8 Gbits/sec

### DIFF
--- a/crates/jmux-proxy/src/codec.rs
+++ b/crates/jmux-proxy/src/codec.rs
@@ -4,9 +4,6 @@ use bytes::BytesMut;
 use jmux_proto::{Header, Message};
 use tokio_util::codec::{Decoder, Encoder};
 
-/// This is a purely arbitrary number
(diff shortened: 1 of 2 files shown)


CBenoit:
Compatibility hazard: if a client sends a big message to an old Gateway, an error will be raised. I'll address this a bit later. First, we'll increase the maximum message size for Devolutions Gateway, and then we could increase the maximum size on client side in RDM in early 2025.

awakecoding (approved):

PR #978, opened 2024-08-15
feat(pedm): fix dead pointer in ACL wrapper

kbouchard-dev:


awakecoding (approved):

awakecoding (approved):

CBenoit:
I’ll fix that when generating the changelog later, but it would be best to use `fix` instead of `feat` in your comment message when the patch is a bugfix. We are using https://www.conventionalcommits.org/en/v1.0.0/ to help us with generating the changelog.

PR #990, opened 2024-08-21
feat(pedm): add context menu icon resource

awakecoding:


crates/devolutions-pedm-contextmenu/devolutions-pedm-contextmenu-shell/src/lib_win.rs
```diff
@@ -68,11 +70,24 @@ impl IElevationContextMenuCommand_Impl for ElevationContextMenuCommand_Impl {}
...
+
+        let module_path = match module.file_name() {
+            Ok(path) => path,
+            Err(_) => return Err(E_NOTIMPL.into()),
+        };
+
+        let module_path_str = module_path.to_string_lossy();
+        let icon_path = format!("{},-101", module_path_str); // current dll path + ",-101" (icon resource id)
+        let mut icon_path = WideString::from(icon_path.as_str());
+        unsafe { SHStrDupW(icon_path.as_pwstr()) }
```
pacmancoder, line 90: WinAPI docs tell that `SHStrDupW` allocated memory, which should be freed via `CoTaskMemFree`, I think we need to return some wrapper type here.

CBenoit (approved):
LGTM!

pacmancoder (approved):
Approving to speed up merge, I only have comment about possible `SHStrDupW` memory leak

crates/devolutions-pedm-contextmenu/devolutions-pedm-contextmenu-shell/src/lib_win.rs
CBenoit: We may have a similar problem with the "GetTitle" function, though maybe it's intended.

crates/devolutions-pedm-contextmenu/devolutions-pedm-contextmenu-shell/src/lib_win.rs
```diff
@@ -68,11 +70,24 @@ impl IElevationContextMenuCommand_Impl for ElevationContextMenuCommand_Impl {}
...
         // SAFETY: The `w` macro creates a constant, so the argument is valid.
-        unsafe { SHStrDupW(w!("Run as administrator with Devolutions PEDM")) }
+        unsafe { SHStrDupW(w!("Run elevated")) }
     }
 
     fn GetIcon(&self, _item_array: Option<&IShellItemArray>) -> Result<PWSTR> {
-        Err(E_NOTIMPL.into())
+        let module = match Module::current() {
+            Ok(mod_) => mod_,
+            Err(_) => return Err(E_NOTIMPL.into()),
```
CBenoit, line 79: nitpick: Maybe there is a better error to return here.

awakecoding:
I'll let you Rust experts make the finishing touches if needed, otherwise we can merge as-is :)

CBenoit:
Since it’s a net improvement with no outstanding problem, we can merge now and follow up later :)

PR #998, opened 2024-08-28
refactor: fix soundness hole in `Handle` API

CBenoit:
The API of `Handle` in `win-api-wrappers` is allowing UB without unsafe code. I adjusted the API to fix the hole.
Note that this PR is not fixing any vulnerability or bug per se. It only fixes the unsound safe API that we are using correctly in this repository only.

cc @kbouchard-dev


crates/win-api-wrappers/src/handle.rs
```diff
@@ -19,15 +19,44 @@ unsafe impl Send for Handle {}
...
+    pub unsafe fn new_owned(handle: HANDLE) -> Self {
+        // SAFETY: Same preconditions as the called function.
+        unsafe { Self::new(handle, true) }
+    }
+
+    // Wraps a borrowed Windows [`HANDLE`].
+    pub fn new_borrowed(handle: HANDLE) -> Self {
+        // SAFETY: It’s safe to wrap a non-owning Handle as we’ll not call `CloseHandle` on it.
+        unsafe { Self::new(handle, false) }
+    }
```
CBenoit, line 53: When we know that the handle is not owned in a static way, we can use the safe function `new_borrowed` which never causes any UB.

crates/win-api-wrappers/src/handle.rs
```diff
@@ -19,15 +19,44 @@ unsafe impl Send for Handle {}
...
+    // Wraps an owned Windows [`HANDLE`].
+    //
+    // # Safety
+    //
+    // - `handle` is a valid handle to an open object.
+    // - `handle` is not a pseudohandle.
+    // - The caller is actually responsible for freeing the `HANDLE` when the value goes out of scope.
+    pub unsafe fn new_owned(handle: HANDLE) -> Self {
+        // SAFETY: Same preconditions as the called function.
+        unsafe { Self::new(handle, true) }
```
CBenoit, line 46: The `Drop` implementation is assuming we constructed the `Handle` object in a sane way to call `CloseHandle`, but there is no way for us to verify that the handle is actually owned outside of the callsite. Conceptually, calling `Handle::new_owned(handle)` or `Handle::new(handle, true)` is like calling `CloseHandle` and inherits the safety preconditions.

pacmancoder (approved):

PR #1013, opened 2024-09-10
build(webapp): angular 18 changed build output path, update config to remove it

irvingoujAtDevolution:
Angular 18 build output now have `browser` and `server` sub path under build output directory. We don't want it, updated settings to remove this difference

reference: https://github.com/angular/angular-cli/issues/26304


CBenoit (approved):

CBenoit:
Sounds reasonable

PR #1043, opened 2024-10-04
chore(webapp): fix all existing Biome errors, add strictTemplate compiler options

irvingoujAtDevolution:
Fixed all biome issues, now we can confidently integrate Biome checking into our CI workflow
Added strictTemplate compiler option, better type safety.

1. Strongly typed WebSession
2. Delete a function (manageScreenSize) that I suppose does nothing (i.e Typescript disagree with it and has no effect on the usage)
3. Removed manual type as much as possible, let TypeScript Infer the types instead.


webapp/biome.json
```diff
@@ -1,13 +1,21 @@
...
+    ]
+  },
   "linter": {
     "rules": {
       "style": {
         "useTemplate": "off",
         "useImportType": "off"
+      },
+      "complexity": {
+        "noUselessConstructor": "off"
```
irvingoujAtDevolution, line 18: Angular sometimes need empty constructor

kristahouse (approved):
LGTM! Thanks!

CBenoit (approved):
Nice! By the way, do we enforce biome & ci in the CI?

irvingoujAtDevolution:
> Nice! By the way, do we enforce biome & ci in the CI?

Not yet, will do so in the next PR

PR #1045, opened 2024-10-07
fix recording player

irvingoujAtDevolution:
- Fix any type on WebSession related code
- chore(webapp): Fix all existing Biome errors, add strictTemplate compiler options
- chore(webapp) enforce biome in recording players


webapp/proxy.conf.json
```diff
@@ -1,33 +1,33 @@
 {
   "/jet/webapp/app-token": {
-    "target": "http://localhost:7272",
+    "target": "http://localhost:7171",
```
CBenoit, line 3: @irvingoujAtDevolution FYI, I revert to 7171 since that’s the default value for the HTTP listener: https://github.com/Devolutions/devolutions-gateway/blob/01fb589311fb25cf6e5c5f27e4b7a55708af5ae7/devolutions-gateway/src/config.rs#L1038-L1041

CBenoit (approved):
LGTM!

PR #1053, opened 2024-10-08
feat(agent): initial DVC server implementation

pacmancoder:
This PR contains the initial implementation DVC server with remote execution logic.

#### Supported execution modes:
- Run(ShellExecute), without IO redirection
- Process(CreateProcessW)
- Batch(bat file execution)
- Powershell (ps1 file execution; PowerShell 5.x)
- PwSh (ps1 file execution; PowerShell 7.x)

#### Features:
- IO redirection (in/out/err)
- Exit code reporting
- Abort/Cancel task


devolutions-session/Cargo.toml
```diff
@@ -35,4 +43,11 @@ features = [
     "Win32_System_IO",
     "Win32_System_Threading",
     "Win32_Security",
+    "Win32_System_Shutdown",
+    "Win32_UI_WindowsAndMessaging",
+    "Win32_UI_Shell",
+    "Win32_System_Console",
+    "Win32_System_Pipes",
+    # Required for Win32_System_Registry
+    "Win32_System_Registry",
```
CBenoit, line 52: Maybe we can remove this comment 🤣
It would be more interesting to put a comment with information about the codebase that we can't know just by looking at the name of the feature

devolutions-session/Cargo.toml
```diff
@@ -14,11 +14,19 @@ camino = { version = "1.1", features = ["serde1"] }
...
+now-proto-pdu = { version = "0.1", git = "https://github.com/Devolutions/IronRDP", rev = "7c268d863048d0a9182b3f7bf778668de8db4ccf" }
+ironrdp = { version = "0.5", git = "https://github.com/Devolutions/IronRDP", rev = "7c268d863048d0a9182b3f7bf778668de8db4ccf", features = ["core", "pdu"]}
+tokio = { version = "1", features = ["macros", "sync"] }
+async-trait = "0.1"
+futures = "0.3"
+tempfile = "3"
+
 
 [lints]
 workspace = true
```
CBenoit, line 32: A bit unrelated, but we should move this above the dependencies section. Currently it's kind of mixed with them.

devolutions-session/src/dvc/fs.rs
CBenoit, line 1: Is the file saved into a temporary directory? Do I understand correctly if the intention is to remove the temporary files as soon as they are not in use by us anymore, but even if the process crashes they will be removed on next reboot?

devolutions-session/src/dvc/fs.rs
```diff
@@ -0,0 +1,38 @@
+use std::path::PathBuf;
+
+pub struct ProcessTmpFile(PathBuf);
```
CBenoit, line 3: question: Is it possible to hold a File instead? Currently we may need to open files using `path()`, but there is then no relation encoded into the type system and we may drop this structure while the file is still is use. I didn't read the whole code yet.

devolutions-session/src/dvc/fs.rs
```diff
@@ -0,0 +1,38 @@
...
+
+impl ProcessTmpFile {
+    pub fn new(extension: &str) -> anyhow::Result<Self> {
+        // Create temporary batch file
```
CBenoit, line 7: style: Per IronRDP STYLE.md

```suggestion
        // Create temporary batch file.
```

devolutions-session/Cargo.toml
pacmancoder: This comment is just wrong xD I remember I wanted to comment about "Win32_System_Registry" requirement for a feature unrelated to the registry.

pacmancoder:
@CBenoit FYI I updated PR with the latest codebase; I'll rebase my branch tomorrow and open it for review

devolutions-session/src/dvc/fs.rs
pacmancoder: This type is used to keep previously opened files alive until the process finishes, I'll add docs

devolutions-session/src/dvc/fs.rs
pacmancoder: > if the process crashes they will be removed on next reboot
This logic is currently missing, I'll add that

devolutions-session/src/dvc/status.rs
```diff
@@ -0,0 +1,25 @@
+#[derive(Debug)]
```
pacmancoder, line 1: FYI: I am planning to revisit now proto in follow-up PR, those codes are temporary

pacmancoder:
@CBenoit PR is ready for review

crates/win-api-wrappers/src/rpc.rs
```diff
@@ -116,6 +116,7 @@ pub struct RpcServerInterfacePointer {
     pub raw: &'static RPC_SERVER_INTERFACE,
 }
 
+// SAFETY: RpcServerInterfacePointer could be sent between threads.
 unsafe impl Send for RpcServerInterfacePointer {}
```
CBenoit, line 120: issue: This is not explaining why. Implementing `Send` on `RpcServerInterfacePointer` _means_ that `RpcServerInterfacePointer` can be sent to another thread. Why is it safe to do so is the purpose of `SAFETY` comments. For this to be `Send`, we must not share mutable state with something else without enforcing exclusive access to it. This depends on `RPC_SERVER_INTERFACE` and the operations called on it by our wrapper. See https://doc.rust-lang.org/nomicon/send-and-sync.html

crates/win-api-wrappers/src/wts.rs
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+}
+
+/// RAII wrapper for WTS memory handle.
+pub struct WTSMemory(*mut core::ffi::c_void);
+
+impl WTSMemory {
+    /// # Safety
+    /// `ptr` must be a valid pointer to a handle returned from `WTSVirtualChannelQuery`.
```
CBenoit, line 33: question: Any chance you also wrap the call to `WTSVirtualChannelQuery`? Then `new` could be renamed to `from_raw`. Something like `Box`: https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw

crates/win-api-wrappers/src/wts.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub unsafe fn new(ptr: *mut core::ffi::c_void) -> Self {
+        Self(ptr)
+    }
+
+    pub fn raw(&self) -> *mut core::ffi::c_void {
+        self.0
+    }
```
CBenoit, line 40: suggestion: Usually this is the naming observed:
```suggestion
    pub fn as_ptr(&self) -> *const core::ffi::c_void {
        self.0
    }

    pub fn as_mut_ptr(&self) -> *mut core::ffi::c_void {
        self.0
    }
```

devolutions-session/Cargo.toml
```diff
@@ -8,31 +8,53 @@ description = "Session host application for Devolutions Agent"
...
+[features]
+default = ["dvc"]
+dvc = [
+    "async-trait",
+    "windows",
+    "now-proto-pdu",
+    "ironrdp",
+    "tempfile",
+    "thiserror",
+    "win-api-wrappers",
```
CBenoit, line 23: suggestion: Use `dep:` prefix for dependencies.
```suggestion
    "dep:async-trait",
    "dep:windows",
    "dep:now-proto-pdu",
    "dep:ironrdp",
    "dep:tempfile",
    "dep:thiserror",
    "dep:win-api-wrappers",
```

devolutions-session/src/dvc/channel.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub async fn send(&self, message: T) -> anyhow::Result<()> {
+        self.tx.send(message).await?;
+
+        // DVC IO loop is controlled by WinAPI events signaling, therefore we need to fire event to
+        // notify DVC IO loop about new incoming message.
+
+        // SAFETY: No preconditions.
+        unsafe {
+            SetEvent(self.event.raw())?;
+        }
```
CBenoit, line 27: suggestion: What about adding a method `set` to our Event wrapper?

devolutions-session/src/dvc/process.rs
```diff
@@ -0,0 +1,713 @@
...
+                        GetOverlappedResult(
+                            pipe_handle,
+                            &overlapped_stderr as *const _,
+                            &mut bytes_read as *mut _,
+                            false,
+                        )
+                    };
+
+                    if let Err(err) = overlapped_result {
+                        // SAFETY: No_preconditions.
```
CBenoit, line 414:
```suggestion
                        // SAFETY: No preconditions.
```

devolutions-session/src/dvc/channel.rs
pacmancoder: This was missed during refactoring, thanks, I'll revisit `SetEvent` calls

devolutions-session/src/main.rs
```diff
@@ -27,28 +40,100 @@ fn main() -> anyhow::Result<()> {
...
+
+#[cfg(not(all(windows, feature = "dvc")))]
+async fn spawn_tasks(_config: &Conf) -> anyhow::Result<Tasks> {
+    Ok(Tasks::new())
+}
+
+struct Tasks {
+    inner: Vec<ChildTask<anyhow::Result<()>>>,
+    shutdown_handle: ShutdownHandle,
+    #[allow(dead_code)]
```
CBenoit, line 116: issue: Add a comment explaining why the warning is silenced.

CBenoit (approved):
Nice work! You also have a conflict in Cargo.lock. Once everything is addressed, feel free to merge!

crates/win-api-wrappers/src/wts.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// # Safety
+    /// `handle` must be a valid handle returned from `WTSVirtualChannelOpenEx`.
+    pub unsafe fn new(handle: HANDLE) -> Self {
+        Self(handle)
+    }
+
+    pub fn raw(&self) -> HANDLE {
+        self.0
+    }
+}
```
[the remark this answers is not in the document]
pacmancoder: No, we indeed want unsafe `new` which takes raw `HANDLE` - that what is returned from `WTSVirtualChannelOpenEx`, this is already described in `new` `#Safety` section.

We handle ownership internally by implementing `Drop` which calls `WTSVirtualChannelClose`

crates/win-api-wrappers/src/wts.rs
pacmancoder: Encapsulated this logic inside `WtsChannel::query_file_handle` instead.

crates/win-api-wrappers/src/wts.rs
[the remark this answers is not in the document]
pacmancoder: `raw` has been [removed](https://github.com/Devolutions/devolutions-gateway/pull/1053#discussion_r1834067846)

crates/win-api-wrappers/src/wts.rs
```diff
@@ -11,6 +15,70 @@ impl WTSVirtualChannel {
         Self(handle)
     }
 
+    pub fn open_dvc(name: &str) -> anyhow::Result<Self> {
+        let channel_name = AnsiString::from(name);
+
+        #[allow(clippy::undocumented_unsafe_blocks)] // false positive
```
CBenoit, line 24: I recommend to use `#[expect]` so we are warned once the false positive is resolved :)

crates/win-api-wrappers/src/wts.rs
```diff
@@ -94,20 +89,16 @@ impl Drop for WTSVirtualChannel {
...
+    pub(crate) unsafe fn new(ptr: *mut core::ffi::c_void) -> Self {
         Self(ptr)
     }
 
-    pub fn as_mut_ptr(&self) -> *mut core::ffi::c_void {
-        self.0
-    }
-
-    pub fn as_handle(&self) -> HANDLE {
+    pub(crate) fn as_handle(&self) -> HANDLE {
```
CBenoit, line 101: Can probably remove `pub(crate)` since WTSMemory is not visible outside of the module?

PR #1058, opened 2024-10-15
fix(subscriber-dummy): make crate version valid by adding the `patch` part

RRRadicalEdward:
Because of this, I can't use DevolutionsGateway as a git dependency. cargo gives a parsing error 
image


RRRadicalEdward:
@CBenoit please have a look (I can't set the reviewing for you)

tools/subscriber-dummy/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "subscriber-dummy"
-version = "0.1"
+version = "0.1.0"
```
CBenoit, line 3:
```suggestion
version = "0.0.0"
```

CBenoit (approved):
Good catch, thank you! I modified the version to be 0.0.0 since there is no plan for publishing on crates.io

CBenoit:
@RRRadicalEdward Our CI is not happy with the repository being a fork, I’ll look into this. Sorry for the delay. In the meantime, you can point the dependency to your fork to test compilation.

PR #1064, opened 2024-10-22
refactor: replace `lazy_static` with `LazyLock`

allan2:
This PR removes the `lazy_static` dependency as per this [FIXME](https://github.com/Devolutions/devolutions-gateway/blob/193630fd7e1624aa1a73592f01205aa788529b60/devolutions-gateway/Cargo.toml#L44).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.24.1"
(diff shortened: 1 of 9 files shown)


CBenoit (approved):
LGTM! The comment was a bit outdated, but you rightfully used LazyLock which was stabilized in 1.80.0! :+1:

PR #1068, opened 2024-10-23
chore: fix some Clippy lints

allan2:


devolutions-session/src/main.rs
```diff
@@ -21,13 +21,17 @@ fn main() -> anyhow::Result<()> {
...
+    #[cfg(not(windows))]
+    let _logger_guard = init_log(config);
+
     // TMP: Copy-paste from MSRDPEX project for testing purposes.
     #[cfg(windows)]
-    loop_dvc(config);
+    {
+        let _logger_guard = init_log(config.clone());
+        loop_dvc(config);
+    }
```
CBenoit, line 34: This is problematic when building on Windows. The guard will get dropped at the end of the very short scope which is only present so that we can conditionally compile the code. I recommend to revert the code to its previous form, and use the `#[expect()]` attribute to silent the warning.

I would go with something like this:
```rust
    #[cfg_attr(not(windows), expect(clippy::redundant_clone))]
    let _logger_guard = init_log(config.clone());
```

The lint is expected to be fired on platforms other than Windows, and if it ever stops being fired, we’ll get warned about that so we can remove the now-useless attribute.

allan2: Alternatively, `init_log` can take a reference.

```rs
pub fn init_log(config: &ConfHandle) -> LoggerGuard {
    ..
}

let _logger_guard = init_log(&config);  // same as before, without a `cfg` attribute
```

This would pass the [`needless_pass_by_value`](https://rust-lang.github.io/rust-clippy/master/index.html#/needless_pass) lint.

CBenoit: Sounds good to me. Actually, this function could be modified to take a `&Conf`. There is no need for the whole `ConfHandle` which exists to allow hot reloading of the configuration.

```rust
pub fn init_log(config: &Conf) -> LoggerGuard {
    devolutions_log::init::<SessionLog>(
        &conf.log_file,
        conf.verbosity_profile.to_log_filter(),
        conf.debug.log_directives.as_deref(),
    )
    .expect("BUG: Failed to initialize log")
}
```

allan2: Addressed in 548cfc3.

allan2:
Ready for review. This PR fixes easy and non-ambiguous lints.

Remaining unfixed lints include:
- unused_crate_dependencies
- unreachable_pub - most warnings are in Agent; I assume more modules will be exported in the future
- unused_qualifications - I did not replace `std::mem::size_of`/`mem::size_of` with `size_of`
- clippy::as_underscore

More PRs to follow for the remaining ones.

devolutions-session/src/dvc.rs
```diff
@@ -15,8 +15,8 @@ use windows::Win32::System::IO::{GetOverlappedResult, OVERLAPPED};
 
 const CHANNEL_PDU_LENGTH: usize = 1024;
 
-pub fn loop_dvc(config: ConfHandle) {
-    if !config.get_conf().debug.enable_unstable {
+pub fn loop_dvc(conf: &Conf) {
+    if !conf.debug.enable_unstable {
         debug!("DVC loop is disabled");
         return;
     }
```
CBenoit, line 22: suggestion: I think it would be better to make the check for `enable_unstable` outside of this function. As it appears, this function does not really require the `Conf` object.

This is better style:
```rust
    if config.get_conf().debug.enable_unstable { // or conf.debug.enable_unstable if you have that at hand
        loop_dvc(config);
    }
```

Here is also a read I recommend: https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html

devolutions-session/src/main.rs
```diff
@@ -19,15 +19,17 @@ fn main() -> anyhow::Result<()> {
 
     std::fs::create_dir_all(get_data_dir()).context("Failed to create data directory")?;
 
-    let config = ConfHandle::init().context("Failed to initialize configuration")?;
+    let conf = ConfHandle::init()
+        .context("Failed to initialize configuration")?
```
CBenoit, line 23: nitpick/style: You’re not behind this line of code, but I would like to mention that this should start by a lower case. It’s a nitpick as it’s not really causing much harm outside of error messages a bit less nice-looking.

Recommended read: https://github.com/Devolutions/IronRDP/blob/f5dd282271df0084f589f4e163459b41cc2f2097/STYLE.md#formatting-of-error-messages

CBenoit:
Thank you!

>     * unused_crate_dependencies

If you want to address this one, let me know beforehand so we can discuss on how it should be approached best.
 
>     * unreachable_pub - most warnings are in Agent; I assume more modules will be exported in the future

Actually, it’s probably a sign that the `pub` keyword should be removed, or `pub(crate)` should be used.
In some cases such a sealed traits, we actually want to silent this warning though.

>     * unused_qualifications - I did not replace `std::mem::size_of`/`mem::size_of` with `size_of`

For that one, in order of preference, I recommend: `mem::size_of` > `std::mem::size_of` > `size_of`

- `mem::size_of` is the best readability and clarity overall
- `std::mem::size_of` when the function is used only once
- `size_of` when it is used a lot for doing arithmetic, I would then recommend a local import at the function level where it is used abundantly

That being said, those are just guidelines, and anything is acceptable :)

devolutions-session/src/dvc.rs
CBenoit: cc @pacmancoder

allan2: Addressed in [54772a4](https://github.com/Devolutions/devolutions-gateway/pull/1068/commits/54772a44f1f704cc22b2aee6d56e28bd82fb4b70).

devolutions-session/src/main.rs
allan2: Addressed in [b769880](https://github.com/Devolutions/devolutions-gateway/pull/1068/commits/b769880362f9c79c02bcc8c139b318fc1429b1d6).

CBenoit (approved):
I removed the duplicated commits from master:

tmpscrot

Maybe something wrong happened when you tried to rebase on top of master or merge master into your branch?

Otherwise, everything is looking good to me!

CBenoit:
I reworded a bit your individual commits so we can merge rebase them without squashing everything. It mostly comes down to the fact that changing code without modifying the semantic is a "refactor", changes to doc only (not code) should be "docs" and I added the scope when it made sense. It’s mostly relevant for features and bug fixes that are actually perceived by the end user as we have scripts looking for these commits when generating the changelog to avoid forgetting important changes. This is based on the [conventional commit specification](https://www.conventionalcommits.org/en/v1.0.0/). All that said, I should the details in the repository directly :)

PR #1070, opened 2024-10-25
fix(dgw): include /jet/jrec/delete/ in OpenAPI document

CBenoit:


pacmancoder (approved):

devolutions-gateway/openapi/gateway-api.yaml
```diff
@@ -144,6 +144,34 @@ paths:
       security:
       - scope_token:
         - gateway.heartbeat.read
+  /jet/jrec/delete/{id}:
+    delete:
+      tags:
+      - Jrec
+      summary: Lists all recordings stored on this instance
```
fdubois1, line 151: It is probably a copy/paste. The endpoint will be used to delete I guess.

CBenoit: Good catch. Fixed in https://github.com/Devolutions/devolutions-gateway/pull/1077

PR #1074, opened 2024-10-25
chore(tokengen): add server binary

irvingoujAtDevolution:
- Refactored `tokengen/main.rs` into `main.rs` and `lib.rs` for better code organization and reusability.
- Introduced a new binary target that runs as a server to generate tokens.

This server enhances the development experience by:
1. Simplifying token generation for web developers, eliminating the manual process of using the CLI and copying tokens to access the gateway.
2. Paving the way for future integration with IronRDP web demos, enabling smoother token handling.

### DIFF
--- a/tools/tokengen/src/bin/server/main.rs
+++ b/tools/tokengen/src/bin/server/main.rs
@@ -1,0 +1,51 @@
+// This Server binary is intended to be used as a standalone server for the provisioner key
+// Primarily for development and testing purposes
+#[cfg(feature = "server")]
+pub mod server;
+
+#[cfg(feature = "server")]
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    use axum::http::Method;
+    use server::{create_router, get_provisioner_key_path};
+    use std::path::PathBuf;
+    use std::sync::Arc;
+    use tower_http::trace::TraceLayer;
+    use tracing::info;
+    use tracing_subscriber::layer::SubscriberExt;
+    use tracing_subscriber::util::SubscriberInitExt;
+
+    tracing_subscriber::registry()
+        .with(
+            tracing_subscriber::EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+                format!(
+                    "{}=debug,tower_http=debug,axum::rejection=trace",
+                    env!("CARGO_CRATE_NAME")
+                )
+                .into()
+            }),
+        )
+        .with(tracing_subscriber::fmt::layer())
+        .init();
+
+    // Read the provisioner key path from the configuration
+    let provisioner_key_path: Arc<PathBuf> = get_provisioner_key_path().await?;
+
+    // Create the router with subcommand routes
+    let app = create_router(provisioner_key_path)
+        .layer(
+            tower_http::cors::CorsLayer::new()
+                .allow_methods([Method::POST, Method::OPTIONS, Method::GET])
+                .allow_headers(tower_http::cors::Any)
+                .allow_origin(tower_http::cors::Any),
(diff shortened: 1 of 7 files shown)


CBenoit (commented):
The idea would be to start this server and let the demo web client request tokens on its own (without authentication)?

Can I suggest you to keep both in a single binary? I think it would not be too big, and we already have a workflow for tokengen itself.

Nest the CLI into two new subcommands:
- `tokengen sign`
- `tokengen server`

I also suggest that you re-use the route we already have in the Devolutions Gateway for the webapp:
https://github.com/Devolutions/devolutions-gateway/blob/eac37e81e3006d747543e331a05137c9281f1109/devolutions-gateway/src/api/webapp.rs#L222-L453

I understand we only need to support session tokens to allow the demo web clients to generate session tokens. For instance, we don’t need to support JRL tokens.

tools/tokengen/src/main.rs
```diff
@@ -1,215 +1,85 @@
+// main.rs
+
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

tools/tokengen/src/bin/server/main.rs
```diff
@@ -0,0 +1,51 @@
...
+                    "{}=debug,tower_http=debug,axum::rejection=trace",
+                    env!("CARGO_CRATE_NAME")
+                )
+                .into()
+            }),
+        )
+        .with(tracing_subscriber::fmt::layer())
+        .init();
+
+    // Read the provisioner key path from the configuration
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

tools/tokengen/src/server/server_impl.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+#[derive(Deserialize)]
+pub(crate) struct CommonRequest {
+    #[serde(default)]
+    validity_duration: Option<String>,
+    #[serde(default)]
+    kid: Option<String>,
+    #[serde(default)]
+    delegation_key_path: Option<String>,
```
CBenoit, line 62: I think this should not be part of the request. Why not proceed like for `provisioner_key_path`?

CBenoit, line 62: I think this should not be part of the request. Why not proceed like for `provisioner_key_path`?

tools/tokengen/src/server/server_impl.rs
```diff
@@ -0,0 +1,308 @@
...
+        provisioner_key_path
+    };
+
+    Ok(Arc::new(provisioner_key_path))
+}
+
+#[derive(Deserialize)]
+pub(crate) struct CommonRequest {
+    #[serde(default)]
+    validity_duration: Option<String>,
```
CBenoit, line 58: naming: `lifetime: Option<u64>` (the lifetime in seconds)

CBenoit (commented):
> Would it be reasonable to keep the API as it is for now? The tokengen app serves a different purpose (mostly for testing), and occasionally we might need more granular control over the token content. Let me know your thoughts on this approach.

Fair enough.
Let me suggest you a few modifications on what you already submitted then.

tools/tokengen/src/server/server_impl.rs
irvingoujAtDevolution: updated

tools/tokengen/src/server/server_impl.rs
irvingoujAtDevolution: updated

CBenoit (approved):
LGTM

PR #1075, opened 2024-10-25
feat(dgw): streaming player web-component

irvingoujAtDevolution:
Added the web-component for the streaming player and a demo for the usage of new  streaming API.

I believe this might not be needed to place inside the gateway repo, please let me know and I'll close the pull request if you think so. @CBenoit


CBenoit (approved):
I’ve been thinking about this. I would have suggested to put that at the same place we keep the recording player; but we have both the standalone web app and the recording player inside this repository. I think it’s fine to keep everything here. We’ll have to package that into the Devolutions Gateway installer too. But, if you prefer to have that in a separate repository, I think it’s fine too! My only concern is that we may be putting too many "unrelated" things into the webapp folder. I may move some things in the future.

Anyway, if you are satisfied with the current state on your side, feel free to merge!

CBenoit:
Note, I adjusted the PR title from `feat(webapp): Streaming player web-component` to `feat(dgw): streaming player web-component`.
Be sure to keep this modification when squashing.
The scope is `dgw` because this is a component that will be part of the Devolutions Gateway standalone.

CBenoit:
@irvingoujAtDevolution read this before merging: https://github.com/Devolutions/devolutions-gateway/pull/1076#pullrequestreview-2399369194

Maybe I misunderstood your intention with this component, and you intended to plug it directly in DVLS and Hub web UI?

Also, may I suggest to name this player "Shadowing Player"?

irvingoujAtDevolution:
@CBenoit , I intent to use it as part of the recording player, made it a web component solely because it is easier to test and maintain with separations of concern

CBenoit (approved):
Thank you! I’ll merge this as the initial implementation.

PR #1084, opened 2024-10-31
feat(dgw): initial implementation for /jet/fwd/http endpoint

CBenoit:
Currently behind the unstable flag.
Some details may be adjusted in the future.

For HTTP requests without upgrade, we do not account the operation as a session: it does not influence DVLS licensing quota, subscriber events, etc.
HTTP requests resulting into a WebSocket upgrade however will count as a fully-fledged session as usual.

Issue: DGW-217
Changelog: ignore

# How to test

## Enable instable features

In `gateway.json`:

```json
{
  /* snip */
  "__debug__": {
    "enable_unstable": true
  }
}
```

## HTTP request without upgrade

You can use `curl`:

```shell
curl -X GET \
  "http://localhost:7171/jet/fwd/http/f4fae090-5c07-4350-8fd6-937e0f3a5b43" \
  -H "Dgw-Request-Target: GET /todos/1" \
  -H "Authorization: Bearer $(
    tokengen \
      sign --provisioner-key /path/to/provisioner.key \
      bridge --target-host https://jsonplaceholder.typicode.com --jet-aid f4fae090-5c07-4350-8fd6-937e0f3a5b43
    )"
```

## HTTP request with WebSocket upgrade

You can use `jetsocat`:

```shell
jetsocat forward \
  "ws://127.0.0.1:7171/jet/fwd/http/d3a8ac78-49a4-4866-9ded-cf73a24f52c4?requestTarget=%2F&token=$(
    tokengen \
      sign --provisioner-key /path/to/provisioner.key \
      bridge --target-host wss://echo.websocket.org --jet-aid d3a8ac78-49a4-4866-9ded-cf73a24f52c4
  )" \
  -
```

### DIFF
--- a/devolutions-gateway/src/api/fwd.rs
+++ b/devolutions-gateway/src/api/fwd.rs
@@ -1,14 +1,13 @@
 use std::net::SocketAddr;
 use std::sync::Arc;
 
-use anyhow::Context;
+use anyhow::Context as _;
 use axum::extract::ws::WebSocket;
 use axum::extract::{self, ConnectInfo, State, WebSocketUpgrade};
 use axum::response::Response;
-use axum::routing::{self, get};
 use axum::Router;
 use tokio::io::{AsyncRead, AsyncWrite};
-use tracing::{field, Instrument};
+use tracing::{field, Instrument as _};
 use typed_builder::TypedBuilder;
 use uuid::Uuid;
 
@@ -22,7 +21,7 @@ use crate::token::{ApplicationProtocol, AssociationTokenClaims, ConnectionMode, 
 use crate::{utils, DgwState};
 
 pub fn make_router<S>(state: DgwState) -> Router<S> {
-    use routing::MethodFilter;
+    use axum::routing::{self, get, MethodFilter};
 
     let router = Router::new()
         .route("/tcp/:id", get(fwd_tcp))
@@ -340,9 +339,10 @@ async fn fwd_http(
     // 4. Forward the request.
 
     let response = if matches!(request.uri().scheme_str(), Some("ws" | "wss")) {
-        // 4.a Perform the WebSocket upgrade.
+        // 4.a Prepare the WebSocket upgrade.
 
-        // We are discarding the original body. There is no HTTP body when performing a WebSocket upgrade.
+        // We are discarding the original body.
+        // There is no HTTP body when performing a WebSocket upgrade.
         let (mut parts, _) = request.into_parts();
 
         let client_ws = WebSocketUpgrade::from_request_parts(&mut parts, &state).await.map_err(
@@ -356,6 +356,7 @@ async fn fwd_http(
         let request = axum::http::Request::from_parts(parts, ());
         let request_uri = request.uri().clone();
 
+        // TODO: Manually configure the TlsConnector and take the `dangerous_tls` parameter into account.
         let (server_ws, server_ws_response) = connect_async(request).await.map_err(
             HttpError::bad_gateway()
                 .with_msg("WebSocket connection to target server")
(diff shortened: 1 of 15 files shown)


devolutions-gateway/src/api/fwd.rs
```diff
@@ -237,3 +254,288 @@ where
...
+        Parameter::new("Dgw-Request-Target", |params| params.request_target.take());
+
+    const DANGEROUS_TLS_PARAM: Parameter<bool> =
+        Parameter::new("Dgw-Dangerous-Tls", |params| params.dangerous_tls.take());
+
+    #[derive(Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    struct QueryParams {
+        request_target: Option<String>,
+        dangerous_tls: Option<bool>, // QUESTION: Maybe we should bake this parameter into the Bridge token.
```
pacmancoder, line 306: I may not have enough context to answer this. Is this blocking for merge, could it me resolved in follow-up?

It first glance, it indeed looks not safe to me to select a dangerous TLS connector via HTTP param, adding it to signed token claims sounds about right

devolutions-gateway/src/api/fwd.rs
```diff
@@ -237,3 +254,288 @@ where
...
+                    .time_to_live(claims.jet_ttl)
+                    .recording_policy(claims.jet_rec)
+                    .build();
+
+                let result = Proxy::builder()
+                    .conf(conf)
+                    .session_info(info)
+                    .address_a(client_addr)
+                    .transport_a(client_stream)
+                    .address_b("8.8.8.8:8888".parse().expect("valid hardcoded value")) // NOTE: We don’t really use this address for anything else other than pcap recording, so it’s fine to use a placeholder for now.
```
pacmancoder, line 417: [nit][style] This line with the comment looks pretty long, could we move the address to constant and add a comment above?

devolutions-gateway/src/api/fwd.rs
```diff
@@ -237,3 +254,288 @@ where
...
+
+        let response = Response::from(response);
+        let (parts, body) = response.into_parts();
+        let body = axum::body::Body::from_stream(body.into_data_stream());
+        Response::from_parts(parts, body)
+    };
+
+    return Ok(response);
+
+    // * Helpers //
```
pacmancoder, line 464: [nit][style] There was different section style in other places (`// -- local helper -- //`) is it intentional?

pacmancoder (approved):
LGTM 🥇

devolutions-gateway/src/api/fwd.rs
CBenoit: Not a blocker, and the API is behind the unstable flag anyway. It's not as bad as it may look since we still need a token to use this endpoint, and we are just disabling checks usually done on client side, usually disabled at the client'a discretion. Devolutions Gateway itself is not impacted by that. Putting this in the token may make things more complicated than they really need to be. We would need DVLS to know when an unsafe TLS token should be provided or not. This may end up being put in the query params of the DVLS endpoint so the client can chooses anyway. At this point it's just a less flexible solution (we can't mix normal and unsafe calls with a single token), ultimately potentially less secure than it needs to be. What do you think?

devolutions-gateway/src/api/fwd.rs
CBenoit: Fair enough! I deviated to follow the 1., 2. (numbered) ... *, * (bullet) style, for a document-like readability. (Almost literate programming.)

devolutions-gateway/src/api/fwd.rs
pacmancoder: > It's not as bad as it may look since we still need a token to use this endpoint, and we are just disabling checks usually done on client side

In that case, putting it in params sounds fine to me 👍

PR #1086, opened 2024-11-01
docs: add json scheme for gateway.json

irvingoujAtDevolution:
Make it easier for manually editing the config file, it can be referenced directly with GitHub URL and users will receives update when we make changes in this file.


CBenoit (commented):
Why not! Can you add the description field too? The readme is documenting everything, you could take the descriptions from there.

irvingoujAtDevolution:
> Why not! Can you add the description field too? The readme is documenting everything, you could take the descriptions from there.

Updated

CBenoit (approved):

PR #1091, opened 2024-11-05
Initial PEDM refactoring

thenextman:
This is a work-in-progress refactoring of the PEDM module, while we get to grips with the functionality and take the prototype to something releasable. Subsequent PRs will be more focused.

This pull request:

- Patches memory bugs in security primitives (in the `sid`, `privilege` and `token` modules) that were causing runtime crashes in debug builds and clear undefined behaviour otherwise
- Disables logging of PEDM elevations. The log target is determined based on finding the file for the current day, but misses the fact that we might not roll the log file daily. This will be replaced by a database-backed logging solution.
- Disables configuration of secure program data directories. The functionality is currently broken due to a loosely defined bug in the relevant security primitives.
- Remove the rules configuration from the core API and regenerates clients accordingly
- Introduces a revised PEDM desktop agent, running as a system tray application and using system notifications

A follow up PR will integrate packaging and build for the revised PEDM desktop agent.


CBenoit (approved):
Thank you!

PR #1095, opened 2024-11-07
fix(dgw): adjust delete_many endpoint

CBenoit:
Instead of failing and returning 404 Not Found when one of the recordings
is not found, we return the number of recordings not found in the
response body with a 200 Ok.

The client may use this information to notify the user about a possible
out of sync state on their side and suggest to rerun an indexing
operation.

Changelog: ignore


devolutions-gateway/openapi/ts-angular-client/api/jrec.service.ts
```diff
@@ -88,6 +90,88 @@ export class JrecService {
         return httpParams;
     }
 
+    /**
+     * Mass-deletes recordings stored on this instance
+     * If you try to delete more than 1,000,000 recordings at once, you should split the list into multiple requests to avoid timing out during the processing of the request.  The request processing consist in 1) checking if one of the recording is active, 2) counting the number of recordings not found on this instance.  When a recording is not found on this instance, a counter is incremented. This number is returned as part of the response. You may use this information to detect anomalies on your side. For instance, this suggests the list of recordings on your side is out of date, and you may want re-index.
```
pacmancoder, line 95: [nit] Something gone wrong with formatting here

devolutions-gateway/openapi/ts-angular-client/api/jrec.service.ts
```diff
@@ -88,6 +90,88 @@ export class JrecService {
...
+    public deleteManyRecordings(requestBody: Array<string>, observe?: 'events', reportProgress?: boolean, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<HttpEvent<DeleteManyResult>>;
+    public deleteManyRecordings(requestBody: Array<string>, observe: any = 'body', reportProgress: boolean = false, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<any> {
+        if (requestBody === null || requestBody === undefined) {
+            throw new Error('Required parameter requestBody was null or undefined when calling deleteManyRecordings.');
+        }
+
+        let localVarHeaders = this.defaultHeaders;
+
+        let localVarCredential: string | undefined;
+        // authentication (scope_token) required
```
pacmancoder, line 111: [question] Are we using the same style of comments here in `ts` as in rust, or are we using a different convention? (Full sentences ending with `.`)

devolutions-gateway/src/api/jrec.rs
```diff
@@ -179,44 +193,32 @@ async fn jrec_delete_many(
...
-        }
-
-        Ok((delete_list, recording_paths))
-    });
-
-    let (delete_list, recording_paths) = join_handle.await.map_err(HttpError::internal().err())??;
+) -> Result<Json<DeleteManyResult>, HttpError> {
+    use std::collections::HashSet;
+
+    const BLOCKING_THRESHOLD: usize = 100_000;
```
pacmancoder, line 199: [qeustion]: This threshold looks pretty big, what is the measured delay for processing 100000 items without `spawn_blocking`?

pacmancoder (approved):

devolutions-gateway/openapi/ts-angular-client/api/jrec.service.ts
CBenoit: The generator is removing newlines, but I think it’s fine though.

devolutions-gateway/openapi/ts-angular-client/api/jrec.service.ts
CBenoit: This is generated code, so we don’t follow anything on our side specifically

PR #1102, opened 2024-11-11
fix(webapp): the initial 401 error is shown when it should not

irvingoujAtDevolution:
Issue: DGW-226

### DIFF
--- a/webapp/src/client/app/modules/login/login.component.ts
+++ b/webapp/src/client/app/modules/login/login.component.ts
@@ -77,7 +77,7 @@ export class LoginComponent extends BaseComponent implements OnInit {
     }
   }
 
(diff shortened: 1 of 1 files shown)


webapp/src/client/app/modules/login/login.component.ts
```diff
@@ -77,7 +78,7 @@ export class LoginComponent extends BaseComponent implements OnInit {
   }
 
   private handleAutoLoginError(error): Observable<boolean> {
-    if (error.status && error.status !== '401') {
+    if (error?.status !== 401) {
```
irvingoujAtDevolution, line 81: it's an integer instead of a string

rxjs catch error does not infer the type of the error

CBenoit (approved):
LGTM! Thank you

PR #1106, opened 2024-11-13
fix(webapp): session toolbar loading issues

kristahouse:
Issue: DGW-230

Resolved Style Loading Delay in Session Component by Leveraging Existing Toolbar Component

Addressed an issue where .scss styles were not loading immediately for the ARD session component when it was loaded dynamically. 

**Utilized Existing Session Toolbar Component:** 
I used the session toolbar component, which was already implemented for the terminal based components (SSH & Telnet). I also updated all the other windows based components (RDP & VNC) as they all were at risk of the UI missing it's styles during a loading delay. 

**Reasoning:**
The toolbar component independently manages its styles, avoiding any delay caused by asynchronous tasks in the parent session component. Leveraging this component for style loading eliminates the need for further adjustments to achieve consistent, immediate style rendering.

**Testing and Validation:**
Verified that styles in the toolbar component load immediately when the session component initializes dynamically.
Confirmed that there are no regressions in other components utilizing the toolbar.

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -38,7 +38,6 @@
         "@angular/cli": "18.2.1",
         "@angular/compiler-cli": "18.2.1",
         "@biomejs/biome": "1.9.3",
-        "@biomejs/cli-linux-x64": "1.9.3",
         "@types/jasmine": "~4.3.0",
         "@types/uuid": "^9.0.8",
         "jasmine-core": "~4.6.0",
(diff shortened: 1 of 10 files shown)


irvingoujAtDevolution (approved):
LGTM

CBenoit:
This is looking good to me too! You need to resolve the conflicts and ensure the CI is green before we can merge :slightly_smiling_face:

PR #1110, opened 2024-11-18
fix(libsql): fix build on windows

irvingoujAtDevolution:
Use Marc's Pull Request from libsql https://github.com/tursodatabase/libsql/pull/1791/commits

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3041,8 +3041,7 @@ dependencies = [
 [[package]]
 name = "libsql"
 version = "0.6.0"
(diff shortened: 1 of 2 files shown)


CBenoit (commented):
Can you use override it using a patch in the top-level Cargo.toml?
https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html

Once a new version is published on crates.io, we want to switch to that :slightly_smiling_face:

irvingoujAtDevolution:
Updated!

CBenoit (approved):
LGTM!

PR #1114, opened 2024-11-20
feat(pedm): add WinAPI wrappers for WTS functions

thenextman:
Add WinAPI wrappers for further WTS (Windows Terminal Services) APIs:

- Enumerate sessions
- Query session username and domain information
- Force a session to logoff
- Send a message to a session

Existing `session_has_logged_in_user` is moved from `sessions` module to `wts` module, since modules here should be factored around the underlying APIs instead of higher level functionality.

Given a session ID, the user SID can already be queried using the `token` module.

Additionally, fix a potential resource leak in the `netmgmt` module.

Changelog: ignore


crates/win-api-wrappers/src/wts.rs
```diff
@@ -87,24 +98,223 @@ impl Drop for WTSVirtualChannel {
...
+
+impl TryFrom<&WTS_SESSION_INFOW> for WTSSessionInfo {
+    type Error = anyhow::Error;
+
+    fn try_from(value: &WTS_SESSION_INFOW) -> Result<Self, Self::Error> {
+        let win_station_name: String;
+
+        // SAFETY: `pWinStationName` will always hold a valid WSTR pointer if
+        // `WTS_SESSION_INFOW` was returned from Remote Desktop Services (`WTSEnumerateSessionsW``)
+        unsafe { win_station_name = value.pWinStationName.to_string()? }
```
CBenoit, line 151: issue: It’s possible to construct a WTS_SESSION_INFOW in safe Rust, and to call this function like `WTSSessionInfo::try_from(my_session_info)`, and this may cause memory corruptions.

And it’s technically okay to create a WTS_SESSION_INFOW on Rust side. The precondition is not really about being allocated my Remote Desktop Services. It’s more about "the pointer being valid for reads up until and including the next `\0`": https://docs.rs/windows-strings/latest/windows_strings/struct.PWSTR.html#method.as_wide

This is called an "soundness" bug and the API is said "unsound": https://docs.rs/dtolnay/latest/dtolnay/macro._03__soundness_bugs.html

suggestion: Remove this `TryFrom` implementation, and replace it by a `unsafe fn new(&WTS_SESSION_INFOW) { … }` function with a `# Safety` section basically requiring the same preconditions as `PWSTR::as_wide()`.

crates/win-api-wrappers/src/netmgmt.rs
```diff
@@ -89,20 +86,45 @@ pub fn get_local_group_members(group_name: String) -> anyhow::Result<Vec<Sid>> {
...
+}
+
+impl Drop for NetMgmtMemory {
+    fn drop(&mut self) {
+        if self.0.is_null() {
+            return;
+        }
+
+        // SAFETY: No preconditions.
+        unsafe { NetApiBufferFree(Some(self.0)) };
```
CBenoit, line 122: suggestion: In term of wording, I prefer
```suggestion
        // SAFETY: FFI call with no outstanding precondition.
        unsafe { NetApiBufferFree(Some(self.0)) };
```

Because there are typically preconditions, but it’s the "FFI usual business". The pointer must be valid, etc.

crates/win-api-wrappers/src/wts.rs
```diff
@@ -87,24 +98,232 @@ impl Drop for WTSVirtualChannel {
...
+}
+
+impl WTSSessionInfo {
+    /// # Safety
+    ///
+    /// - The `pWinStationName` pointer needs to be valid for reads up until and including the next `\0`.
+    unsafe fn new(wts_session: &WTS_SESSION_INFOW) -> anyhow::Result<Self> {
+        Ok(Self {
+            session_id: wts_session.SessionId,
+            win_station_name: unsafe { wts_session.pWinStationName.to_string()? },
```
thenextman, line 150: Question @CBenoit: I understand why the linter wants me to add the `unsafe` block here; but is it necessary to provide a SAFETY comment also - given that it's the same precondition documented at the function level?

crates/win-api-wrappers/src/wts.rs
thenextman: Indeed; my thinking here was that since this is a private function we _know_ we're not building `WTS_SESSION_INFOW` ourselves. Since we _could_ do this, however, I understand the refactoring. Starting to think more like a rustacean :)

thenextman:
@CBenoit Thanks for the detailed explanations; I've think I've addressed all comments and also added one question.

crates/win-api-wrappers/src/wts.rs
CBenoit: Yes, absolutely. The rationale is that someone reading this code has no idea that the preconditions are the same, will be confused, and end up looking up the documentation again to double check. (Personal experience. :joy: )
It’s probably not too bad when skimming in a code editor with a rich LSP integration because you can hover over the function to read the documentation. But you don’t always have that: reading code in GitHub, or even just reading the code from a Linux machine where this windows-specific code is disabled.

It’s enough to go with something like
```rust
// SAFETY: Same preconditions as the current function.
```

crates/win-api-wrappers/src/wts.rs
CBenoit: I would avoid this kind of implicit contracts even with a private function, but keep in mind that the functions implemented via a trait have the same visibility as the trait, i.e.: for TryFrom it’s public. So when I reviewed I considered this as an unsound _public_ API :slightly_smiling_face:

crates/win-api-wrappers/src/wts.rs
```diff
@@ -87,24 +98,232 @@ impl Drop for WTSVirtualChannel {
...
+    let sessions_slice: &[WTS_SESSION_INFOW] =
+        unsafe { std::slice::from_raw_parts(sessions_ptr.0 as *mut u8 as *mut WTS_SESSION_INFOW, len as usize) };
+
+    let mut sessions = Vec::<WTSSessionInfo>::with_capacity(sessions_slice.len());
+    for session in sessions_slice {
+        // SAFETY: `session` is valid
+        unsafe {
+            sessions.push(WTSSessionInfo::new(session)?);
+        }
+    }
```
CBenoit, line 179: suggestion: It’s considered good style to shadow variables that should not be used anymore, especially when you can use this to prove correctness of the code:

```rust
// SAFETY:
// - `WTSEnumerateSessionsW` returns a pointer valid for `len` reads of `WTS_SESSION_INFOW`.
// - `WTSEnumerateSessionsW` is also returning a pointer properly aligned.
// - We ensure the memory referenced by the slice is not mutated by shadowing the variable.
// - There are never so many sessions that `len * mem::size_of::<WTS_SESSION_INFOW>()` overflows `isize`.
let sessions: &[WTS_SESSION_INFOW] =
    unsafe { std::slice::from_raw_parts(sessions.0 as *mut u8 as *mut WTS_SESSION_INFOW, len as usize) };
```

Here is an improved version of this code:

```suggestion
    let mut sessions: *mut WTS_SESSION_INFOW = std::ptr::null_mut();
    let mut len = 0u32;

    // SAFETY: All buffers are valid.
    unsafe {
        WTSEnumerateSessionsW(WTS_CURRENT_SERVER_HANDLE, 0, 1, &mut sessions, &mut len)?;
    };

    // SAFETY: `sessions` is always a valid pointer on success.
    // `sessions` must be freed by `WTSFreeMemory`.
    let sessions = unsafe { WTSMemory::from_raw(sessions.cast()) };

    // SAFETY:
    // - `WTSEnumerateSessionsW` returns a pointer valid for `len` reads of `WTS_SESSION_INFOW`.
    // - `WTSEnumerateSessionsW` is also returning a pointer properly aligned.
    // - We ensure the memory referenced by the slice is not mutated by shadowing the variable.
    // - There are never so many sessions that `len * mem::size_of::<WTS_SESSION_INFOW>()` overflows `isize`.
    let sessions: &[WTS_SESSION_INFOW] =
        unsafe { std::slice::from_raw_parts(sessions.0 as *mut u8 as *mut WTS_SESSION_INFOW, len as usize) };

    let sessions = sessions.iter()
        .map(|session| {
            // SAFETY: `session` is a WTS_SESSION_INFOW initialized by Remote Desktop Services that we trust to be valid.
            unsafe { WTSSessionInfo::new(session) }
        })
        .collect::<Result<Vec<WTSSessionInfo>, _>>()?;
```

CBenoit (approved):
LGTM! Let me open a follow up PR.

PR #1116, opened 2024-11-21
refactor(agent): updater improvements

pacmancoder:
This PR adds logic to allow arbitrary version upgrades for the devolutions gateway updater (previously we were only able to install the latest stable version) and also allows us to downgrade the product version.

However, currently, our MSI Installer [forbids ](https://github.com/Devolutions/devolutions-gateway/blob/fix/devolutions-updater-improvements/package/WindowsManaged/Program.cs#L209-L211)to perform the downgrade. MSI-related changes will be pushed in follow-up PR, as more in-depth manual testing is required for MSI. 

Additionally, multiple clippy fixes were performed in `updater` Agent module

JIRA Ticket: https://devolutions.atlassian.net/browse/DGW-233 (WIP)

### DIFF
--- a/devolutions-agent/src/updater/mod.rs
+++ b/devolutions-agent/src/updater/mod.rs
@@ -24,10 +24,11 @@ use crate::config::ConfHandle;
 
 use integrity::validate_artifact_hash;
 use io::{download_binary, download_utf8, save_to_temp_file};
-use package::{install_package, validate_package};
+use package::{install_package, uninstall_package, validate_package};
 use productinfo::DEVOLUTIONS_PRODUCTINFO_URL;
 use security::set_file_dacl;
 
+use detect::get_product_code;
 pub(crate) use error::UpdaterError;
 pub(crate) use product::Product;
 
@@ -42,8 +43,14 @@ struct UpdaterCtx {
     conf: ConfHandle,
 }
 
+struct DowngradeInfo {
+    installed_version: DateVersion,
+    product_code: String,
+}
+
 struct UpdateOrder {
     target_version: DateVersion,
+    downgrade: Option<DowngradeInfo>,
     package_url: String,
     hash: Option<String>,
 }
@@ -169,11 +176,22 @@ async fn update_product(conf: ConfHandle, product: Product, order: UpdateOrder) 
         return Ok(());
     }
 
-    install_package(&ctx, &package_path)
+    if let Some(downgrade) = order.downgrade {
+        let installed_version = downgrade.installed_version;
+        info!(%product, %installed_version, %target_version, "Downgrading product...");
+
+        let uninstall_log_path = package_path.with_extension("uninstall.log");
+
+        uninstall_package(&ctx, downgrade.product_code, &uninstall_log_path).await?;
+    }
+
+    let log_path = package_path.with_extension("log");
+
+    install_package(&ctx, &package_path, &log_path)
         .await
         .context("failed to install package")?;
 
-    info!(%product, "Product updated to v{target_version}!");
+    info!(%product, %target_version, "Product updated!");
 
     Ok(())
 }
@@ -210,15 +228,13 @@ async fn check_for_updates(product: Product, update_json: &UpdateJson) -> anyhow
 
     trace!(%product, %detected_version, "Detected installed product version");
 
-    let is_target_version_same_as_installed = match target_version {
-        VersionSpecification::Latest => false,
-        VersionSpecification::Specific(target) => target == detected_version,
-    };
-
-    // Early exit without checking remote database.
-    if is_target_version_same_as_installed {
-        info!(%product, %detected_version, "Product is up to date, skipping update");
-        return Ok(None);
+    match target_version {
+        VersionSpecification::Specific(target) if target == detected_version => {
+            // Early exit without checking remote database.
+            info!(%product, %detected_version, "Product is up to date, skipping update");
+            return Ok(None);
+        }
+        VersionSpecification::Latest | VersionSpecification::Specific(_) => {}
     }
 
     info!(%product, %target_version, "Ready to update the product");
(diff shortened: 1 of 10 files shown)


devolutions-agent/src/updater/mod.rs
```diff
@@ -240,10 +243,38 @@ async fn check_for_updates(product: Product, update_json: &UpdateJson) -> anyhow
             }
         }
         VersionSpecification::Specific(version) => {
-            if version != remote_version {
-                warn!(%product, %version, "Product uptate target version does not match available version on devolutions.net, skipping update");
+            if version == detected_version {
+                info!(%product, %detected_version, "Product is up to date, skipping update");
                 return Ok(None);
             }
```
CBenoit, line 249: suggestion: For `VersionSpecification::Specfic`, this logic is duplicated, and can be removed.

devolutions-agent/src/updater/mod.rs
```diff
@@ -208,12 +210,13 @@ async fn check_for_updates(product: Product, update_json: &UpdateJson) -> anyhow
...
+        VersionSpecification::Latest => false,
+        VersionSpecification::Specific(target) => target == detected_version,
     };
 
-    if !is_target_version_newer {
+    // Early exit without checking remote database.
+    if is_target_version_same_as_installed {
         info!(%product, %detected_version, "Product is up to date, skipping update");
         return Ok(None);
     }
```
CBenoit, line 222: suggestion: Just to make it clear that we are excluding "latest" from this.

```rust
    match target_version {
        // Early exit without checking remote database.
        VersionSpecification::Specific(target) if target == detected_version => {
            info!(%product, %detected_version, "Product is up to date, skipping update");
            return Ok(None);
        }
        _ => {}
    }
```

devolutions-agent/src/updater/mod.rs
```diff
@@ -240,10 +243,38 @@ async fn check_for_updates(product: Product, update_json: &UpdateJson) -> anyhow
...
+                let response = reqwest::Client::builder().build()?.head(&modified_url).send().await?;
+                if !response.status().is_success() {
+                    warn!(
+                        %product,
+                        %version,
+                        %modified_url,
+                        "Modified product URL does not exist, skipping update"
+                    );
+                    return Ok(None);
+                }
```
CBenoit, line 269: suggestion: Not necessarily a 404 status, so the wording should be changed.

```suggestion
                if let Err(error) = response.error_for_status() {
                    warn!(
                        %error,
                        %product,
                        %version,
                        %modified_url,
                        "Failed to access the product URL, skipping update"
                    );

                    return Ok(None);
                }
```

devolutions-agent/src/updater/mod.rs
```diff
@@ -282,3 +313,54 @@ async fn init_update_json() -> anyhow::Result<Utf8PathBuf> {
...
+/// - Modified URL: https://cdn.devolutions.net/download/DevolutionsGateway-x86_64-2024.4.0.0.msi
+fn try_modify_product_url_version(
+    url: &str,
+    original_version: DateVersion,
+    version: DateVersion,
+) -> anyhow::Result<String> {
+    let new_url = url.replace(&original_version.to_string(), &version.to_string());
+
+    if new_url == url {
+        return Err(anyhow!("Product URL has unexpected format, version cannot be modified"));
```
CBenoit, line 334: style:
```suggestion
        return Err(anyhow!("product URL has unexpected format, version cannot be modified"));
```

devolutions-agent/src/updater/package.rs
```diff
@@ -87,9 +88,11 @@ fn ensure_enough_rights() -> Result<(), UpdaterError> {
         GetTokenInformation(
             *token_handle,
             TokenElevation,
-            Some(&mut token_elevation as *mut _ as _),
-            size_of::<TOKEN_ELEVATION>().try_into().unwrap(),
-            &mut return_size as _,
+            Some(&mut token_elevation as *mut TOKEN_ELEVATION as *mut core::ffi::c_void),
+            size_of::<TOKEN_ELEVATION>()
+                .try_into()
+                .expect("Unreachable: TOKEN_ELEVATION size always fits into u32"),
```
CBenoit, line 94: style:
```suggestion
                .expect("TOKEN_ELEVATION size always fits into u32"),
```

pacmancoder: I think it's debatable -- triggering `expect` is always either some world-breaking bug, which happened because previous assumptions about the execution path were wrong, or just simply unreachable code path, which should never be reached, I like to prepend expect's with category like `BUG:` or `Unreachable:` to quickly categorize the problem source during log analysis/debug, but if it's deal breaker, I could remove this prefix 🙂

devolutions-agent/src/updater/mod.rs
CBenoit: Feel free to adjust my suggestion with
```rust
VersionSpecification::Latest => {}
```
instead of
```rust
_ => {}
```

devolutions-agent/src/updater/mod.rs
CBenoit: Oh, I’m referring to this: https://github.com/Devolutions/devolutions-gateway/pull/1116#discussion_r1852319041

devolutions-agent/src/updater/package.rs
CBenoit: I would argue that if you get a panic with a "Unreachable:" prefix, you have a "bug", and conversely we always expect the panic branch to be "unreachable" unless there is a bug. Then, sometimes the absence of bug is obvious, sometimes it’s not, but I don’t really see how it matters once the program is crashed. I think the `Unreachable:` prefix is making the already busy line even more busy for not much. If you’re still not convinced, feel free to leave that way!

pacmancoder: Makes sense 👍

pacmancoder:
@CBenoit FYI: I made minor refactoring after review and added uninstall-install logic for downgrades

PR #1117, opened 2024-11-22
fix(pedm): audit and refactor netmgmt and wts modules

CBenoit:
Applied remaining suggestions from https://github.com/Devolutions/devolutions-gateway/pull/1114 and fixed some issues. See the comments.

### DIFF
--- a/crates/win-api-wrappers/src/memory.rs
+++ b/crates/win-api-wrappers/src/memory.rs
@@ -1,0 +1,90 @@
+use core::ffi::c_void;
+use core::marker::PhantomData;
+use core::ptr;
+
+pub trait FreeMemory {
+    /// # Safety
+    ///
+    /// `ptr` must be a pointer which must be freed by this implementation.
+    unsafe fn free(ptr: *mut c_void);
+}
+
+/// RAII wrapper for some form of memory.
+///
+/// The free function is not called if the pointer is null.
+pub struct MemoryWrapper<FreeImpl: FreeMemory, T = c_void> {
+    // INVARIANT: `ptr` is a pointer which must be freed by `FreeImpl`
+    ptr: *mut T,
+
+    _marker: PhantomData<FreeImpl>,
+}
+
+impl<FreeImpl: FreeMemory, T> MemoryWrapper<FreeImpl, T> {
+    pub const fn null() -> Self {
+        Self {
+            ptr: ptr::null_mut(),
+            _marker: PhantomData,
+        }
+    }
+
+    /// Constructs a MemoryWrapper from a raw pointer.
+    ///
+    /// # Safety
+    ///
+    /// - `ptr` must be a valid pointer.
+    /// - `ptr` must be freed by the associated free implementation `FreeImpl`.
+    pub const unsafe fn from_raw(ptr: *mut T) -> Self {
+        Self {
+            ptr,
+            _marker: PhantomData,
+        }
+    }
+
+    pub const fn as_ptr(&self) -> *const T {
+        self.ptr.cast_const()
+    }
+
+    pub fn as_mut_ptr(&mut self) -> *mut T {
+        self.ptr
+    }
+
+    /// Forms a slice from the inner pointer and a length.
+    ///
+    /// The `len` argument is the number of **elements**, not the number of bytes.
+    ///
+    /// # Safety
+    ///
+    /// - Pointer must be valid for reads for `len * mem::size_of::<T>()` many bytes, and it must be properly aligned.
+    /// - Pointer must point to `len` consecutive properly initialized values of type `T`.
+    /// - The memory referenced by the returned slice must not be mutated for the duration of the borrowing, except inside an `UnsafeCell`.
+    /// - The total size `len * mem::size_of::<T>()` of the slice must be no larger than `isize::MAX`, and adding that size to `data` must not "wrap around" the address space.
+    pub unsafe fn cast_slice(&self, len: usize) -> &[T] {
+        // SAFETY:
+        // - Same preconditions as the current function.
+        // - We are also tying the lifetime of the slice to the MemoryWrapper instance, ensuring the memory is not freed as long as it is being used.
+        unsafe { std::slice::from_raw_parts(self.ptr, len) }
+    }
+
+    pub fn cast<U>(self) -> MemoryWrapper<FreeImpl, U> {
+        let casted = MemoryWrapper {
+            ptr: self.ptr.cast::<U>(),
+            _marker: PhantomData,
+        };
+
+        // We transferred the owneship, do not call drop.
+        core::mem::forget(self);
+
+        casted
+    }
+}
+
+impl<FreeImpl: FreeMemory, T> Drop for MemoryWrapper<FreeImpl, T> {
+    fn drop(&mut self) {
+        if self.ptr.is_null() {
+            return;
+        }
+
+        // SAFETY: Per invariant on `ptr`, `ptr` is a pointer which must be freed by `FreeImpl`.
+        unsafe { FreeImpl::free(self.ptr.cast()) };
+    }
+}
--- a/crates/win-api-wrappers/src/netmgmt.rs
+++ b/crates/win-api-wrappers/src/netmgmt.rs
@@ -11,7 +11,7 @@ use crate::utils::WideString;
 use crate::Error;
 
 pub fn add_local_group_member(group_name: &str, security_identifier: &Sid) -> anyhow::Result<()> {
-    // SAFETY: It is safe to zero out the structure as it is a simple POD type.
+    // SAFETY: The structure is a simple POD type.
     let mut group_info = unsafe { core::mem::zeroed::<LOCALGROUP_MEMBERS_INFO_0>() };
 
     let group_name = WideString::from(group_name);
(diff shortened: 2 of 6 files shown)


crates/win-api-wrappers/src/memory.rs
CBenoit, line 1: I created an abstraction over the repeating pattern we had with `NetMgmtMemory` and `WtsMemory`

crates/win-api-wrappers/src/wts.rs
```diff
@@ -229,30 +232,26 @@ pub fn send_message_to_session(
...
-            &mut string_ptr,
-            &mut len,
-        )?;
+        WTSQuerySessionInformationW(WTS_CURRENT_SERVER_HANDLE, session_id, info_class, &mut string, &mut len)?;
     }
 
-    // SAFETY: `string_ptr` is always a valid pointer on success.
-    let _ = unsafe { WTSMemory::from_raw(string_ptr.0.cast()) };
+    // SAFETY: On success, WTSQuerySessionInformationW places a valid pointer which must be freed by `WTSFreeMemory`.
+    let _guard = unsafe { WtsMemory::from_raw(string.as_ptr()) };
```
CBenoit, line 250: `let _ = ` will cause the value to be immediately dropped.

See in [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6c1de0468ef41093c81aa61658052861)

crates/win-api-wrappers/src/wts.rs
```diff
@@ -20,78 +19,77 @@ use crate::ui::MessageBoxResult;
 use crate::utils::{AnsiString, SafeWindowsString};
 
 /// RAII wrapper for WTS virtual channel handle.
-pub struct WTSVirtualChannel(HANDLE);
+pub struct WtsVirtualChannel(HANDLE);
```
CBenoit, line 22: Rust naming convention:

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. 

https://rust-lang.github.io/api-guidelines/naming.html

crates/win-api-wrappers/src/netmgmt.rs
```diff
@@ -60,74 +58,61 @@ pub fn get_local_admin_group_members() -> anyhow::Result<Vec<Sid>> {
...
+        unsafe { NetApiBufferFree(Some(ptr)) };
     }
 }
 
-impl Default for NetMgmtMemory {
-    fn default() -> Self {
-        Self(std::ptr::null_mut())
-    }
-}
+type NetmgmtMemory<T = core::ffi::c_void> = crate::memory::MemoryWrapper<NetmgmtFreeMemory, T>;
```
CBenoit, line 118: Rust naming convention:

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. 

https://rust-lang.github.io/api-guidelines/naming.html

crates/win-api-wrappers/src/netmgmt.rs
```diff
@@ -60,74 +58,61 @@ pub fn get_local_admin_group_members() -> anyhow::Result<Vec<Sid>> {
...
+    // SAFETY:
+    // - group_name holds a null-terminated UTF-16 string, and as_pcwstr() returns a valid pointer to it.
+    // - Specifying `MAX_PREFERRED_LENGTH` allocates the required amount of memory for the data, and the function will not return `ERROR_MORE_DATA`.
+    let ret = unsafe {
+        NetLocalGroupGetMembers(
             None,
             group_name.as_pcwstr(),
             0,
-            &mut group_members_ptr.cast(),
+            &mut group_members,
```
CBenoit, line 73: Using `.cast()` or `as *mut _` to convert the pointer will create a temporary value that we then immediately borrow and _drop_. The original variable is left unchanged. We need work with the right type from the start, and convert _afterwards_.

Working example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=de03e34c057c117ab745625c5acca955

Buggy example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7a61ab03938a721b8aeebf97bf0a7322

crates/win-api-wrappers/src/wts.rs
```diff
@@ -143,40 +141,45 @@ pub struct WTSSessionInfo {
...
-    }
+    // SAFETY: The pointer placed into `sessions` by `WTSEnumerateSessionsW` must be freed by `WTSFreeMemory`.
+    let sessions = unsafe { WtsMemory::from_raw(sessions) };
+
+    // SAFETY:
+    // - `WTSEnumerateSessionsW` returns a pointer valid for `count` reads of `WTS_SESSION_INFOW`.
+    // - `WTSEnumerateSessionsW` is also returning a pointer properly aligned.
+    // - We ensure the memory referenced by the slice is not mutated by shadowing the variable.
+    // - There are never so many sessions that `count * mem::size_of::<WTS_SESSION_INFOW>()` overflows `isize`.
+    let sessions = unsafe { sessions.cast_slice(count as usize) };
```
CBenoit, line 174: `cast_slice` is tying the lifetime of the slice to the lifetime of the WtsMemory object, preventing use-after-free bugs.

crates/win-api-wrappers/src/wts.rs
```diff
@@ -143,40 +141,45 @@ pub struct WTSSessionInfo {
...
-    /// - The `pWinStationName` pointer needs to be valid for reads up until and including the next `\0`.
+    /// - The `wts_session.pWinStationName` pointer must be valid for reads up until and including the next `\0`.
     unsafe fn new(wts_session: &WTS_SESSION_INFOW) -> anyhow::Result<Self> {
+        // SAFETY: Same preconditions as the current function.
+        let win_station_name = unsafe { wts_session.pWinStationName.to_string()? };
+
         Ok(Self {
             session_id: wts_session.SessionId,
-            win_station_name: unsafe { wts_session.pWinStationName.to_string()? },
+            win_station_name,
```
thenextman, line 151: Question: you removed the unsafe block but this triggers [unsafe-op-in-unsafe-fn](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unsafe-op-in-unsafe-fn) for me

Is my rust-analyzer maybe giving me a false result based on project settings?

_Or_ is it just something to ignore?

thenextman (approved):
Thanks!

crates/win-api-wrappers/src/wts.rs
CBenoit: Are you still getting a warning?? I didn’t remove the unsafe block, just moved it above:
```rust
   unsafe fn new(wts_session: &WTS_SESSION_INFOW) -> anyhow::Result<Self> {
        // SAFETY: Same preconditions as the current function.
        let win_station_name = unsafe { wts_session.pWinStationName.to_string()? };
```
So the warning should not be fired normally :thinking:

thenextman: My bad. Need to finish my coffee before looking 😂

crates/win-api-wrappers/src/wts.rs
```diff
@@ -20,78 +19,77 @@ use crate::ui::MessageBoxResult;
...
     pub fn open_dvc(name: &str) -> anyhow::Result<Self> {
         let channel_name = AnsiString::from(name);
 
-        // SAFETY: Channel name is always a valid pointer to a null-terminated string.
-        let raw_wts_handle = unsafe {
-            WTSVirtualChannelOpenEx(WTS_CURRENT_SESSION, channel_name.as_pcstr(), WTS_CHANNEL_OPTION_DYNAMIC)
-        }?;
+        // SAFETY: `channel_name.as_pcstr()` is always a valid pointer to a null-terminated string.
+        let wts_handle = unsafe {
+            WTSVirtualChannelOpenEx(WTS_CURRENT_SESSION, channel_name.as_pcstr(), WTS_CHANNEL_OPTION_DYNAMIC)?
```
pacmancoder, line 37: Question(style): I was always wondering, which is the best -- to place `?` operator inside `unsafe` block or outside. I usually lean towards placing it outside, because `try` operation by itself is not unsafe, and thus could be moved out -- however from a readability standpoint it looks better if `?` is inside the block. What is your spin on this, @CBenoit ?

crates/win-api-wrappers/src/wts.rs
pacmancoder: Wow, I didn't know that, you live and learn... 😅 IMHO, that **huge** difference on such implicit thing gives me C++ vibes 🤔

crates/win-api-wrappers/src/wts.rs
CBenoit: I think it’s better to keep the ? inside the unsafe block, just at the end of the function call. That’s how we usually do in safe Rust, and as you said, the `try` operation is not unsafe, and will never be, so we’re probably not going to be confused about. We also have the `clippy::multiple_unsafe_ops_per_block` lint to avoid inadvertently calling more than one unsafe operation in the same block. If we follow the logic to put the `?` outside, then why not put things like `as_pcstr()` outside as well? I think it would be too much :thinking:
Of course there is a balance to strike for. In this case, the outermost function `WTSVirtualChannelOpenEx` is (I think) obviously the unsafe operation.
But when we perform pointer dereferencing, it may be worth spilling on multiple lines.

pacmancoder: > If we follow the logic to put the ? outside, then why not put things like as_pcstr() outside as well? I think it would be too much

Fair enough 👍 

> But when we perform pointer dereferencing, it may be worth spilling on multiple lines.

Yep, agree. I think Clippy will catch that as multiple unsafe operations in a single `unsafe` block iirc

crates/win-api-wrappers/src/wts.rs
pacmancoder: Yeah, the lint definitely would be useful in such cases

PR #1122, opened 2024-11-27
feat(dgw): add agent version field to heartbeat API

pacmancoder:


CBenoit:
The CI is red for a few jobs.

> I was planning to change the following versions:
> 
>     * `devolutions-gateway\openapi\ts-angular-client\config.json`: 0.13.0 -> 0.14.0
> 
>     * `devolutions-gateway\openapi\dotnet-client\config.json`: 2024.11.12 -> 2024.11.27
> 
>     * `devolutions-gateway\openapi\dotnet-subscriber\config.json` `2023.2.22.0` -> `2024.11.27` (version looks stale, ignore this instead?)
>       -Also, I see `gateway-api.yaml`, `subscriber-api.yaml`, updated automatically
> 
>     * `devolutions-gateway\openapi\doc\config.json` - There is no version in file, I guess version `index.adoc` should be updated automatically from crate version too?

That’s correct! You can ignore the subscriber. We may delete it completely at some point, as no one is actually using it in fact.

crates/devolutions-agent-shared/Cargo.toml
```diff
@@ -15,5 +15,9 @@ cfg-if = "1"
 serde = { version = "1", features = ["derive"] }
 thiserror = "1"
 
+[target.'cfg(windows)'.dependencies]
+winreg = "0.52"
```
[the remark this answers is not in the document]
pacmancoder: Hm, there is no specific reason actually, I just used it previously - the first which came to mind 🙂 Looks like a nice opportunity to switch to an official one during refactoring, thanks!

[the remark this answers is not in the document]
pacmancoder: UPD: switched over `windows-registry`

pacmancoder:
@CBenoit Updated PR, split up commits for easier review (package version update produced huge diff)

pacmancoder:
UPD: Do not merge yet, switching crate has proven to be not so smooth during testing, stand by

crates/devolutions-agent-shared/src/windows/mod.rs
```diff
@@ -7,4 +7,4 @@ pub use uuid::UuidError;
 /// MSI upgrade code for the Devolutions Gateway.
 pub const GATEWAY_UPDATE_CODE: &str = "{db3903d6-c451-4393-bd80-eb9f45b90214}";
 /// MSI upgrade code for the Devolutions Agent.
-pub const AGENT_UPDATE_CODE: &str = "{c3d81328-f118-4d5d-9a82-b7c31b076755}";
+pub const AGENT_UPDATE_CODE: &str = "{82318d3c-811f-4d5d-9a82-b7c31b076755}";
```
pacmancoder, line 10: Note to myself: Never trust copilot to autocomplete constants 🤣

crates/devolutions-agent-shared/src/date_version.rs
```diff
@@ -18,6 +18,12 @@ pub struct DateVersion {
     pub revision: u32,
 }
 
+impl DateVersion {
+    pub fn to_string_without_revision(&self) -> String {
```
CBenoit, line 22: naming (non-blocking): What about this naming? Feel free to ignore if you prefer the current one!
```suggestion
    pub fn fmt_without_revision(&self) -> String {
```

crates/devolutions-agent-shared/src/windows/mod.rs
```diff
@@ -0,0 +1,10 @@
+mod uuid;
+
+pub mod registry;
+
+pub use uuid::UuidError;
+
+/// MSI upgrade code for the Devolutions Gateway.
+pub const GATEWAY_UPDATE_CODE: &str = "{db3903d6-c451-4393-bd80-eb9f45b90214}";
+/// MSI upgrade code for the Devolutions Agent.
+pub const AGENT_UPDATE_CODE: &str = "{c3d81328-f118-4d5d-9a82-b7c31b076755}";
```
CBenoit: question: Why the terminology used is different? Upgrade vs Update? Is it also used for downgrades? What exactly these IDs are representing? Is a unique ID for the MSI?

crates/devolutions-agent-shared/src/windows/registry.rs
```diff
@@ -0,0 +1,86 @@
...
+        Err(_) => return Ok(None),
+    };
+
+    // Product code is the name of the only value in the registry key.
+    let (product_code, _) = match update_code_key.enum_values().next() {
+        Some(value) => value.map_err(|source| RegistryError::EnumKeyValues { key: key_path, source })?,
+        None => return Ok(None),
+    };
+
+    Ok(Some(reversed_hex_to_uuid(&product_code)?))
```
CBenoit, line 46: question: What about working with the `Uuid` type instead of `String`s? Is there a reason for not doing that?

crates/devolutions-agent-shared/src/windows/registry.rs
```diff
@@ -0,0 +1,86 @@
...
+    let short_year = (product_version >> 24) + 2000;
+    let month = (product_version >> 16) & 0xFF;
+    let day = product_version & 0xFFFF;
+
+    Ok(Some(DateVersion {
+        year: short_year,
+        month,
+        day,
+        // NOTE: Windows apps could only have 3 version numbers (major, minor, patch).
+        revision: 0,
```
CBenoit, line 88: note: Hopefully should not cause any problem because we don’t use this number ourselves, but if we end up using the revision number one day, we’ll be basically reporting an older version regardless of what was actually installed… :thinking:
We may want to use a different approach at some point.

crates/devolutions-agent-shared/src/windows/registry.rs
CBenoit: suggestion: For the hex-reversed representation, you could define a wrapper type too.
See this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=641a51af984214092e0c365d07c0af3e

crates/devolutions-agent-shared/src/windows/uuid.rs
CBenoit, line 1: suggestion: I think the whole logic here could be made simpler and more robust. See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=641a51af984214092e0c365d07c0af3e

devolutions-gateway/src/api/heartbeat.rs
```diff
@@ -6,6 +6,7 @@ use crate::extract::HeartbeatReadScope;
 use crate::http::HttpError;
 use crate::DgwState;
 use devolutions_agent_shared::windows::registry::get_installed_product_version;
+#[cfg(windows)]
 use devolutions_agent_shared::windows::AGENT_UPDATE_CODE;
```
CBenoit: Instead of putting many `#[cfg(windows)]` inside the heartbeat module of the Gateway, could use expose a cross-platform API from `devolutions_agent_shared`, returning `None` on all platforms except Windows (for now)?

crates/devolutions-pedm/openapi/pedm-api.yaml
CBenoit, line 1: suggestion: Let’s update the OpenAPI client in a separate PR!

CBenoit:
I’m done with the review! @pacmancoder

crates/devolutions-agent-shared/src/windows/registry.rs
pacmancoder: Windows "reversed UUID" representation, as I call it, is pretty strange, and reverses UUID byte groups of different size per-hex character, not per-byte, e.g:
`C3D81328F118D5D4A9287B3CB1707655` => `{82318d3c-811f-4d5d-9a82-b7c31b076755}`
`C3D81328 => 82318d3c`
`F118` => `811f`

So we can't use from_str_radix with base16 here, and if we use UUID instead of string, then we still need to format `Uuid` to `String`,

I am not sure that switching to `uuid` will improve things here - right after we parsed UUID from string read from registry, we need to immediately format it to string back for conversion (or reverse in binary splitting hex data on 4-bit nibbles which is even more messy) 🤔

crates/devolutions-agent-shared/src/windows/uuid.rs
pacmancoder: see https://github.com/Devolutions/devolutions-gateway/pull/1122#discussion_r1862368113

devolutions-gateway/src/api/heartbeat.rs
pacmancoder: Good idea, I moved cfg to `devolutions_agent_shared`

crates/devolutions-pedm/openapi/pedm-api.yaml
pacmancoder: reverted commit

crates/devolutions-agent-shared/src/windows/registry.rs
pacmancoder: Discussed in slack, switched to using Uuid for non-reversed form

crates/devolutions-agent-shared/src/windows/mod.rs
pacmancoder: done

pacmancoder:
@CBenoit please re-review

crates/devolutions-agent-shared/src/lib.rs
```diff
@@ -11,6 +13,16 @@ use cfg_if::cfg_if;
...
 
+#[cfg(windows)]
+pub fn get_installed_agent_version() -> Result<Option<DateVersion>, windows::registry::RegistryError> {
+    windows::registry::get_installed_product_version(windows::AGENT_UPDATE_CODE)
+}
+
+#[cfg(not(windows))]
+pub fn get_installed_agent_version() -> Result<Option<DateVersion>, Infallible> {
+    Ok(None)
+}
```
CBenoit, line 24: suggestion: It’s typically safer to not return a different error type to avoid inconsistency (and ending up with cfgs again).

devolutions-gateway/src/api/heartbeat.rs
```diff
@@ -15,6 +16,9 @@ pub(crate) struct Heartbeat {
     hostname: String,
     /// Gateway service version
     version: &'static str,
+    /// Agent version (If installed)
```
CBenoit, line 19:
```suggestion
    /// Agent version, if installed
```

crates/devolutions-agent-shared/src/windows/reversed_hex_uuid.rs
```diff
@@ -0,0 +1,90 @@
+//! Windows-specific UUID format conversion functions.
+
+use std::str::FromStr;
+use uuid::Uuid;
+
+const UUID_CHARS: usize = uuid::fmt::Simple::LENGTH;
```
CBenoit, line 6: naming:
```suggestion
const REVERSED_UUID_STR_LENGTH: usize = uuid::fmt::Simple::LENGTH;
```

crates/devolutions-agent-shared/src/lib.rs
```diff
@@ -1,6 +1,10 @@
 mod date_version;
 mod update_json;
+#[cfg(windows)]
+pub mod windows;
```
CBenoit, line 4: suggestion: We likely don’t need this to be public anymore.
```suggestion
#[cfg(windows)]
mod windows;
```

CBenoit (approved):
LGTM. I left a few last minor comments. Feel free to merge yourself once addressed!

CBenoit:
I updated the PR title to `feat(dgw): add agent version field to heartbeat API` to follow our convention.

crates/devolutions-agent-shared/src/lib.rs
pacmancoder: We still need this to be public as the Agent uses product codes to invoke msiexec, and we can't make MSI product codes cross-platform in shared library, lets keep it as-is for now

CBenoit:
FYI @pacmancoder
Added https://github.com/Devolutions/devolutions-gateway/labels/publish-required

PR #1124, opened 2024-11-29
chore: regenerate OpenAPI clients

pacmancoder:


crates/devolutions-pedm/openapi/pedm-api.yaml
```diff
@@ -342,7 +342,7 @@ components:
       properties:
         Seconds:
           type: integer
-          format: int64
+          format: uint64
```
CBenoit, line 345: question: Why all the integer types are changed?

CBenoit (approved):
LGTM! Not sure what happened with the PEDM OpenAPI document though.

crates/devolutions-pedm/openapi/pedm-api.yaml
pacmancoder: I don't know either, I just re-generated packages as described in `README.md` 🤔

PR #1132, opened 2024-12-04
chore(openapi): regenerate the clients

irvingoujAtDevolution:
This PR contains only changes introduced by generate_all.ps1 script


webapp/player/js/gateway.js
CBenoit, line 1: You included changes in the player which are not the result of the openapi clients being regenerated. Please open two separate PRs.

irvingoujAtDevolution: updated! sorry I was coding on unmerged commit, I rebased in on master and it should now be good!

CBenoit:
The commit message is also typically "chore(openapi): regenerate the clients". The actual API changes are part of the service-side PRs.

PR #1136, opened 2024-12-06
fix(dgw): fix recording remux failing when streaming

irvingoujAtDevolution:
The recording remux task would fail because the the file cannot be overwrite as streamers are holding it, streamers cannot finish themselves as they have no information about weather the recording is finished or not. Added a subscriber to recording manager so that the recording task can signal the streaming task that the recording has ended, as well as the streamer could tell the recording that they have released the file.

Changelog: ignore


crates/streamer/src/streamer/mod.rs
```diff
@@ -213,7 +281,7 @@ fn spawn_sending_task<W>(
         let task_result = task.await;
 
         if let Err(e) = task_result {
-            tracing::error!("Error while sending data: {:?}", e);
+            tracing::warn!(error=?e);
```
CBenoit, line 284: note: For anyhow::Error, you should use the alternate formatting of Display. `error = format!("{e:#}")`. If it’s not anyhow::Error, just use the normal Display `error = %e`.

webapp/shadow-player/src/websocket.ts
```diff
@@ -15,8 +20,13 @@ export class ServerWebSocket {
       const reader = new FileReader();
       reader.onload = () => {
         const arrayBuffer = reader.result as ArrayBuffer;
-        const serverResponse = parseServerMessage(arrayBuffer);
-        callback(serverResponse);
+        try {
```
irvingoujAtDevolution, line 23: prepare streamer player for future features.

CBenoit:
General request: read the STYLE.md file and try to follow the conventions.

crates/streamer/src/streamer/mod.rs
irvingoujAtDevolution: updated

irvingoujAtDevolution: I think I got them all, I made every `?e` into `error=%e` if display implemented, otherwise "error=?e", and I made anyhow errors to "error = format!("{e:#}")". Also made the error message start with lowercase letters for error and warn, used uppercase letters for info! and debug!

CBenoit: Great, thank you!

crates/streamer/src/streamer/mod.rs
```diff
@@ -142,69 +155,124 @@ pub fn webm_stream(
...
     let task = async move {
         info!("Starting streaming task");
-
         loop {
             tokio::select! {
                 err = error_receiver.recv() => {
                     if let Some(err) = err {
-                        ws_frame.send(protocol::ServerMessage::Error(err)).await?;
+                        let _ = ws_frame_clone.lock().await.send(protocol::ServerMessage::Error(err)).await.inspect_err(|e| {
+                            warn!(error=?e, "failed to send error message");
```
CBenoit, line 239: Error messages (the thing shown by the Display implementation of error types) should be lowercased: https://github.com/Devolutions/IronRDP/blob/4b1dbaa9102c6506d8c5124d3e8a5eb0d79efd5e/STYLE.md#formatting-of-error-messages

The logging messages are not intended to be composed and may be upper cased for consistency: https://github.com/Devolutions/IronRDP/blob/4b1dbaa9102c6506d8c5124d3e8a5eb0d79efd5e/STYLE.md#logging

CBenoit:
I’ll go through the PR on Monday!

webapp/shadow-player/src/websocket.ts
CBenoit: I don’t understand how it’s "preparing". Typically I would suggest to do the changes when they are actually necessary, but let me know.

crates/streamer/src/streamer/mod.rs
irvingoujAtDevolution: Thank you! I updated them all again. I misunderstood it!

webapp/shadow-player/src/websocket.ts
irvingoujAtDevolution: You are right, I'll remove this part as well.

crates/streamer/src/streamer/protocol.rs
```diff
@@ -2,6 +2,7 @@ use tokio_util::{
     bytes::{self, Buf, BufMut},
     codec,
 };
+use tracing::info;
```
CBenoit, line 5: suggestion:
https://github.com/Devolutions/devolutions-gateway/blob/68f0523118593285198cff35a3c4edffe6632eaa/devolutions-gateway/src/lib.rs#L10-L11

CBenoit (commented):
> streamers cannot finish themselves as they have no information about weather the recording is finished or not

question: Don’t we have the information of whether the connection (TCP stream, etc) is EOF via the AsyncWriter / AsyncRead traits?

irvingoujAtDevolution:
> > streamers cannot finish themselves as they have no information about weather the recording is finished or not
> 
> question: Don’t we have the information of whether the connection (TCP stream, etc) is EOF via the AsyncWriter / AsyncRead traits?

Oh, do we? I am not aware of it. However remux task still need to find a way to wait for streaming tasks to release the file before proceeding.

CBenoit:
Since this is not relevant for the changelog (fix for an unreleased bug), I added the footer "Changelog: ignore".

crates/streamer/src/streamer/channel_writer.rs
```diff
@@ -37,3 +37,14 @@ impl io::Write for ChannelWriter {
         Ok(())
     }
 }
+
+#[derive(Debug)]
+pub(crate) struct ChannelWriterReceiver {
```
irvingoujAtDevolution, line 42: ChannelWriterReceiver  was introduced to resolve the issue where the main loop is not terminating, however it turns not this is not needed. I kept it anyway because it makes code cleaner.

crates/streamer/src/streamer/channel_writer.rs
```diff
@@ -37,3 +37,14 @@ impl io::Write for ChannelWriter {
...
+    receiver: tokio::sync::mpsc::Receiver<Vec<u8>>,
+}
+
+impl ChannelWriterReceiver {
+    pub(crate) async fn recv(&mut self) -> Result<Option<Vec<u8>>, ChannelWriterError> {
+        Ok(self.receiver.recv().await)
+    }
```
CBenoit, line 49: suggestion:
```suggestion
    pub(crate) async fn recv(&mut self) -> Option<Vec<u8>> {
        self.receiver.recv().await
    }
```

crates/streamer/src/streamer/mod.rs
```diff
@@ -59,30 +62,51 @@ pub fn webm_stream(
...
+                _ = when_new_chunk_appended_receiver => {
+                    let _ = tx.send(WhenEofControlFlow::Continue);
+                },
+                _ = stop_notifier.notified() => {
+                    let _ = tx.send(WhenEofControlFlow::Break);
+                }
+            }
+        });
+        rx.blocking_recv()
+    }
```
CBenoit, line 108: style: Put the function at the end.
https://github.com/Devolutions/IronRDP/blob/054f812f2e81d45737695fc3ee20f1465a31bd40/STYLE.md#helper-functions

devolutions-gateway/src/recording.rs
```diff
@@ -228,6 +228,10 @@ enum RecordingManagerMessage {
         id: Uuid,
         session_must_be_recorded: bool,
     },
+    SubscribeToSessionEndNotification {
+        id: Uuid,
+        channel: oneshot::Sender<anyhow::Result<Arc<Notify>>>,
```
CBenoit, line 233: suggestion: It’s fine to not return anything when there is an error.

```suggestion
        channel: oneshot::Sender<Arc<Notify>>,
```

See how we handle that for the Connect message:

https://github.com/Devolutions/devolutions-gateway/blob/31e4ae426f2de5fe8735a36f4d32072ebc7c2165/devolutions-gateway/src/recording.rs#L729-L736

crates/streamer/src/streamer/mod.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/recording.rs
irvingoujAtDevolution: updated

crates/streamer/src/streamer/mod.rs
```diff
@@ -182,6 +163,25 @@ pub fn webm_stream(
...
     result
```
CBenoit: That’s not what is described in the link.

```rust
return result;

fn helper() {}
```

devolutions-gateway/src/recording.rs
```diff
@@ -722,6 +768,14 @@ async fn recording_manager_task(
                             );
                         }
                     },
+                    RecordingManagerMessage::SubscribeToSessionEndNotification {id, channel } => {
+                        match manager.subscribe(id) {
+                            Ok(notifier) => {
+                                let _ = channel.send(notifier);
+                            },
+                            Err(e) => error!(error = format!("{e:#}"), "subscribe to session end notification"),
```
CBenoit, line 776:
```suggestion
                            Err(e) => error!(error = format!("{e:#}"), "Subscribe to session end notification"),
```

CBenoit (approved):
Thank you!

PR #1137, opened 2024-12-06
fix(jetsocat): (also) return one link per certificate

CBenoit:
Make jetsocat doctor return one link per certificate in addition to the chain link.

Issue: DGW-235


pacmancoder (approved):
LGTM!

PR #1143, opened 2024-12-11
fix(dgw): fix streaming with RDM initiated sessions

irvingoujAtDevolution:
This one is waiting for #1136 and https://github.com/austinleroy/ebml-iterable/pull/12 to be merged first. It's not ready, don't review it yet.

Edit:
It's ready


crates/streamer/Cargo.toml
```diff
@@ -21,6 +21,7 @@ webm-iterable = { version = "0.6", features = ["futures"] }
 cadeau = { version = "0.5", features = ["dlopen"] }
 thiserror = "2"
 num_cpus = "1.16.0"
+ebml-iterable = "0.6.3"
```
CBenoit, line 24:
```suggestion
ebml-iterable = "0.6"
```

crates/streamer/src/streamer/iter.rs
```diff
@@ -3,7 +3,7 @@ use std::io::Seek;
 use anyhow::Context;
 use cadeau::xmf::vpx::is_key_frame;
 use thiserror::Error;
-use tracing::debug;
+use tracing::{debug, trace};
```
CBenoit, line 6: Did you use `#[macro_use]` above `extern crate tracing` in lib.rs?

crates/streamer/src/lib.rs
```diff
@@ -7,4 +7,5 @@ pub use config::StreamingConfig;
 pub use streamer::reopenable_file::ReOpenableFile;
 pub use streamer::signal_writer::SignalWriter;
 pub use streamer::{webm_stream, Signal};
-use webm_iterable as _;
+// We need ebml_iterable for TagIterator::emit_master_end_when_eof method
+use ebml_iterable as _;
```
CBenoit, line 11: I don't see anything imported from this crate. Am I missing something?

irvingoujAtDevolution: we depends on webm_iterable which depends on embm_iterable 0.6.2 that does not have emit_master_end_when_eof method. So we manually include latest embm_iterable for that

crates/streamer/src/streamer/iter.rs
irvingoujAtDevolution: I can do this in separate PR

crates/streamer/src/lib.rs
CBenoit: You can update the Cargo.lock with `cargo update -p ebml_iterable` instead. It’s fine to remove the dependency from our Cargo.toml.

CBenoit (approved):
LGTM!

PR #1145, opened 2024-12-12
fix(pedm): properly localize UI string values in the shell extension

thenextman:


awakecoding (approved):

crates/devolutions-pedm-shell-ext/src/lib_win.rs
```diff
@@ -288,10 +301,22 @@ enum ChannelCommand {
     Elevate(PathBuf),
 }
 
+static mut DLL_MODULE: HINSTANCE = HINSTANCE(std::ptr::null_mut());
```
CBenoit, line 304: suggestion: I think the benefit of using a `static mut` is negligible over using a synchronized primitive such as `Mutex` and remove the unsafe blocks related to reading and writing to a static mut variable.

crates/win-api-wrappers/src/user.rs
```diff
@@ -0,0 +1,30 @@
...
+    let mut resource_ptr: PWSTR = PWSTR::null();
+
+    // SAFETY:
+    // No preconditions. An invalid `hinstance` or `resource_id` will cause `LoadStringW` to return 0.
+    // We pass a null-initialized mutable reference to `resource_ptr`. This is safe because:
+    // - `resource_ptr` is a properly aligned, initialized `PWSTR` variable that is of sufficient length to hold a pointer.
+    // - if `cchBufferMax` is equal to 0, `LoadStringW` guarantees to write a valid pointer to the start of a resource string or leave it as null if no resource is found.
+    // The use of `std::mem::transmute` here provides compatibility between `&mut PWSTR` and the raw pointer type expected by `LoadStringW`.
+    // - This is sound because `&mut PWSTR` has the same memory layout as the pointer type expected by `LoadStringW`.
+    let length = unsafe { LoadStringW(hinstance, resource_id, std::mem::transmute(&mut resource_ptr), 0) };
```
CBenoit, line 14: note: It’s suspicious to use transmute here, but I see that the signature in the [windows crate](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/WindowsAndMessaging/fn.LoadStringW.html) and the [Win32 doc](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadstringw) is not matching:

image
image

A problem of the windows crate?

I recommend to specify the generic parameters of transmute. I recommend to read this lint description:

- https://rust-lang.github.io/rust-clippy/master/index.html#missing_transmute_annotations

CBenoit (approved):
LGTM! I left a few minor comments that you may address before merging.

crates/win-api-wrappers/src/user.rs
thenextman: I don't think the `windows` crate is wrong _per se_ because `PWSTR` and `LPWSTR` are synonymous; they both resolve to a pointer to `WCHAR`.

What we see here is a quirk of Win32; where `LoadString*` has two modes of operation:

- Allocate a `WCHAR` of the size you want, pass the length in `cchBufferMax`; and if the resource is found, it will be copied into the buffer. Be careful of null termination and etc
- Pass "0" for `cchBufferMax` and the address of a `wchar_t*` to `lpBuffer`. and if the resource is found, you get back a read-only pointer to the string resource itself (not necessarily null-terminated, but we have the length in characters in the return value).

The second case obviously has a type mismatch. In C++ you would do something like this:

```
const wchar_t* pchStringBegin = nullptr;
int cchStringLength = ::LoadStringW(
        hInstance, 
        resourceId, 
        reinterpret_cast<PWSTR>(&pchStringBegin), 
        0);
```

I will look into the generic parameters for transmute, so thanks for that tip. But overall I'm hoping that the safety comment sufficiently explains the use of transmute here.

crates/devolutions-pedm-shell-ext/src/lib_win.rs
thenextman: question: why synchronize access when it's not needed in this case? Is it just so we don't need the `unsafe` blocks? If that's the case, I'm happy to change it, but I wonder if there's more reasoning or a more suitable primitive.

CBenoit: It’s just so that we don’t have to allocate brain power to verifying that the code is valid (= to remove the unsafe blocks to review and to audit in case of problem).
In practice, the cost of the synchronisation here is just an atomic operation to check that the lock is not held by anything else, there is no syscall unless the synchronisation was actually necessary. Since this is not a busy loop, it’s practically free and I wouldn’t bother with unsafe code for so little. 

Typically https://doc.rust-lang.org/beta/std/sync/struct.LazyLock.html or https://doc.rust-lang.org/beta/std/sync/struct.OnceLock.html would be more appropriate primitive (initialize once, and use many times), but the cost is the same (an atomic operation to check that the value is initialized).

Alternatively, if the value is initialized and used on the same thread, you can use a thread local variable and a Cell:
- https://doc.rust-lang.org/std/macro.thread_local.html
- https://doc.rust-lang.org/std/cell/struct.Cell.html

This way there is no synchronization.
(There are also LazyCell and OnceCell.)

crates/win-api-wrappers/src/user.rs
CBenoit: Thank you for the explanation. Yes it’s good I think! The suspicious part is that it’s typically not a good idea to use `transmute` to forcibly change the type of something and there exist many better way of casting, but for this API very Rustophobic I don’t think we can do any better! Specifying the generic parameters will be the only thing I would ask for.

crates/devolutions-pedm-shell-ext/src/lib_win.rs
thenextman: Ok, maybe you can short circuit some research for me here :) 

Actually `OnceLock` looks like a good fit, but regardless I cannot use it or any `Mutex` because the underlying type of `HMODULE` / `HINSTANCE` is `c_void`.

At least, the only way(s) I can see to make it work, also use unsafe code....

CBenoit: Oh I see. It would have worked if the windows crate implemented Sync and Send on HMODULE, but I understand it does not. In that case you may as well keep the current code!

thenextman: 😞

thenextman:
> LGTM! I left a few minor comments that you may address before merging.

PR updated

PR #1146, opened 2024-12-13
ci: set default shell at workflow level

allan2:


CBenoit:
I think I would rather not change the default shell given that we don't always use PowerShell, even if we use it a lot (maybe indeed a bit more than bash?). With this change the CI workflow is currently broken because bash was expected to be the default and it was not. Instead of specifying pwsh, we will specify bash explicitly. We will have to verify all our workflows and check if pwsh is appropriate or not. I think we are also used to the defaults across all our workflows in all our repositories, so it may end up being a bit of a footgun (granted it's easy to fix). I'm willing to be convinced though.

thenextman:
I'm inclined to agree with @CBenoit. If there's a good argument for making `pwsh` the default, I'm happy to listen, but this change subverts my expectations.

I've internalized the fact that Windows runners will be using PowerShell; and for anything remotely complex on Linux or macOS I'll specify and use PowerShell. But my expectation on those platforms is that the default will be bash and this is likely to cause frustration.

Like I said - and as Benoit wrote - I'm happy to hear otherwise, but currently I think the benefit is negligible versus the cost (_runs workflow that takes 40 minutes then fails because of syntax error..._ "oh darn! The Linux runners in this repo default to PowerShell!" 😮 )

PR #1149, opened 2024-12-16
Update versioning

thenextman:
Set the baseline version for the PEDM shell extension and .msix, PEDM desktop application and PEDM .NET library. Update the [bump-version](./tools/bump-version.ps1) script to keep the values consistent.

Note that I had to regenerate the API client for the PEDM .NET library, so I recommend looking at the individual commits here.

Changelog: ignore


crates/devolutions-pedm-shell-ext/AppxManifest.xml
```diff
@@ -10,10 +10,10 @@
     xmlns:uap10="http://schemas.microsoft.com/appx/manifest/uap/windows10/10"
     xmlns:com="http://schemas.microsoft.com/appx/manifest/com/windows10"
     IgnorableNamespaces="uap uap2 uap3 rescap desktop desktop4 desktop5 uap10 com">
-    <Identity Name="DevolutionsAgent" ProcessorArchitecture="neutral" Publisher="CN=Devolutions"
-        Version="1.0.0.0" />
+    <Identity Name="DevolutionPEDMShellExtension" ProcessorArchitecture="neutral" Publisher="CN=Devolutions"
+        Version="2024.3.6.0" />
```
awakecoding, line 14: isn't this updated in the CI, or is my memory incorrect and I never scripted that?

crates/devolutions-pedm-shell-ext/AppxManifest.xml
```diff
@@ -10,10 +10,10 @@
...
     <Properties>
-        <DisplayName>Devolutions Agent Shell Extension</DisplayName>
+        <DisplayName>Devolutions Agent</DisplayName>
```
awakecoding, line 16: should we call it Devolutions Agent without "shell extension" since it's not actually Devolutions Agent, but the shell extensions for Devolutions Agent?

awakecoding (commented):
just a few questions for the AppxManifest.xml

crates/devolutions-pedm-shell-ext/AppxManifest.xml
thenextman: I see the msix getting manipulated but nowhere is the version being set AFAICT. This tracks with the fact that the published package has a 1.0.0.0 version.

crates/devolutions-pedm-shell-ext/AppxManifest.xml
thenextman: Yes and no :) 

I'm not sure how the display name actually matters for the .msix since it's not something the user typically interacts with and I'm not sure where Windows actually shows you this (install MSIX / shell extensions)

However; if the shell extension launches the desktop agent (the tray icon application) and causes it to fire a UWP notification, the notification gets the name of the parent process. So you get a notification titled "Devolutions Agent Shell Extension" which is a bit of a weird experience.

I'm happy to change this back if there's a concern, though.

crates/devolutions-pedm/openapi/dotnet-client/config.json
```diff
@@ -2,7 +2,7 @@
   "packageAuthors": "Devolutions Inc.",
   "packageName": "Devolutions.Pedm.Client",
   "packageTitle": "Devolutions PEDM REST API Client",
-  "packageVersion": "2024.10.25",
+  "packageVersion": "2024.3.6.0",
```
CBenoit, line 5: For the Devolutions.Pedm.Client library, I suggest we keep following the date format that we follow for all the other libraries. 2024.3.6.0 (and so on) would only be for user-facing applications that we actually ship.

CBenoit:
Do I understand correctly that devolutions-pedm-shell-ext is a binary included into Devolutions.Agent.Desktop? Is the version we set to devolutions-pedm-shell-ext relevant?

crates/devolutions-pedm-shell-ext/AppxManifest.xml
awakecoding: my bad, I didn't recall we had a script for bumping the version numbers everywhere

crates/devolutions-pedm/openapi/dotnet-client/config.json
thenextman: Yes you are right, I missed this detail. Will correct that.

CBenoit: Thank you! Feel free to rebase-merge afterwards! :slightly_smiling_face:

CBenoit:
> > Do I understand correctly that devolutions-pedm-shell-ext is a binary included into Devolutions.Agent.Desktop? Is the version we set to devolutions-pedm-shell-ext relevant?
> 
> I thought you would pick up on this.
> 
> It's built as a separate package (devolutions-pedm-shell-ext) and deployed to %programfiles% along with the .msix manifest. The .msi executes the manifest, which registers the DLL with the shell to provide the shell extension for PEDM.
> 
> I would like the DLL to be versioned but since we're shipping "just a DLL" here rather than a rust crate, I thought we should follow the versioning expected there rather than the rust versioning scheme.
> 
> Does it make sense?

Thank you for the details! Sounds good to me!

crates/devolutions-pedm/openapi/dotnet-client/config.json
thenextman: @CBenoit Please check the last commit. I updated the generate-clients script to automatically bump the version to today's date.

CBenoit: Good idea! Do you think you could update all the config.json files instead of only the first?

thenextman: You mean, also the `devolutions-pedm-client-http` config?

Is it what we want? As far as I can tell, the version is useless here because it's only consumed as a reference by it's parent crate...

CBenoit: My bad, you are right, all the other packages are following a different convention.

PR #1161, opened 2024-12-18
fix(pedm): change policy to be keyed by Uuid

thenextman:
Apologies for the noisy PR due to regenerating the OpenAPI data. The relevant changes are in

Cargo.lock
crates/devolutions-pedm-shared/Cargo.toml
crates/devolutions-pedm-shared/src/policy.rs

The PEDM policy was using a `String` for the policy ID; enforced as an alphanumeric value not greater than 32 characters. The ID is referenced in the .json document with the serialized policy, as well as used for the .json filename.

This places an unnecessary toll on the user and UX; where they must manually specify an ID alongside the policy name for little gain.

This PR changes the policy ID to be a `Uuid` / `GUID`.


crates/devolutions-pedm/openapi/pedm-api.yaml
```diff
@@ -501,7 +502,7 @@ components:
               MaximumSeconds: 0
           $ref: '#/components/schemas/ElevationConfigurations'
         Id:
-          default: default
+          default: 65811918-5aa5-4eec-ae77-a2843f3eccdc
```
CBenoit, line 505: question: Why this arbitrary default value?

crates/devolutions-pedm-shared/src/policy.rs
```diff
@@ -360,7 +356,7 @@ impl Identifiable for Profile {
 impl Default for Profile {
     fn default() -> Self {
         Self {
-            id: Id("default".to_owned()),
+            id: Id(Uuid::new_v4()),
```
CBenoit, line 359: Is it caused by this implementation of the Default trait? Does that mean we’ll change the default value in the OpenAPI document at each generation? I’m thinking that the OpenAPI document should not include a default value.

thenextman: Yes, you're right, and I hadn't considered why this was here

We're forced to have a `Default` here; I'll see if I can remove it since it's really not needed in the documentation. This API is expressly private.

Otherwise, I'll change it to an empty GUID.

thenextman: I've changed it to `uuid::default()` so we get an empty GUID in the doc examples.

It's OpenAPI that forces a `Default` implementation here; perhaps we can disable this but for now I think this is good.

Next time I'll separate commits for the actual changes versus API generation, sorry about that

PR #1163, opened 2024-12-19
refactor(player): player project, introduce typescript and vite

irvingoujAtDevolution:
Now the webapp/player folder will be the build output directory for player-project.
I am currently having a asciicast terminal streaming working in local branch, that is based on this Pull Request. I wanted to submit them all in one but that would be too much change at once.


CBenoit:
Thanks for separating the refactor PR from the feature PR.

CBenoit (approved):
I think this is very nice, but do you think we could avoid having two folders?
I feel "player-project" should really be a package named "player".

Going further, I think it would be good to change the workspace in depth:
- A `frontend` folder which is a npm workspace.
- Inside this folder we find:
  - Tooling configuration files (typescript, biome, etc)
  - The `webapp` package (the standalone frontend)
  - The `player` package
  - The `shadow-player` package
- [pnmp](https://pnpm.io/) is good for managing npm packages workspace.
- Building the frontend would mean recreating the `webapp` folder arborescence using dist artifacts, and it would itself be an artifact.

As this is a net improvement as-is, I’ll approve the PR so you can merge now if you prefer. Let me know what you think about the suggested refactor, and maybe we can go ahead later.

PR #1164, opened 2024-12-20
fix(pedm): change policy to be keyed by Uuid

thenextman:
#### The "I'm fed up with OpenAPI generators PR"

This PR recreates my change from #1161:

The PEDM policy was using a String for the policy ID; enforced as an alphanumeric value not greater than 32 characters. The ID is referenced in the .json document with the serialized policy, as well as used for the .json filename.

This places an unnecessary toll on the user and UX; where they must manually specify an ID alongside the policy name for little gain.

This PR changes the policy ID to be a Uuid / GUID.

#### Notes

- View the individual commits, since I regenerated the OpenAPI clients
- The documentation examples for types using `uuid` contain the magic number "046b6c7f-0b8a-43b9-b35d-6489e6daee91". This is hard-coded into the OpenAPI [examples](https://github.com/OpenAPITools/openapi-generator/blob/b143313cf7d7f9ac27723db637bb213890fca738/modules/openapi-generator/src/main/java/org/openapitools/codegen/examples/ExampleGenerator.java#L296), and (at least, with the csharp generator) there is no way to disable example generation.
- The RestSharp and HttpClient OpenAPI backends [do not support](https://github.com/OpenAPITools/openapi-generator/issues/18017) generating certain value types as `nullable`, including `Guid`. This is relevant in the  `GetProfilesMeResponse` where, if the user doesn't have a policy assigned, we don't have a value here. I don't have the capacity to fix this in OpenAPI, so for now, we are careful to always return a value here (an empty GUID in the default case).
- I tweaked the generation script to call `rustfmt` on the generated rust source files. Again, I don't have the bandwidth to fix the generator; and this is an easy solution to prevent every source file being changed by the generator (and needing to remember to format them before pushing).
- Add .md and .mustache files to the .gitattributes. On Windows, with the right git configuration; these files were converted to CRLF when pulled and generated a lot of noise in the local git status (especially for C# files, where the CRLF header from the template would be mixed with the LF from the generated code file 🤮)


crates/devolutions-pedm-shared/src/policy.rs
```diff
@@ -309,23 +310,26 @@ pub struct ElevationConfigurations {
 #[repr(transparent)]
 #[derive(Serialize, Deserialize, JsonSchema, Debug, PartialEq, Eq, Hash, Clone)]
 #[serde(try_from = "String")]
-pub struct Id(String);
+pub struct Id(Uuid);
```
CBenoit, line 313: suggestion: I think it’s fine to just use `Uuid` directly, and remove the `Id` type which was just a wrapper for validating the string-based ID using a Regex.

crates/devolutions-pedm/openapi/generate_clients.ps1
```diff
@@ -71,12 +71,12 @@ try {
...
 
 		$FilesPath = Join-Path $(Convert-Path -LiteralPath $target.Folder) ".openapi-generator" "FILES"
-		$PathsToFormat = Get-Content -Path $FilesPath | Where-Object { $_ -Like "*.rs" } | ForEach-Object { Join-Path $(Convert-Path -LiteralPath $target.Folder) $_ }
+		$PathsToFormat = Get-Content -Path $FilesPath | ForEach-Object { Join-Path $(Convert-Path -LiteralPath $target.Folder) $_ }
 
 		Push-Location $RootDirectory
 
 		try {
-			foreach ($PathToFormat in $PathsToFormat)
+			foreach ($PathToFormat in $PathsToFormat | Where-Object { $_ -Like "*.rs" })
```
CBenoit: nitpick: I think it makes sense to filter first, especially since the variable is named "PathsToFormat".

CBenoit (approved):
If it does not make sense to have a default value for "Profile", I think you can remove that like this:

```diff
diff --git a/crates/devolutions-pedm-shared/src/policy.rs b/crates/devolutions-pedm-shared/src/policy.rs
index defb58de..1cacb5e5 100644
--- a/crates/devolutions-pedm-shared/src/policy.rs
+++ b/crates/devolutions-pedm-shared/src/policy.rs
@@ -351,7 +351,6 @@ pub trait Identifiable {

 #[derive(Serialize, Deserialize, JsonSchema, Debug, Clone, Hash, PartialEq, Eq)]
 #[serde(rename_all = "PascalCase")]
-#[serde(default)]
 pub struct Profile {
     pub id: Id,
     pub name: String,
@@ -367,19 +366,6 @@ impl Identifiable for Profile {
     }
 }

-impl Default for Profile {
-    fn default() -> Self {
-        Self {
-            id: Id(Uuid::default()),
-            name: "Unnamed profile".to_owned(),
-            elevation_method: ElevationMethod::LocalAdmin,
-            elevation_settings: ElevationConfigurations::default(),
-            default_elevation_kind: ElevationKind::Deny,
-            prompt_secure_desktop: true,
-        }
-    }
-}
-
 #[derive(Serialize, Deserialize, JsonSchema, Debug, Default)]
 #[serde(rename_all = "PascalCase")]
 pub struct Configuration {
```

Approving so you can merge when you see fit.

CBenoit:
By the way,

> The "I'm fed up with OpenAPI generators PR"

I’m sorry you had a hard time with that, thank you for looking it it :joy:

crates/devolutions-pedm/openapi/generate_clients.ps1
thenextman: Quite right. This was a holdover from when I was post-processing other file types, but changed my approach. Fixed.

thenextman:
> If it does not make sense to have a default value for "Profile", I think you can remove that like this:
> 
> ```diff
> diff --git a/crates/devolutions-pedm-shared/src/policy.rs b/crates/devolutions-pedm-shared/src/policy.rs
> index defb58de..1cacb5e5 100644
> --- a/crates/devolutions-pedm-shared/src/policy.rs
> +++ b/crates/devolutions-pedm-shared/src/policy.rs
> @@ -351,7 +351,6 @@ pub trait Identifiable {
> 
>  #[derive(Serialize, Deserialize, JsonSchema, Debug, Clone, Hash, PartialEq, Eq)]
>  #[serde(rename_all = "PascalCase")]
> -#[serde(default)]
>  pub struct Profile {
>      pub id: Id,
>      pub name: String,
> @@ -367,19 +366,6 @@ impl Identifiable for Profile {
>      }
>  }
> 
> -impl Default for Profile {
> -    fn default() -> Self {
> -        Self {
> -            id: Id(Uuid::default()),
> -            name: "Unnamed profile".to_owned(),
> -            elevation_method: ElevationMethod::LocalAdmin,
> -            elevation_settings: ElevationConfigurations::default(),
> -            default_elevation_kind: ElevationKind::Deny,
> -            prompt_secure_desktop: true,
> -        }
> -    }
> -}
> -
>  #[derive(Serialize, Deserialize, JsonSchema, Debug, Default)]
>  #[serde(rename_all = "PascalCase")]
>  pub struct Configuration {
> ```
> 
> Approving so you can merge when you see fit.

Excellent, this is much better. Thanks!

crates/devolutions-pedm/src/api/policy.rs
```diff
@@ -110,7 +110,11 @@ async fn get_me(
     let policy = policy::policy().read();
 
     Ok(Json(GetProfilesMeResponse {
-        active: policy.user_current_profile(&named_pipe_info.user).map(|p| p.id.clone()),
+        // OpenAPI won't generate a guid? for the `active` field, so be sure to return a value
+        active: policy
+            .user_current_profile(&named_pipe_info.user)
+            .map(|p| p.id.clone())
+            .or_else(|| Some(Id::default())),
```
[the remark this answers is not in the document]
thenextman: Excellent. This is a case of me wresting for so long with OpenAPI that I then couldn't "see the wood for the trees"

crates/devolutions-pedm-shared/src/policy.rs
thenextman: I've done this too; I thought of it earlier but didn't bother. But actually: fearless refactoring FTW!

thenextman:
> By the way,
> 
> > The "I'm fed up with OpenAPI generators PR"
> 
> I’m sorry you had a hard time with that, thank you for looking it it 😂

Quite. In the end it was worthwhile because I now have a much better understanding of how this piece works. But very frustrating to wrestle with it for a long time to little gain. It's very much a black box.

Microsoft has added OpenAPI document generation directly to .NET 9: https://devblogs.microsoft.com/dotnet/dotnet9-openapi/. So maybe we'll start to see some better options in this space.

CBenoit:
> > By the way,
> > > The "I'm fed up with OpenAPI generators PR"
> > 
> > 
> > I’m sorry you had a hard time with that, thank you for looking it it 😂
> 
> Quite. In the end it was worthwhile because I now have a much better understanding of how this piece works. But very frustrating to wrestle with it for a long time to little gain. It's very much a black box.
> 
> Microsoft has added OpenAPI document generation directly to .NET 9: https://devblogs.microsoft.com/dotnet/dotnet9-openapi/. So maybe we'll start to see some better options in this space.

Oh, I would _love_ an official .NET client generator from Microsoft…!!
Kiota seems to be that: https://learn.microsoft.com/ja-jp/openapi/kiota/quickstarts/dotnet
This does not produce Rust code yet, but maybe the current generator for Rust is good enough already?

I have to admit I’ve observed the idea of manually rewriting the API clients many times. With `openapi-generator`, the generated code is very far from optimal, and
- It’s frustrating to deal with bugs in the generated code,
- Modifying the templates is painful,
- Keeping the modified templates up to date is painful,
- Configuring it is somewhat time consuming in the first place.

At least, the code is well in-sync, which is still a very nice property. If only it was cheaper to achieve :joy:
OpenAPI itself is a powerful standard, and I don’t think it’s a complete mistake either to use it either.

thenextman:
> > > By the way,
> > > > The "I'm fed up with OpenAPI generators PR"
> > > 
> > > 
> > > I’m sorry you had a hard time with that, thank you for looking it it 😂
> > 
> > 
> > Quite. In the end it was worthwhile because I now have a much better understanding of how this piece works. But very frustrating to wrestle with it for a long time to little gain. It's very much a black box.
> > Microsoft has added OpenAPI document generation directly to .NET 9: https://devblogs.microsoft.com/dotnet/dotnet9-openapi/. So maybe we'll start to see some better options in this space.
> 
> Oh, I would _love_ an official .NET client generator from Microsoft…!! Kiota seems to be that: https://learn.microsoft.com/ja-jp/openapi/kiota/quickstarts/dotnet This does not produce Rust code yet, but maybe the current generator for Rust is good enough already?
> 
> I have to admit I’ve observed the idea of manually rewriting the API clients many times. With `openapi-generator`, the generated code is very far from optimal, and
> 
> * It’s frustrating to deal with bugs in the generated code,
> * Modifying the templates is painful,
> * Keeping the modified templates up to date is painful,
> * Configuring it is somewhat time consuming in the first place.
> 
> At least, the code is well in-sync, which is still a very nice property. If only it was cheaper to achieve 😂 OpenAPI itself is a powerful standard, and I don’t think it’s a complete mistake either to use it either.

I basically agree on all points. The rust generator with OpenAPI is certainly good enough. For .NET, right now things are pretty far from ideal but it's also "the best we've got"; the alternatives are not more palatable. At some point it would be good to investigate the new MS generator although it's not clear to me right now if it's general purpose, or closely tied to ASP.NET.

PR #1167, opened 2024-12-21
fix(dgw):add error message when xmf is not loaded for session shadowing

irvingoujAtDevolution:


devolutions-gateway/src/api/jrec.rs
```diff
@@ -503,6 +504,10 @@ async fn shadow_recording(
         return Err(HttpError::not_found().msg("no active recording found for the specified ID"));
     }
 
+    if !xmf::is_init() {
+        return Err(HttpError::internal().msg("XMF dynamic library is not initialized"));
```
CBenoit, line 508:
```suggestion
        return Err(HttpError::internal().msg("XMF native library is not loaded"));
```

devolutions-gateway/src/service.rs
```diff
@@ -97,7 +97,7 @@ impl GatewayService {
                 Err(error) => warn!(
                     %path,
                     %error,
-                    "Failed to load XMF native library, video processing features are disabled"
+                    "Failed to load XMF native library, video processing (recording remuxing,video streaming) features are disabled"
```
CBenoit, line 100:
```suggestion
                    "Failed to load XMF native library, video processing features such as remuxing and shadowing are disabled"
```

CBenoit (approved):

PR #1172, opened 2024-12-24
fix(agent): add logic to start Gateway service after update

pacmancoder:
This PR adds the following logic to the updater:
* Gateway service startup mode when correctly performing MSI install
* Start service if it was in "Manual" startup mode before installation, but in "Running" state, to avoid unintuitive behavior

### DIFF
--- a/crates/win-api-wrappers/src/service.rs
+++ b/crates/win-api-wrappers/src/service.rs
@@ -1,3 +1,5 @@
+use std::alloc::Layout;
+
 use windows::core::Owned;
 use windows::Win32::Foundation::{ERROR_INSUFFICIENT_BUFFER, GENERIC_READ};
 use windows::Win32::System::Services::{
@@ -7,49 +9,52 @@ use windows::Win32::System::Services::{
     SERVICE_SYSTEM_START,
 };
 
+use crate::raw_buffer::RawBuffer;
 use crate::utils::WideString;
-use crate::Error;
 
 pub struct ServiceManager {
     handle: Owned<SC_HANDLE>,
 }
 
 impl ServiceManager {
-    pub fn open_read() -> Result<Self, Error> {
+    pub fn open_read() -> anyhow::Result<Self> {
         Self::open_with_access(GENERIC_READ.0)
     }
 
-    pub fn open_all_access() -> Result<Self, Error> {
+    pub fn open_all_access() -> anyhow::Result<Self> {
         Self::open_with_access(SC_MANAGER_ALL_ACCESS)
     }
 
-    fn open_with_access(access: u32) -> Result<Self, Error> {
-        // SAFETY: No preconditions.
+    fn open_with_access(access: u32) -> anyhow::Result<Self> {
+        // SAFETY: FFI call with no outstanding preconditions.
         let raw_sc_handle = unsafe { OpenSCManagerW(None, None, access)? };
 
-        // SAFETY: No preconditions.
+        // SAFETY: Handle returned by `OpenSCManagerW` is valid and needs to be closed after use,
+        // thus it is safe to take ownership of it via `Owned`.
         let handle = unsafe { Owned::new(raw_sc_handle) };
 
         Ok(Self { handle })
     }
 
-    fn open_service_with_access(&self, service_name: &str, access: u32) -> Result<Service, Error> {
+    fn open_service_with_access(&self, service_name: &str, access: u32) -> anyhow::Result<Service> {
         let service_name_wide = WideString::from(service_name);
 
-        // SAFETY: No preconditions.
+        // SAFETY: Value passed as hSCManager is valid as long as `ServiceManager` instance is
+        // alive. Service name is valid null-terminated UTF-16 string allocated on the heap.
         let raw_service_handle = unsafe { OpenServiceW(*self.handle, service_name_wide.as_pcwstr(), access)? };
 
-        // SAFETY: No preconditions.
+        // SAFETY: Handle returned by `OpenServiceW` is valid and needs to be closed after use,
+        // thus it is safe to take ownership of it via `Owned`.
         let handle = unsafe { Owned::new(raw_service_handle) };
 
         Ok(Service { handle })
     }
 
-    pub fn open_service_read(&self, service_name: &str) -> Result<Service, Error> {
+    pub fn open_service_read(&self, service_name: &str) -> anyhow::Result<Service> {
         self.open_service_with_access(service_name, SERVICE_QUERY_CONFIG | SERVICE_QUERY_STATUS)
     }
 
-    pub fn open_service_all_access(&self, service_name: &str) -> Result<Service, Error> {
+    pub fn open_service_all_access(&self, service_name: &str) -> anyhow::Result<Service> {
         self.open_service_with_access(service_name, SERVICE_ALL_ACCESS)
     }
 }
@@ -68,7 +73,7 @@ pub struct Service {
 }
 
 impl Service {
-    pub fn startup_mode(&self) -> Result<ServiceStartupMode, Error> {
+    pub fn startup_mode(&self) -> anyhow::Result<ServiceStartupMode> {
         let mut cbbufsize = 0u32;
         let mut pcbbytesneeded = 0u32;
 
@@ -83,38 +88,38 @@ impl Service {
             Ok(_) => panic!("QueryServiceConfigW should fail with ERROR_INSUFFICIENT_BUFFER"),
         }
 
-        // Event though `QueryServiceConfigW` states that `lpserviceconfig` is a `buffer`,
-        // it needs to be aligned ir order to avoid undefined behavior when accessing
-        // `QUERY_SERVICE_CONFIGW.
+        // The most typical buffer we work with in Rust are homogeneous arrays of integers such
+        // as [u8] or Vec<u8>, but in Microsoft’s Win32 documentation, a `buffer` generally refers
+        // to a caller-allocated memory region that an API uses to either input or output data, and
+        // it is ultimately coerced into some other type with various alignment requirements.
         //
-        // Clippy warning:
-        // casting from `*mut u8` to a more-strictly-aligned pointer
-        // (`*mut windows::Win32::System::Services::QUERY_SERVICE_CONFIGW`) (1 < 8 bytes)
-        assert_eq!(align_of::<QUERY_SERVICE_CONFIGW>(), size_of::<u64>());
+        // lpServiceConfig should point to aligned buffer that could hold a QUERY_SERVICE_CONFIGW
+        // structure.
+        let layout = Layout::from_size_align(
+            usize::try_from(pcbbytesneeded).expect("pcbbytesneeded < 8K as per MSDN"),
+            align_of::<QUERY_SERVICE_CONFIGW>(),
+        )?;
 
-        let pcbbytesneeded_usize = usize::try_from(pcbbytesneeded).expect("pcbbytesneeded < 8K as per MSDN");
+        // SAFETY: The layout initialization is checked using the Layout::from_size_align method.
+        let mut buffer = unsafe { RawBuffer::alloc_zeroed(layout)? };
 
-        let aligned_u64_array_size =
-            pcbbytesneeded_usize / size_of::<u64>() + usize::from(pcbbytesneeded_usize % size_of::<u64>() != 0);
-
-        let aligned_size = u32::try_from(aligned_u64_array_size * size_of::<u64>()).expect("pcbbytesneeded <= 8K");
-
-        let mut buffer: Box<Vec<u64>> = Box::new(vec![0; aligned_u64_array_size]);
-
-        // SAFETY: Buffuer with correct size was allocated prior to the call.
+        // SAFETY: Buffer passed to `lpServiceConfig` have enough size to hold a
+        // QUERY_SERVICE_CONFIGW structure, as required size was queried and allocated above.
+        // Passed buffer have correct alignment to hold QUERY_SERVICE_CONFIGW structure.
         unsafe {
+            // NOTE: Pointer cast is valid, as `buffer` is allocated with correct alignment above.
+            #[allow(clippy::cast_ptr_alignment)]
             QueryServiceConfigW(
                 *self.handle,
-                Some(buffer.as_mut_ptr() as *mut QUERY_SERVICE_CONFIGW),
-                aligned_size,
+                Some(buffer.as_mut_ptr().cast::<QUERY_SERVICE_CONFIGW>()),
+                pcbbytesneeded,
                 &mut cbbufsize,
             )?
         };
 
-        let ptr = buffer.as_mut_ptr() as *mut QUERY_SERVICE_CONFIGW;
         // SAFETY: `QueryServiceConfigW` succeeded, thus `lpserviceconfig` is valid and contains
         // QUERY_SERVICE_CONFIGW structure.
-        let config = unsafe { &*ptr };
+        let config = unsafe { buffer.as_ref_cast::<QUERY_SERVICE_CONFIGW>() };
 
         match config.dwStartType {
             SERVICE_BOOT_START => Ok(ServiceStartupMode::Boot),
@@ -126,15 +131,18 @@ impl Service {
         }
     }
 
-    pub fn is_running(&self) -> Result<bool, Error> {
+    pub fn is_running(&self) -> anyhow::Result<bool> {
         let mut service_status = SERVICE_STATUS::default();
+
+        // SAFETY: hService is valid handle, lpServiceStatus is valid pointer to stack-allocated
+        // SERVICE_STATUS structure.
         unsafe { QueryServiceStatus(*self.handle, &mut service_status as *mut _)? };
 
         Ok(service_status.dwCurrentState == SERVICE_RUNNING)
     }
 
-    pub fn start(&self) -> Result<(), Error> {
-        // SAFETY: No preconditions, arguments are not used.
+    pub fn start(&self) -> anyhow::Result<()> {
+        // SAFETY: FFI call with no outstanding preconditions.
         unsafe { StartServiceW(*self.handle, None)? };
 
         Ok(())
--- a/crates/win-api-wrappers/src/service.rs
+++ b/crates/win-api-wrappers/src/service.rs
@@ -1,5 +1,6 @@
 use std::alloc::Layout;
 
+use thiserror::Error;
 use windows::core::Owned;
 use windows::Win32::Foundation::{ERROR_INSUFFICIENT_BUFFER, GENERIC_READ};
 use windows::Win32::System::Services::{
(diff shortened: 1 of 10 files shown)


crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    pub fn is_running(&self) -> Result<bool, Error> {
+        let mut service_status = SERVICE_STATUS::default();
+        unsafe { QueryServiceStatus(*self.handle, &mut service_status as *mut _)? };
+
+        Ok(service_status.dwCurrentState == SERVICE_RUNNING)
+    }
+
+    pub fn start(&self) -> Result<(), Error> {
+        // SAFETY: No preconditions, arguments are not used.
```
CBenoit, line 137: question: Arguments are not used? Even `self.handle`?

suggestion: The safety comment I typically recommend when it’s a straightforward usual business stuff FFI call is:

```rust
// SAFETY: FFI call with no outstanding preconditions.
```

Pointers such as `self.handle` being valid is part of the usual business.

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,142 @@
...
+            SERVICE_AUTO_START => Ok(ServiceStartupMode::Automatic),
+            SERVICE_DEMAND_START => Ok(ServiceStartupMode::Manual),
+            SERVICE_DISABLED => Ok(ServiceStartupMode::Disabled),
+            _ => panic!("WinAPI returned invalid service startup mode"),
+        }
+    }
+
+    pub fn is_running(&self) -> Result<bool, Error> {
+        let mut service_status = SERVICE_STATUS::default();
+        unsafe { QueryServiceStatus(*self.handle, &mut service_status as *mut _)? };
```
CBenoit, line 131: issue: Missing safety comment (you probably have a warning for that)

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,142 @@
...
+    SC_HANDLE, SC_MANAGER_ALL_ACCESS, SERVICE_ALL_ACCESS, SERVICE_AUTO_START, SERVICE_BOOT_START, SERVICE_DEMAND_START,
+    SERVICE_DISABLED, SERVICE_QUERY_CONFIG, SERVICE_QUERY_STATUS, SERVICE_RUNNING, SERVICE_STATUS,
+    SERVICE_SYSTEM_START,
+};
+
+use crate::utils::WideString;
+use crate::Error;
+
+pub struct ServiceManager {
+    handle: Owned<SC_HANDLE>,
```
CBenoit, line 14: praise: Nice, I didn’t notice about these wrappers until now. We should definitely use `Ref` and `Owned` more!

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,142 @@
...
+        Self::open_with_access(GENERIC_READ.0)
+    }
+
+    pub fn open_all_access() -> Result<Self, Error> {
+        Self::open_with_access(SC_MANAGER_ALL_ACCESS)
+    }
+
+    fn open_with_access(access: u32) -> Result<Self, Error> {
+        // SAFETY: No preconditions.
+        let raw_sc_handle = unsafe { OpenSCManagerW(None, None, access)? };
```
CBenoit, line 28: nitpick: Most of the time "No preconditions" is not true. The following may be used to declare that this is just a FFI call with the usual FFI business preconditions (valid pointers, etc).

```rust
// SAFETY: FFI call with no outstanding preconditions.
```

**If the function really have no precondition at all, it should be possible to wrap it as-is as a safe function.**

It’s not possible for OpenSCManagerW where the two first parameters are pointers. As per the documentation, it’s safe to call the function with NULL (and that’s why the windows crate is using `Option`) for that. It’s obvious enough for using the above safety comment.

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        // SAFETY: No preconditions.
+        let handle = unsafe { Owned::new(raw_sc_handle) };
```
CBenoit, line 31: issue: There is a very important precondition. "The handle must be owned by the caller and safe to free.". The safety doc should explain how it is upheld.

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,142 @@
...
+        let pcbbytesneeded_usize = usize::try_from(pcbbytesneeded).expect("pcbbytesneeded < 8K as per MSDN");
+
+        let aligned_u64_array_size =
+            pcbbytesneeded_usize / size_of::<u64>() + usize::from(pcbbytesneeded_usize % size_of::<u64>() != 0);
+
+        let aligned_size = u32::try_from(aligned_u64_array_size * size_of::<u64>()).expect("pcbbytesneeded <= 8K");
+
+        let mut buffer: Box<Vec<u64>> = Box::new(vec![0; aligned_u64_array_size]);
+
+        // SAFETY: Buffuer with correct size was allocated prior to the call.
```
CBenoit, line 104: typo:
```suggestion
        // SAFETY: Buffer with correct size was allocated prior to the call.
```

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        match result {
+            Err(err) if err.code() == ERROR_INSUFFICIENT_BUFFER.to_hresult() => {
+                // Expected error, continue.
+            }
+            Err(err) => return Err(err.into()),
+            Ok(_) => panic!("QueryServiceConfigW should fail with ERROR_INSUFFICIENT_BUFFER"),
+        }
+
+        // Event though `QueryServiceConfigW` states that `lpserviceconfig` is a `buffer`,
```
CBenoit, line 86: note: FYI, the terminology "buffer" used in Windows doc is a bit different to what you may be expecting. The most typical buffer we work with in Rust are arrays of bytes such as `[u8]`, `Vec<u8>`. In Microsoft’s Win32 documentation, a “buffer” generally refers to a caller-allocated memory region that an API uses to either input or output data, and it is ultimately coerced into some other type with various alignment requirements. Feel free to reuse this verbatim in your comment if you find this notion useful.

crates/win-api-wrappers/src/service.rs
CBenoit, line 104: nitpick: Can you elaborate a bit more? As far as I understand, we should mention that:
- How the alignment is correct (code above). May be considered usual business in the case of Win32 API.
- What does the correct size refer to?

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,142 @@
...
+        // it needs to be aligned ir order to avoid undefined behavior when accessing
+        // `QUERY_SERVICE_CONFIGW.
+        //
+        // Clippy warning:
+        // casting from `*mut u8` to a more-strictly-aligned pointer
+        // (`*mut windows::Win32::System::Services::QUERY_SERVICE_CONFIGW`) (1 < 8 bytes)
+        assert_eq!(align_of::<QUERY_SERVICE_CONFIGW>(), size_of::<u64>());
+
+        let pcbbytesneeded_usize = usize::try_from(pcbbytesneeded).expect("pcbbytesneeded < 8K as per MSDN");
```
CBenoit, line 95: style: In Rust, it’s better style to shadow related variables that should not be used anymore.

devolutions-agent/src/updater/product_actions.rs
```diff
@@ -0,0 +1,88 @@
+use win_api_wrappers::service::{ServiceManager, ServiceStartupMode};
+
+use crate::updater::{Product, UpdaterError};
+
+const SERVICE_NAME: &str = "DevolutionsGateway";
```
CBenoit, line 5: thought: I am under the impression that this const is already defined elsewhere. Could we reuse it?

devolutions-agent/src/updater/product_actions.rs
CBenoit, line 1: praise: Good architecture.

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        // SAFETY: Buffuer with correct size was allocated prior to the call.
+        unsafe {
+            QueryServiceConfigW(
+                *self.handle,
+                Some(buffer.as_mut_ptr() as *mut QUERY_SERVICE_CONFIGW),
+                aligned_size,
+                &mut cbbufsize,
+            )?
+        };
```
CBenoit, line 112: suggestion: What about using `RawBuffer`?

https://github.com/Devolutions/devolutions-gateway/blob/ebc6ba46ba6009193dd0987e7281311398e5388f/crates/win-api-wrappers/src/raw_buffer.rs#L27

I feel your code is kind of re-implementing a similar logic in an indirect way using `Vec` and `u64`.

See an example usage here:

https://github.com/Devolutions/devolutions-gateway/blob/ebc6ba46ba6009193dd0987e7281311398e5388f/crates/win-api-wrappers/src/token.rs#L232-L276

BUT!

If the struct is not variable-sized (DST), the best is to initialize it on the stack using `let mut config = QUERY_SERVICE_CONFIGW::default()` and pass it as a mut reference `&mut config`. If necessary, convert into a `*mut c_void` pointer (`&mut config as *mut c_void`).

CBenoit (commented):
Good work!

General request: Review the safety comments according to my suggestions, even if I didn’t comment directly.

crates/win-api-wrappers/src/service.rs
pacmancoder: Ah, bad wording, sorry, I meant to say that service params (pcwstr vec) is not used here

crates/win-api-wrappers/src/service.rs
pacmancoder: Huh, strange, I missed this lint, my bad, thanks for spotting!

devolutions-agent/src/updater/product_actions.rs
pacmancoder: Hm, I didn't find it anywhere in the rust code, only in C# MSI installer code (service creation/removal logic) 🤔

CBenoit: I was thinking of this:

https://github.com/Devolutions/devolutions-gateway/blob/2e3aa7dd459b538eaaeb6bd3f1260c71e632f441/devolutions-agent/src/service.rs#L19-L21

But it appears to be slightly different… Is it normal? :thinking:

crates/win-api-wrappers/src/service.rs
pacmancoder: Indeed! `RawBuffer` is what I need, thanks!

devolutions-agent/src/updater/product_actions.rs
pacmancoder: Ah.... I believe this const is used to create/start service via ceviche controller, strange that it have a different value.

The real service name which is created by MSI is `DevolutionsGateway`

crates/win-api-wrappers/src/service.rs
pacmancoder: NOTE: I still needed to silence Clippy (`windows` crate still takes buffer via `*mut QUERY_SERVICE_CONFIGW` and  *mut u8 -> *mut QUERY_SERVICE_CONFIGW` cast still should occur), but now thanks to RawBuffer it is guaranteed to be aligned.

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,150 @@
...
+        let raw_sc_handle = unsafe { OpenSCManagerW(None, None, access)? };
+
+        // SAFETY: Handle returned by `OpenSCManagerW` is valid and needs to be closed after use,
+        // thus it is safe to take ownership of it via `Owned`.
+        let handle = unsafe { Owned::new(raw_sc_handle) };
+
+        Ok(Self { handle })
+    }
+
+    fn open_service_with_access(&self, service_name: &str, access: u32) -> anyhow::Result<Service> {
```
pacmancoder, line 39: Thought: `Error` type in this library should be refactored and used consistently across all modules, falling back to `anyhow` for now (I need to throw AllocError from RawBuffer somehow, mapping it to some win32 error code feels wrong).

pacmancoder:
@CBenoit I am ready for review round 2 😃

crates/win-api-wrappers/src/service.rs
CBenoit: How are you thinking we should go about this? I don’t think it’s necessarily better to have a single Error type which can be anything, but there is no easy way to create lazy enums like in TypeScript to combine multiple errors easily either :/

We may want to provide a top-level Error type for complex errors:

```rust
enum Error {
    Alloc(AllocError),
    System(windows::core::Error),
    Other(anyhow::Error),
    …
}
```

Is it something like that you were thinking about?

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,150 @@
...
+        Self::open_with_access(SC_MANAGER_ALL_ACCESS)
+    }
+
+    fn open_with_access(access: u32) -> anyhow::Result<Self> {
+        // SAFETY: FFI call with no outstanding preconditions.
+        let raw_sc_handle = unsafe { OpenSCManagerW(None, None, access)? };
+
+        // SAFETY: Handle returned by `OpenSCManagerW` is valid and needs to be closed after use,
+        // thus it is safe to take ownership of it via `Owned`.
+        let handle = unsafe { Owned::new(raw_sc_handle) };
```
CBenoit, line 34: nitpick: "… thus it is safe to do …" is generally not useful and typically could be removed. A SAFETY: comment, by definition, is here "**to explain why the unsafe block is safe**, and which invariants are used and must be respected". It’s important to elaborate enough, but let’s try to keep the wording minimal.

suggestion:
```suggestion
        // SAFETY: On success, the handle returned by `OpenSCManagerW` is valid and owned by the caller.
        let handle = unsafe { Owned::new(raw_sc_handle) };
```

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        Ok(Self { handle })
+    }
+
+    fn open_service_with_access(&self, service_name: &str, access: u32) -> anyhow::Result<Service> {
+        let service_name_wide = WideString::from(service_name);
+
+        // SAFETY: Value passed as hSCManager is valid as long as `ServiceManager` instance is
+        // alive. Service name is valid null-terminated UTF-16 string allocated on the heap.
+        let raw_service_handle = unsafe { OpenServiceW(*self.handle, service_name_wide.as_pcwstr(), access)? };
```
CBenoit, line 44: nitpick: In term on readability, I recommend a newline for each item, optionally formatted as a list.

nitpick: Use the name of the variable as-is.

suggestion: Two options:
```suggestion
        // SAFETY: 
        // - Value passed as hSCManager is valid as long as `ServiceManager` instance is alive.
        // - service_name_wide is a valid, null-terminated UTF-16 string allocated on the heap.
        let raw_service_handle = unsafe { OpenServiceW(*self.handle, service_name_wide.as_pcwstr(), access)? };
```
```suggestion
        // SAFETY: Value passed as hSCManager is valid as long as `ServiceManager` instance is alive. 
        // service_name_wide is a valid, null-terminated UTF-16 string allocated on the heap.
        let raw_service_handle = unsafe { OpenServiceW(*self.handle, service_name_wide.as_pcwstr(), access)? };
```

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,150 @@
...
+pub struct Service {
+    handle: Owned<SC_HANDLE>,
+}
+
+impl Service {
+    pub fn startup_mode(&self) -> anyhow::Result<ServiceStartupMode> {
+        let mut cbbufsize = 0u32;
+        let mut pcbbytesneeded = 0u32;
+
+        // SAFETY: No preconditions. Query required buffer size.
```
CBenoit, line 80: Same as https://github.com/Devolutions/devolutions-gateway/pull/1172#discussion_r1896855936

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        // SAFETY: Buffer passed to `lpServiceConfig` have enough size to hold a
+        // QUERY_SERVICE_CONFIGW structure, as required size was queried and allocated above.
+        // Passed buffer have correct alignment to hold QUERY_SERVICE_CONFIGW structure.
+        unsafe {
+            // NOTE: Pointer cast is valid, as `buffer` is allocated with correct alignment above.
+            #[allow(clippy::cast_ptr_alignment)]
+            QueryServiceConfigW(
+                *self.handle,
+                Some(buffer.as_mut_ptr().cast::<QUERY_SERVICE_CONFIGW>()),
```
CBenoit, line 114: suggestion: Put the comment to justify the suppression on the same line. This avoid confusion with other unrelated comments.

suggestion: Use `expect` instead of `allow`, in case the lint is not fired anymore some day

nitpick: In fact, the `NOTE:` prefix is rarely meaningful.

```suggestion
            #[expect(clippy::cast_ptr_alignment)] // Pointer cast is valid, as `buffer` is allocated with correct alignment above.
            QueryServiceConfigW(
                *self.handle,
                Some(buffer.as_mut_ptr().cast::<QUERY_SERVICE_CONFIGW>()),
```

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,150 @@
...
+                pcbbytesneeded,
+                &mut cbbufsize,
+            )?
+        };
+
+        // SAFETY: `QueryServiceConfigW` succeeded, thus `lpserviceconfig` is valid and contains
+        // QUERY_SERVICE_CONFIGW structure.
```
CBenoit, line 121: typo:

```suggestion
        // a QUERY_SERVICE_CONFIGW structure.
```

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,150 @@
...
+            SERVICE_DISABLED => Ok(ServiceStartupMode::Disabled),
+            _ => panic!("WinAPI returned invalid service startup mode"),
+        }
+    }
+
+    pub fn is_running(&self) -> anyhow::Result<bool> {
+        let mut service_status = SERVICE_STATUS::default();
+
+        // SAFETY: hService is valid handle, lpServiceStatus is valid pointer to stack-allocated
+        // SERVICE_STATUS structure.
```
CBenoit, line 138:
```suggestion
        // SAFETY: hService is a valid handle.
        // lpServiceStatus is a valid pointer to a stack-allocated SERVICE_STATUS structure.
```

devolutions-agent/src/updater/product_actions.rs
CBenoit: Maybe something to refactor here!

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,150 @@
...
+
+        // SAFETY: Handle returned by `OpenSCManagerW` is valid and needs to be closed after use,
+        // thus it is safe to take ownership of it via `Owned`.
+        let handle = unsafe { Owned::new(raw_sc_handle) };
+
+        Ok(Self { handle })
+    }
+
+    fn open_service_with_access(&self, service_name: &str, access: u32) -> anyhow::Result<Service> {
+        let service_name_wide = WideString::from(service_name);
```
CBenoit, line 40: Same comment: https://github.com/Devolutions/devolutions-gateway/pull/1172#discussion_r1896863198

crates/win-api-wrappers/src/service.rs
```diff
@@ -29,19 +29,20 @@ impl ServiceManager {
...
     fn open_service_with_access(&self, service_name: &str, access: u32) -> anyhow::Result<Service> {
-        let service_name_wide = WideString::from(service_name);
+        let service_name = WideString::from(service_name);
 
-        // SAFETY: Value passed as hSCManager is valid as long as `ServiceManager` instance is
-        // alive. Service name is valid null-terminated UTF-16 string allocated on the heap.
-        let raw_service_handle = unsafe { OpenServiceW(*self.handle, service_name_wide.as_pcwstr(), access)? };
+        // SAFETY:
+        // - Value passed as hSCManager is valid as long as `ServiceManager` instance is alive.
+        // - service_name_wide is a valid, null-terminated UTF-16 string allocated on the heap.
```
CBenoit, line 44: nitpick: Outdated comment (service_name_wide)

devolutions-agent/src/updater/product_actions.rs
pacmancoder: I'll move that const to `devolutions-gateway-shared`

crates/win-api-wrappers/src/service.rs
pacmancoder: The problem is that crate uses a mix of our own Error type (e.g in `acl.rs`) but in many places `anyhow` is used (e.g. token.rs). I think we should use a single Error type throughout all crate modules instead of this mix. IMHO `anyhow` should only be used in top-level crates like `devolutions-gateway`/`devolutions-agent` (exception - if third party library throws anyhow, wrap it in its enum variant. 

What is really needed in this crate is to eliminate the mixing of error types in library API, but this is definitely out of the scope of PR.

However, I don't know why I decided to switch to `anyhow` instead of adding new Error variant before going on vacation, thank you for bringing that up, looks like my brain was very blurry during those last-minute changes, I'll look into it again, most likely switch to Error back 😄

crates/win-api-wrappers/src/service.rs
```diff
@@ -0,0 +1,151 @@
...
+        let service_name = WideString::from(service_name);
+
+        // SAFETY:
+        // - Value passed as hSCManager is valid as long as `ServiceManager` instance is alive.
+        // - service_name_wide is a valid, null-terminated UTF-16 string allocated on the heap.
+        let raw_service_handle = unsafe { OpenServiceW(*self.handle, service_name.as_pcwstr(), access)? };
+
+        // SAFETY: Handle returned by `OpenServiceW` is valid and needs to be closed after use,
+        // thus it is safe to take ownership of it via `Owned`.
+        let handle = unsafe { Owned::new(raw_service_handle) };
```
CBenoit, line 49: Same as https://github.com/Devolutions/devolutions-gateway/pull/1172#discussion_r1899677253

CBenoit (approved):
I’m done reviewing. You can address the last nitpicks, and feel free to merge!

crates/win-api-wrappers/src/service.rs
CBenoit: Okay, yeah we could stop using anyhow though I would like some way to attach context to the error. Sometimes we just get an obscure system error in the `Error::System` variant with no idea of which call caused it.

Given that we typically don’t inspect programmatically the structured errors, I think surprisingly anyhow is kind of fine even in libraries, though I tend to agree with you; it’s cleaner to have a little bit more of structure for libraries.

I don’t think we should necessarily have a strict rule of "one crate = one Error type" though. The standard library is not following this rule, because errors returned by a given module are often structurally completely unrelated to the kind of errors returned in another module.

We do have that in `win-api-wrappers` as well: we don’t need to use a wider error in the `raw_buffer` module, because `AllocError` is all we need. Everything is cleanly encapsulated inside the module.

FYI: I think we should split the `win-api-wrappers` into two crates: one for the cross-platform primitives (the code that we use to interface with the Windows API, but that is still pure Rust), such as `RawBuffer`, `Win32Dst` and other helpers, and one for the code actually wrapping Windows API such as `token` module. I’m thinking about doing this refactor a little bit later this week. The code in the helper crate would not have a single top-level error because everything is only loosely related, but a top-level error in the API wrapper crate would be very useful.

We could imagine something similar to what we have in IronRDP: https://github.com/Devolutions/IronRDP/blob/25bbb2682c95419fc18a61cab81e29a0ab9f23d4/crates/ironrdp-error/src/lib.rs

This should be adjusted to our needs. For instance, maybe we only need an optional context string for system errors:

```rust
enum Error {
    Alloc(AllocError),
    System { source: windows::core::Error, context: Option<String> },
    Other(anyhow::Error),
    …
}
```

If you want to go ahead and design that, let me know your ideas in Slack!

CBenoit: Oh, by the way, this didn’t occur to me either when I initially reviewed, but you don’t necessarily need to propagate the AllocError errors. In the `dst` module, we make assume there is enough memory to allocate the structure:
https://github.com/Devolutions/devolutions-gateway/blob/0ed70d21bac2cbf31dcc9dbe5615a8e573b73717/crates/win-api-wrappers/src/dst.rs#L103
(pretty much like Vec is doing)

And we make some reasonable assumptions about the number of elements: https://github.com/Devolutions/devolutions-gateway/blob/0ed70d21bac2cbf31dcc9dbe5615a8e573b73717/crates/win-api-wrappers/src/dst.rs#L116

pacmancoder: Hmm, I thought about it for a bit and I think having one crate=one error type error is not optimal either...

I like the idea of per-module custom error types. That way the error types will have only minimal nesting (easier to match for specific errors if needed) while keeping the possibility to add more context on-demand and new error kinds without breaking the API (Significantly at least. I think we could make all custom errors `#[non_exhaustive]`  to enforce error matching with catch-all but I don't know if it worth it?).

For the scope of this PR, I'll add a new `ServiceError` type with a single `ServiceError::Win32` variant (I'll follow your suggestion and add `.expect` for OOM)

pacmancoder: To be honest, I think it makes sense only in the higher-level crates - e.g. I still use `Product` in `devolutions-agent/src/updater/error.rs` to distinguish errors for different products in updater module, but `win-api-wrapper` errors are usually wrapped in higher level errors and passed as `source`

PR #1175, opened 2025-01-06
build(dgw): move libxmf.so to /usr/lib

allan2:
This moves _libxmf.so_ from `/usr/share/devolutions-gateway` to
`/usr/lib/devolutions-gateway`.

According to the [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html), `/usr/share` is for architecture-independent data files.

### DIFF
--- a/ci/tlk.ps1
+++ b/ci/tlk.ps1
@@ -769,7 +769,7 @@ class TlkRecipe
             arch = $DebianArchitecture
             deps = $($Dependencies -Join ", ")
             email = $Email
-            package = $Packager
+            packager = $Packager
             website = $Website
         } -OutputFile $ControlFile
 
--- a/devolutions-gateway/src/config.rs
+++ b/devolutions-gateway/src/config.rs
@@ -315,27 +315,17 @@ impl Conf {
             return Some(path.to_owned());
         }
 
-        #[cfg(target_os = "windows")]
-        {
+        if cfg!(target_os = "windows") {
             let path = env::current_exe().ok()?.parent()?.join("xmf.dll");
             return Utf8PathBuf::from_path_buf(path).ok();
+        } else if cfg!(all(
+            target_os = "linux",
+            any(target_arch = "x86_64", target_arch = "aarch64")
+        )) {
+            Some(Utf8PathBuf::from("/usr/lib/devolutions-gateway/libxmf.so"))
+        } else {
+            None
         }
-
-        #[cfg(target_os = "linux")]
-        {
-            #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
-            {
-                return None;
-            }
-
-            let arch_path = format!(
-                "/usr/lib/{arch}-linux-gnu/devolutions-gateway/libxmf.so",
-                arch = env::consts::ARCH,
-            );
-            return Some(Utf8PathBuf::from(arch_path));
-        }
-
-        None
     }
 }
 
--- a/devolutions-gateway/src/config.rs
+++ b/devolutions-gateway/src/config.rs
@@ -317,7 +317,7 @@ impl Conf {
 
         if cfg!(target_os = "windows") {
             let path = env::current_exe().ok()?.parent()?.join("xmf.dll");
-            return Utf8PathBuf::from_path_buf(path).ok();
+            Utf8PathBuf::from_path_buf(path).ok()
         } else if cfg!(target_os = "linux") {
             Some(Utf8PathBuf::from("/usr/lib/devolutions-gateway/libxmf.so"))
         } else {
--- a/devolutions-gateway/src/config.rs
+++ b/devolutions-gateway/src/config.rs
@@ -309,20 +309,33 @@ impl Conf {
     pub fn get_lib_xmf_path(&self) -> Option<Utf8PathBuf> {
         if let Ok(path) = env::var(ENV_VAR_DGATEWAY_LIB_XMF_PATH) {
             return Some(Utf8PathBuf::from(path));
-        };
+        }
 
         if let Some(path) = self.debug.lib_xmf_path.as_deref() {
             return Some(path.to_owned());
         }
 
-        if cfg!(target_os = "windows") {
+        #[cfg(target_os = "windows")]
+        {
             let path = env::current_exe().ok()?.parent()?.join("xmf.dll");
-            Utf8PathBuf::from_path_buf(path).ok()
-        } else if cfg!(target_os = "linux") {
-            Some(Utf8PathBuf::from("/usr/share/devolutions-gateway/libxmf.so"))
-        } else {
-            None
+            return Utf8PathBuf::from_path_buf(path).ok();
         }
+
+        #[cfg(target_os = "linux")]
+        {
+            #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
+            {
+                return None;
+            }
+
+            let arch_path = format!(
+                "/usr/lib/{arch}-linux-gnu/devolutions-gateway/libxmf.so",
+                arch = env::consts::ARCH,
+            );
+            return Some(Utf8PathBuf::from(arch_path));
+        }
+
+        None
     }
 }
 
--- a/devolutions-gateway/src/config.rs
+++ b/devolutions-gateway/src/config.rs
@@ -318,10 +318,7 @@ impl Conf {
         if cfg!(target_os = "windows") {
             let path = env::current_exe().ok()?.parent()?.join("xmf.dll");
             return Utf8PathBuf::from_path_buf(path).ok();
-        } else if cfg!(all(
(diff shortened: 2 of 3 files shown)


ci/tlk.ps1
```diff
@@ -769,7 +769,7 @@ class TlkRecipe
             arch = $DebianArchitecture
             deps = $($Dependencies -Join ", ")
             email = $Email
-            package = $Packager
+            packager = $Packager
```
CBenoit, line 772: issue: Verify that your branch is up to date with master

devolutions-gateway/src/config.rs
```diff
@@ -309,20 +309,33 @@ impl Conf {
...
+            }
+
+            let arch_path = format!(
+                "/usr/lib/{arch}-linux-gnu/devolutions-gateway/libxmf.so",
+                arch = env::consts::ARCH,
+            );
+            return Some(Utf8PathBuf::from(arch_path));
+        }
+
+        None
```
CBenoit, line 338: suggestion: In this case, I think it’s best to stick to using `cfg!` with regular Rust code.

```diff
diff --git a/devolutions-gateway/src/config.rs b/devolutions-gateway/src/config.rs
index 9037ac37..46e29059 100644
--- a/devolutions-gateway/src/config.rs
+++ b/devolutions-gateway/src/config.rs
@@ -315,27 +315,21 @@ impl Conf {
             return Some(path.to_owned());
         }

-        #[cfg(target_os = "windows")]
-        {
+        if cfg!(target_os = "windows") {
             let path = env::current_exe().ok()?.parent()?.join("xmf.dll");
             return Utf8PathBuf::from_path_buf(path).ok();
-        }
-
-        #[cfg(target_os = "linux")]
-        {
-            #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
-            {
-                return None;
-            }
-
+        } else if cfg!(all(
+            target_os = "linux",
+            not(any(target_arch = "x86_64", target_arch = "aarch64"))
+        )) {
             let arch_path = format!(
                 "/usr/lib/{arch}-linux-gnu/devolutions-gateway/libxmf.so",
                 arch = env::consts::ARCH,
             );
-            return Some(Utf8PathBuf::from(arch_path));
+            Some(Utf8PathBuf::from(arch_path))
+        } else {
+            None
         }
-
-        None
     }
 }
```

```rust
        if cfg!(target_os = "windows") {
            let path = env::current_exe().ok()?.parent()?.join("xmf.dll");
            return Utf8PathBuf::from_path_buf(path).ok();
        } else if cfg!(all(
            target_os = "linux",
            any(target_arch = "x86_64", target_arch = "aarch64")
        )) {
            let path = format!(
                "/usr/lib/{arch}-linux-gnu/devolutions-gateway/libxmf.so",
                arch = env::consts::ARCH,
            );
            Some(Utf8PathBuf::from(path))
        } else {
            None
        }
```

Notice that I also combined the conditions using `all` so that we are not increasing the number of branches.

rationale:
- The code is both simpler to read and to write, with less sigils and easier code flow.
- The guarded code is compiled and checked on all platforms, but optimized out during the compilation. It’s a nice property because it means that all branches needs to be valid for the code to compile at all, no matter the platform used to compile it. We get rich information from lsp on all platforms, and so on. It’s not always possible or desirable to use `cfg!` hence the existence of `#[cfg]`, but I would typically avoid reaching for that first.

package/Linux/gateway/template/rules
```diff
@@ -1,4 +1,16 @@
 #!/usr/bin/make -f
+
+ifeq ($(DEB_HOST_ARCH),amd64)
+ARCH = x86_64
+else ifeq ($(DEB_HOST_ARCH),arm64)
+ARCH = aarch64
+else
+$(error Unsupported architecture: $(DEB_HOST_ARCH))
+endif
```
CBenoit, line 9: note: libxmf is an optional dependency, so we could still allow packaging to happen, but without libxmf.so. We’re not going to deliver artifacts ourselves, but it’s fine to still allow the packaging on other platforms if someone wants to do it.

CBenoit:
> According to the [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html), `/usr/share` is for architecture-independent data files.

Good catch.

> This moves _libxmf.so_ from `/usr/share/` to `/usr/lib/x86-64-linux-gnu/` or `/usr/lib/aarch64-linux-gnu/`.

I was going to argue about that, but I see that it’s actually placed under another subfolder called `devolutions-gateway` giving, e.g.: `/usr/lib/x86-64-linux-gnu/devolutions-gateway/libxmf.so`. I would be against installing libxmf.so directly in `/usr/lib` without this namespacing because for packaging and installation simplicity, as you know, we are not packaging a libxmf package on which devolutions-gateway then depends on. Instead the the library is part of the devolutions-gateway package. It’s also useful because we may need to package different versions of libxmf for different softwares (agent vs gateway). Even if we typically want updated versions at all time, the user may want to update one or the other software on their own schedule, and the agent can’t rely on the gateway being installed and providing the right version either. In the first place, the package manager would complain about the same file being touched by two different packages. I think we are on the same page, but I just wanted to comment on that.

From here, I have a question on how it’s implemented.
Do we really need to put the library behind an architecture folder like this?
What about `/usr/lib/devolutions-gateway/libxmf.so`? 
Typically, a user should not install two different versions of the same package for different architectures.
Only the package for the architecture of his machine.
Most packages I know install the library without architecture namespacing because it typically does not make sense to namespace in the first place.
For this reason, "portable systems" installed on USB keys typically needs to target a specific architecture too.
I’m not sure of all it would take to achieve such a portable system, but it’s probably not our job to care about that.

Besides, I don’t feel very good about the Rust-to-Debian architecture name conversion.
Removing this architecture logic means that we can simplify the new code, removing this conversion as well.

Get rid of this logic in the rules file:
```make
ifeq ($(DEB_HOST_ARCH),amd64)
ARCH = x86_64
else ifeq ($(DEB_HOST_ARCH),arm64)
ARCH = aarch64
else
$(error Unsupported architecture: $(DEB_HOST_ARCH))
endif
```

Get rid of this compilation check: `any(target_arch = "x86_64", target_arch = "aarch64")`.

This:
```rust
            let path = format!(
                "/usr/lib/{arch}-linux-gnu/devolutions-gateway/libxmf.so",
                arch = env::consts::ARCH,
            );
            Some(Utf8PathBuf::from(path))
```

Becomes this:
```rust
            Some(Utf8PathBuf::from("/usr/lib/devolutions-gateway/libxmf.so"))
```

It’s fine if the file does not exist. As said, this is an optional dependency, and we’ll only log a warning about this:

https://github.com/Devolutions/devolutions-gateway/blob/dd17375552f13f97151138fd938514f977b1af35/devolutions-gateway/src/service.rs#L91-L103

Let’s not assume that the library can’t exist on other platforms, as it’s perfectly possible for someone to build it from sources for its platform too. If someone really wants to make a build for their hardware, we don’t need to make their life harder.

CBenoit:
Also, this is a bit borderline as we also change the code a little bit, but I think the commit type should be `build`, as this is mostly about stuff external to the application itself.

devolutions-gateway/src/config.rs
```diff
@@ -309,17 +309,20 @@ impl Conf {
...
             let path = env::current_exe().ok()?.parent()?.join("xmf.dll");
-            Utf8PathBuf::from_path_buf(path).ok()
-        } else if cfg!(target_os = "linux") {
-            Some(Utf8PathBuf::from("/usr/share/devolutions-gateway/libxmf.so"))
+            return Utf8PathBuf::from_path_buf(path).ok();
+        } else if cfg!(all(
+            target_os = "linux",
+            any(target_arch = "x86_64", target_arch = "aarch64")
+        )) {
+            Some(Utf8PathBuf::from("/usr/lib/devolutions-gateway/libxmf.so"))
```
CBenoit, line 325: issue: With multi-arch code removed, this should be:

```suggestion
        } else if cfg!(target_os = "linux") {
            Some(Utf8PathBuf::from("/usr/lib/devolutions-gateway/libxmf.so"))
```

devolutions-gateway/src/config.rs
```diff
@@ -309,17 +309,17 @@ impl Conf {
...
+        }
 
         if let Some(path) = self.debug.lib_xmf_path.as_deref() {
             return Some(path.to_owned());
         }
 
         if cfg!(target_os = "windows") {
             let path = env::current_exe().ok()?.parent()?.join("xmf.dll");
-            Utf8PathBuf::from_path_buf(path).ok()
+            return Utf8PathBuf::from_path_buf(path).ok();
```
CBenoit, line 320: nitpick: The return is not required.

CBenoit (approved):
LGTM! Thank you!

PR #1176, opened 2025-01-06
build(dgw): correct typo causing missing packager

allan2:
The control template expects a variable named `packager` but `package` was passed.


CBenoit (approved):
Good catch! Thank you.

PR #1169, opened 2024-12-23
ci: build player artifacts on demand

irvingoujAtDevolution:

PR #1179, opened 2025-01-06
build: add RPM for gateway and agent

allan2:
This commit adds RPM packages for Gateway and Agent to the release assets.

`tlk.ps1 -TlkVerb package -Platform linux` now generates both a deb and an rpm for the specified product.

The rpm is generated with fpm, a Linux packaging tool.
In CI, no cache is used for the fpm dependency since the cache would not persist across release jobs.

The RPM includes all the assets of the corresponding Debian package, including the changelog, copyright, maintainer scripts, and webapp/libxmf.so for Gateway.

The changelog is copied directly from Debian. A separate PR will be made to convert it to an RPM-friendly changelog format.

Tested with RHEL 9 (glibc 2.34).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -606,6 +606,7 @@ jobs:
           }
 
           $Env:DGATEWAY_WEBCLIENT_PATH = Join-Path "webapp" "client" | Resolve-Path
+          $Env:DGATEWAY_WEBPLAYER_PATH = Join-Path "webapp" "player" | Resolve-Path
 
           ./ci/tlk.ps1 package -Product gateway -Platform ${{ matrix.os }} -Architecture ${{ matrix.arch }} -CargoProfile ${{ needs.preflight.outputs.rust-profile }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -553,6 +553,10 @@ jobs:
           rustup target add aarch64-unknown-linux-gnu
           echo "STRIP_EXECUTABLE=aarch64-linux-gnu-strip" >> $GITHUB_ENV
 
+      - name: Install fpm
+        if: matrix.os == 'Linux'
+        run: sudo gem install --no-document fpm
+
       - name: Configure Windows runner
         if: matrix.os == 'windows'
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -727,6 +727,10 @@ jobs:
         if: matrix.os == 'windows'
         uses: microsoft/setup-msbuild@v2
 
+      - name: Install fpm
+        if: runner.os == 'Linux'
+        run: sudo gem install --no-document fpm
+
       - name: Build
         shell: pwsh
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -702,6 +702,10 @@ jobs:
           rustup target add aarch64-unknown-linux-gnu
           echo "STRIP_EXECUTABLE=aarch64-linux-gnu-strip" >> $GITHUB_ENV
 
+      - name: Install fpm
+        if: matrix.os == 'Linux'
+        run: sudo gem install --no-document fpm
+
       - name: Configure Windows runner
         if: matrix.os == 'windows'
         run: |
@@ -730,10 +734,6 @@ jobs:
       - name: Add msbuild to PATH
         if: matrix.os == 'windows'
         uses: microsoft/setup-msbuild@v2
-
-      - name: Install fpm
-        if: runner.os == 'Linux'
-        run: sudo gem install --no-document fpm
 
       - name: Build
         shell: pwsh
--- a/ci/tlk.ps1
+++ b/ci/tlk.ps1
@@ -823,9 +823,6 @@ class TlkRecipe
             $DpkgBuildPackageArgs += @('-a', $this.Target.DebianArchitecture())
         }
 
-        # Strip first so the included binary is identical for deb and rpm.
-        & 'strip' $Executable | Out-Host
-
         # Disable dpkg-buildpackage stripping as the binary is already stripped.
         $Env:DEB_BUILD_OPTIONS = "nostrip"
         & 'dpkg-buildpackage' $DpkgBuildPackageArgs | Out-Host
--- a/ci/tlk.ps1
+++ b/ci/tlk.ps1
@@ -807,10 +807,14 @@ class TlkRecipe
             date = $($(Get-Date -UFormat "%a, %d %b %Y %H:%M:%S %Z00") -Replace '\.','')
         } -OutputFile $ChangelogFile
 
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -727,6 +727,10 @@ jobs:
         if: matrix.os == 'windows'
         uses: microsoft/setup-msbuild@v2
 
+      - name: Install fpm
```
thenextman, line 730: Is it in the right place? This is the devolutions-agent job, but the PR references Gateway

.github/workflows/ci.yml
```diff
@@ -727,6 +727,10 @@ jobs:
...
+        if: runner.os == 'Linux'
```
thenextman, line 731: minor nit: maybe use matrix.os; it doesn't really matter here but in a cross compilation environment we could be building on a Linux runner but targeting another OS. The matrix.os holds the OS we are targetting.

thenextman:
> This PR is not ready for review yet. I'm still adding things to bring it to parity with the Debian package.

As a hint, can be good to make the PR a draft in this case. Because I think the repo automatically adds arch-maintainers as reviewer, which generates an email prompting for review :)

ci/tlk.ps1
```diff
@@ -823,6 +823,10 @@ class TlkRecipe
             $DpkgBuildPackageArgs += @('-a', $this.Target.DebianArchitecture())
         }
 
+        # Strip first so the included binary is identical for deb and rpm
+        & 'strip' $Executable | Out-Host
```
CBenoit, line 827: question: Maybe we should strip somewhere else? There is an option in Cargo.toml to enable automatic stripping in profiles: https://doc.rust-lang.org/cargo/reference/profiles.html#strip
Configuring this appropriately for the production profile, and remove manual calls to `strip` would be superior I think.

ci/tlk.ps1
```diff
@@ -823,6 +823,10 @@ class TlkRecipe
...
+
+        $env:DEB_BUILD_OPTIONS = "nostrip"
```
CBenoit, line 829: nitpick: Don’t forget a comment explaining why we disable strip at this place. "Binary already stripped by Cargo in production profile"

CBenoit, line 829: style: It’s PowerShell so technically valid… but usually `$Env` is spelled with a leading uppercase.

ci/tlk.ps1
CBenoit: > Keeping symbols in the release build can be useful for troubleshooting.

Are you thinking about troubleshooting using a local "release build" on your dev machine?

For binaries that we ship, separating the debug symbols from the binaries and only delivering the binaries without the symbols is still superior (splitting the symbols from the binary). We are doing that at least for Windows builds, not yet for Linux and macOS builds. There is also an option in Cargo that may be used for splitting debug symbols: https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo
But here we are just stripping unconditionally in the script, so I’ll assume you are referring to testing with a local build that is not going through the script.

So for local builds, don’t worry. I’m not asking that we modify the _release profile_, but the _production profile_.
We defined additional profiles in the top-level Cargo.toml:
https://github.com/Devolutions/devolutions-gateway/blob/0ed70d21bac2cbf31dcc9dbe5615a8e573b73717/Cargo.toml#L23-L29

The production profile is only used when we produce final artifacts that we want to ship, in the CI, and this is a parameter of the script itself:
https://github.com/Devolutions/devolutions-gateway/blob/0ed70d21bac2cbf31dcc9dbe5615a8e573b73717/.github/workflows/ci.yml#L117-L127
https://github.com/Devolutions/devolutions-gateway/blob/0ed70d21bac2cbf31dcc9dbe5615a8e573b73717/.github/workflows/ci.yml#L270

I think it’s a more declarative and flexible way of handling this problem. What do you think? I’m open to hear otherwise if I’m missing some drawback.

ci/tlk.ps1
```diff
@@ -674,6 +676,7 @@ class TlkRecipe
                 } else {
                     throw ("Specify DGATEWAY_WEBCLIENT_PATH environment variable")
                 }
+                $DGatewayWebPlayer = "$($this.SourcePath)/webapp/player"
```
thenextman, line 679: disclaimer: I'm probably just not understanding (I'm only reading the PR without full context)

The player comes direct from the source tree (i.e. it's not something compiled)? 

Just reading this implies that there should be something in ./webapp/player, but I don't see anything in that path when I browse the repo?

thenextman:
I queried one more thing that I'm probably just missing context for.

Wouldn't it be good if we could build the .deb using dpkg as well? As it stands now, if I want to modify the packaging, I need to understand both dpkg and fpm. I understood the promise of fpm was that we could build different package types and have all that inner knowledge abstracted away.

Anyway, just a thought; I'm not suggesting making it part of this PR. But it would be a nice additional improvement in future!

ci/tlk.ps1
CBenoit: I think this is good. Can you check if you get the same binary size with this option enabled without running `strip`, and this optional disabled but running `strip`?

ci/tlk.ps1
CBenoit: Probably a good catch. @irvingoujAtDevolution  modified that recently, maybe this branch wasn’t up to date?
- https://github.com/Devolutions/devolutions-gateway/pull/1169
- https://github.com/Devolutions/devolutions-gateway/pull/1163

ci/tlk.ps1
CBenoit: Oh… I’m sorry I didn’t mean to check with the various optimization profiles :joy:
Just wanted to see if `strip = "symbols"` was indeed identical to calling `strip` manually for the same profile, because there are other possible values for this option. Anyway, it’s likely the right option :)

ci/tlk.ps1
thenextman: Excellent, thanks!

thenextman:
> Two more things are unresolved:
> 
> 1. dependencies are missing in rpm. I'm currently working on this.
> 2. [_WindowsManager/Program.cs_](https://github.com/Devolutions/devolutions-gateway/blob/master/package/WindowsManaged/Program.cs#L79-L101) has a check for the web client path but not for the web player path. Should I add it in?

Yes, if you can add the environment variable and usage into the managed installer, that would be great. Can just follow what is there for the web client currently. Thank you! 💯

thenextman:
One more thing if you wouldn't mind @allan2, can you update the file spec here https://github.com/Devolutions/devolutions-gateway/blob/0ed70d21bac2cbf31dcc9dbe5615a8e573b73717/package/WindowsManaged/Program.cs#L314 to use the property you added instead the hardcoded path? Can be a followup PR if that suits you better.

CBenoit (commented):
Good job Allan!

thenextman (approved):

PR #1180, opened 2025-01-07
refactor: enable clippy::needless_return lint

allan2:


CBenoit (commented):
This lint is warn-by-default, can you double check if it is really necessary to add this in the Cargo.toml?

image

PR #1181, opened 2025-01-08
fix(dgw): fallback to recording playback when shadowing is not possible

irvingoujAtDevolution:
This happens when the session is already terminated.

https://github.com/user-attachments/assets/b92a5c5e-cdf4-4191-819d-44c5a76b722c

### DIFF
--- a/webapp/player-project/src/main.ts
+++ b/webapp/player-project/src/main.ts
@@ -1,6 +1,9 @@
 import { GatewayAccessApi } from './gateway';
+import { showNotification } from './notification.ts';
 import { getPlayer } from './players/index.js';
-import { getShadowPlayer } from './streamers/index.js';
+import { cleanUpStreamers, getShadowPlayer } from './streamers/index.js';
+import './ws-proxy.ts';
+import { OnBeforeClose as BeforeWebsocketClose } from './ws-proxy.ts';
 
 async function main() {
   const { sessionId, token, gatewayAccessUrl, isActive } = getSessionDetails();
@@ -23,6 +26,18 @@ async function playSessionShadowing(gatewayAccessApi) {
   try {
     const recordingInfo = await gatewayAccessApi.fetchRecordingInfo();
     const fileType = getFileType(recordingInfo);
+    BeforeWebsocketClose((closeEvent) => {
+      if (closeEvent.code !== 1000) {
+        // faild, try to play static recording
+        cleanUpStreamers();
+        playStaticRecording(gatewayAccessApi);
+        showNotification('Session may have ended,playing recording instead', 'info');
+        return {
+          ...closeEvent,
+          code: 1000, // for avoid extra handling by other listeners
+        };
+      }
+    });
 
     getShadowPlayer(fileType).play(gatewayAccessApi);
   } catch (error) {
--- a/webapp/player-project/src/main.ts
+++ b/webapp/player-project/src/main.ts
@@ -32,8 +32,9 @@ async function playSessionShadowing(gatewayAccessApi) {
         playStaticRecording(gatewayAccessApi);
         return {
           ...closeEvent,
-          // for avoid extra handling by other listeners, for asciinema-player particularly in this case.
-          // see asciinema-player websocket driver at on socket close handler
+          // This prevents extra handling by other listeners, particularly for asciinema-player in this scenario.
+          // For more details, see the asciinema-player WebSocket driver’s socket close handler.
+          // https://github.com/asciinema/asciinema-player/blob/c09e1d2625450a32e9e76063cdc315fd54ecdd9d/src/driver/websocket.js#L219
           code: 1000,
         };
       }
--- a/webapp/player-project/src/main.ts
+++ b/webapp/player-project/src/main.ts
@@ -27,7 +27,7 @@ async function playSessionShadowing(gatewayAccessApi) {
     const fileType = getFileType(recordingInfo);
     BeforeWebsocketClose((closeEvent) => {
       if (closeEvent.code !== 1000) {
-        // faild, try to play static recording
+        // The session playback failed; attempt to play the recording as usual as a fallback.
         cleanUpStreamers();
         playStaticRecording(gatewayAccessApi);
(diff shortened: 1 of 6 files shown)


webapp/player-project/src/main.ts
```diff
@@ -23,6 +26,18 @@ async function playSessionShadowing(gatewayAccessApi) {
   try {
     const recordingInfo = await gatewayAccessApi.fetchRecordingInfo();
     const fileType = getFileType(recordingInfo);
+    BeforeWebsocketClose((closeEvent) => {
+      if (closeEvent.code !== 1000) {
+        // faild, try to play static recording
+        cleanUpStreamers();
+        playStaticRecording(gatewayAccessApi);
+        showNotification('Session may have ended,playing recording instead', 'info');
```
CBenoit, line 34: suggestion: Since this shown to the user, let’s observe strict rules about ponctuations (a comment must be followed by a space). That said, I suggest the following wording, what do you think?
 
```suggestion
        showNotification('Session already ended. The recording will be played from the start.', 'info');
```

We’ll have to think about internalization too if we’re really going to have messages displayed like this.

irvingoujAtDevolution: True, I was thinking about emit the message back to the consumer, but it will be hard for RDM to catch it.

CBenoit: I sent a Slack message in our Slack channel for further discussion about this!

webapp/player-project/src/main.ts
```diff
@@ -23,6 +26,18 @@ async function playSessionShadowing(gatewayAccessApi) {
...
+        return {
+          ...closeEvent,
+          code: 1000, // for avoid extra handling by other listeners
```
CBenoit, line 37: nitpick: I understand you are not used to it, but I would appreciate if you could follow this style more closely.
https://github.com/Devolutions/IronRDP/blob/master/STYLE.md#inline-code-comments-are-proper-sentences

The important thing with this suggestion is:

> Rationale: writing a sentence (or maybe even a paragraph) rather just "a comment" creates a more appropriate frame of mind. **It tricks you into writing down more of the context you keep in your head while coding**.

At my level, I don’t really understand the connection between "1000" and "avoid extra handling by other listeners", suggesting the comment could give me more relevant context to understand what is happening. Of course, I could start digging up the code to try to understand why this is good or necessary, but it’s always nice if a nice comment can save me this time.

CBenoit:
Do you have a Jira ticket for this task? If so, add the footer: `Issue: DGW-XXX` in the PR description, and make sure it is included in the squash commit.

irvingoujAtDevolution:
No, this is not related to any existing Jira tickets

webapp/player-project/src/main.ts
irvingoujAtDevolution: updated , it is for asciinema-player, I added the link of the specific line that would cause problem in the comments

CBenoit: Did you forget to push your commit? I don’t see the change. Also, what do you mean by link? Since the codebase is evolving, the link may become outdated at anytime. Instead, describes using module name, function name, and other stable descriptions. It’s also okay to define anchors. For instance:

```rust
// ANCHOR: some_component
let foo = bar;
```

```rust
// Search for the some_component anchor to see where this is relevant.
```

irvingoujAtDevolution: Sorry, my muscle memory makes switched me to another terminal without looking at the push result properly. I push the proper comment, and it is a link to the asciinema-player github project at a line where code 1000 is required.

CBenoit: Oh, it’s a different GitHub project? Then forget what I said, the link is fine!

webapp/player-project/src/main.ts
```diff
@@ -23,6 +25,19 @@ async function playSessionShadowing(gatewayAccessApi) {
...
+    BeforeWebsocketClose((closeEvent) => {
+      if (closeEvent.code !== 1000) {
+        // faild, try to play static recording
+        cleanUpStreamers();
+        playStaticRecording(gatewayAccessApi);
+        return {
+          ...closeEvent,
+          // for avoid extra handling by other listeners, for asciinema-player particularly in this case.
+          // see asciinema-player websocket driver at on socket close handler
+          code: 1000,
```
CBenoit, line 37: suggestion: Here is an improved wording. Both a stable description and a link are provided since this is an _external_ repository.

```suggestion
          // This prevents extra handling by other listeners, particularly for asciinema-player in this scenario.
          // For more details, see the asciinema-player WebSocket driver’s socket close handler.
          // https://github.com/asciinema/asciinema-player/blob/c09e1d2625450a32e9e76063cdc315fd54ecdd9d/src/driver/websocket.js#L219
          code: 1000,
```

webapp/player-project/src/streamers/index.ts
```diff
@@ -17,3 +17,12 @@ export const getShadowPlayer = (fileType) => {
 
   return player;
 };
+
+export const cleanUpStreamers = () => {
+  // Clean up any existing shadow-player elements
```
CBenoit, line 22: suggestion: This comment is probably unnecessary, but if you want to keep it, here is a better wording.
```suggestion
  // Remove all shadow-player elements.
```

CBenoit (approved):
I’m sorry for the pedantic comments, but it would really help me as a reviewer, and all future readers of the code, including yourself, if you could take some time to consider what I said carefully, and include the suggested practices in your coding style in your future contributions. I left and applied new comments, and I’ll leave you double check and merge the PR. If you struggle with English redaction, it’s fine to let an AI help you with that. ChatGPT or Copilot will do a great job at improving the text itself, but you need to write with intention. Thank you.

webapp/player-project/src/streamers/index.ts
CBenoit: Note that in this case, it’s much more useful to describe the _why_, rather than the _what_. Maybe it makes more sense to explain the why at the callsite of this function.

PR #1182, opened 2025-01-08
fix(dgw): remove success notification from player

irvingoujAtDevolution:
Changelog: ignore

### DIFF
--- a/webapp/player-project/src/streamers/webm.ts
+++ b/webapp/player-project/src/streamers/webm.ts
@@ -20,10 +20,6 @@ export async function handleWebm(gatewayAccessApi: GatewayAccessApi) {
   shadowPlayer.srcChange(gatewayAccessApi.sessionShadowingUrl());
   shadowPlayer.play();
 
-  shadowPlayer.onEnd(() => {
-    showNotification('Playback has ended', 'success');
-  });
-
   shadowPlayer.onError((error) => {
     showNotification(`An error occurred: ${error}`, 'error');
   });


webapp/player-project/src/streamers/webm.ts
```diff
@@ -20,13 +19,5 @@ export async function handleWebm(gatewayAccessApi: GatewayAccessApi) {
   shadowPlayer.srcChange(gatewayAccessApi.sessionShadowingUrl());
   shadowPlayer.play();
 
-  shadowPlayer.onEnd(() => {
-    showNotification('Playback has ended', 'success');
-  });
-
-  shadowPlayer.onError((error) => {
-    showNotification(`An error occurred: ${error}`, 'error');
-  });
```
CBenoit, line 29: If we remove this, how can we learn more about the error?

irvingoujAtDevolution: Maybe we can show an error icon or animation? 🤔

CBenoit: To be honest, I think it’s best if we keep showing the error, even if it’s not localized :thinking: 
Let’s stick to removing only strings that are not critical, and hope that errors are not coming up too often.

irvingoujAtDevolution: ok,I get your idea, I'll keep the notification, for playback end message, I'll use a replay button to replace it

PR #1186, opened 2025-01-11
fix(dgw): send close code 1005 when video streaming ends

irvingoujAtDevolution:
Trying to make the `websocket_compact ` function return a result with a new trait.
I tried to make a Generic type for the `WithExplict` but faild as the combination for With and Map trait generic and bounds is beyond my comprehension. I hope @CBenoit  could make a better type than this... 
feel free to edit it, and let me know your thoughts, thanks!

### DIFF
--- a/crates/video-streamer/src/streamer/mod.rs
+++ b/crates/video-streamer/src/streamer/mod.rs
@@ -23,6 +23,7 @@ pub(crate) mod signal_writer;
 pub(crate) mod tag_writers;
 
 use crate::{reopenable::Reopenable, StreamingConfig};
+use tokio::io::AsyncWriteExt;
 
 #[instrument(skip_all)]
 pub fn webm_stream(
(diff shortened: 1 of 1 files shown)


irvingoujAtDevolution:
@CBenoit Thanks for much easier and cleaner advice!

crates/video-streamer/src/streamer/mod.rs
```diff
@@ -23,6 +23,7 @@ pub(crate) mod signal_writer;
 pub(crate) mod tag_writers;
 
 use crate::{reopenable::Reopenable, StreamingConfig};
+use tokio::io::AsyncWriteExt;
```
CBenoit, line 26: style: You may want to put the imports above, with the others!

CBenoit (approved):
LGTM! I’ll leave you merge the PR! Don’t forget to update the description to include a changelog-friendly message, or to add `Changelog: ignore` if the change is for something that we didn’t release yet. (And I think that’s indeed the case!)

PR #1192, opened 2025-01-14
fix(dgw): streamer error handling with custom code

irvingoujAtDevolution:
Custom error handling in streaming endpoint and the streamer. 
Since we are going to have localization in the short future, this change makes error handling more clear and manageable.

Changelog: ignore

### DIFF
--- a/devolutions-gateway/src/api/jrec.rs
+++ b/devolutions-gateway/src/api/jrec.rs
@@ -493,10 +493,19 @@ where
 
 // Code from 4000 to 4999 are reserved for private custom use
 // https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code
-enum StreamerWebsocketCloseCode {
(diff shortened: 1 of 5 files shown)


CBenoit (approved):
naming: We could think of a shorter name for `StreamerWebsocketCloseCode`. Since it’s a private enum to the module, e.g.: `CloseCode` or `StreamerCloseCode` is probably good enough.

suggestion: Define`From<StreamerWebsocketCloseCode> for CloseFrame`. You can include a stringified version of the close code in the reason field, though it typically does not need to be redundant, it may be good for easier inspecting in the web browser network tool.

note: As the feature has not been released yet, it’s not relevant to keep every small changes in the CHANGELOG.md file, and I think this should be `Changelog: ignore` as well. Once we release the shadow player for real, this will become relevant again.

Feel free to merge once you’re satisfied.

PR #1195, opened 2025-01-16
chore: convert CRLF to LF; remove BOM

allan2:
This PR converts line endings to follow _.gitattributes_.

Some files that should be LF were CRLF. This PR fixes all instances. Every `.sln` file had a CRLF which is intended, so no changes there.

Reviewers can use
```
git diff master..line-endings --ignore-cr-at-eol 
```
to verify there is no real code change.

The above command still shows some changed lines like this:

```
--- a/crates/devolutions-pedm/DevolutionsPedmDesktop/AppView.resx
+++ b/crates/devolutions-pedm/DevolutionsPedmDesktop/AppView.resx
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```

This is because a BOM character was removed.

```
> git checkout master
> file crates/devolutions-pedm/DevolutionsPedmDesktop/AppView.resx
crates/devolutions-pedm/DevolutionsPedmDesktop/AppView.resx: XML 1.0 document, Unicode text, UTF-8 (with BOM) text, with CRLF line terminators

> git checkout line-endings
> file crates/devolutions-pedm/DevolutionsPedmDesktop/AppView.resx
crates/devolutions-pedm/DevolutionsPedmDesktop/AppView.resx: XML 1.0 document, ASCII text
```


CBenoit (approved):
Thank you! Did you use a specific tool to go through it?

PR #1199, opened 2025-01-23
ci: build the PEDM desktop app so it can be signed

thenextman:
Build the PEDM desktop application independently - it's a .NET application and architecture "neutral". The artifacts are uploaded in CI, then extracted and signed before packaging.

Fix some minor issues that cropped up while debugging the workflows.


CBenoit (approved):
LGTM!

PR #1202, opened 2025-01-29
fix(dgw,jetsocat): implement WebSocket keep-alive logic

CBenoit:
Our WebSockets are already responding Pong messages to Ping messages, but they were never sending Ping messages.

### DIFF
--- a/crates/transport/src/pinnable.rs
+++ b/crates/transport/src/pinnable.rs
@@ -1,0 +1,122 @@
+use std::pin::Pin;
+use std::{fmt, ops};
+
+/// A wrapper for [`parking_lot::Mutex`] that supports obtaining `Pin<&mut T>` references to the contained value.
+///
+/// [`parking_lot::Mutex<T>`] itself does not have structural pinning.
+/// The pinned-ness of the mutex type does not propagate to the field (the `T`).
+///
+/// `Pin<PinnableMutex<T>>` however can return a `Pin<MutexGuard<T>>`.
+/// It’s a trade-off, because it can no longer provide mutable access without being pinned.
+///
+/// # Example
+///
+/// ```
+/// use std::future::Future;
+/// use std::pin::Pin;
+/// use std::sync::Arc;
+/// use std::task::{Context, Poll};
+/// use transport::pinnable::PinnableMutex;
+///
+/// fn poll_shared_future<F: Future>(
+///     fut: &Pin<Arc<PinnableMutex<F>>>,
+///     ctx: &mut Context<'_>,
+/// ) -> Poll<F::Output> {
+///     fut.as_ref().lock().as_mut().poll(ctx)
+/// }
+/// ```
+pub struct PinnableMutex<T: ?Sized>(parking_lot::Mutex<T>);
+
+impl<T: ?Sized> fmt::Debug for PinnableMutex<T>
+where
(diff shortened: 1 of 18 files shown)


crates/transport/src/pinnable.rs
```diff
@@ -0,0 +1,122 @@
...
+/// }
+/// ```
+pub struct PinnableMutex<T: ?Sized>(parking_lot::Mutex<T>);
+
+impl<T: ?Sized> fmt::Debug for PinnableMutex<T>
+where
+    T: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(&self.0, f)
```
pacmancoder, line 35: Question: Doesn't `#[derive(Debug)]` produce the same output?

crates/transport/src/shared.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl<S, T, E> Sink<T> for Shared<S>
+where
+    S: Sink<T, Error = E>,
+{
+    type Error = E;
+
+    fn poll_ready(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.as_ref().lock().as_mut().poll_ready(cx)
```
pacmancoder, line 47: Question: Blocking on sync mutex in async future feels wrong, does using async mutex reduce performance drastically?

pacmancoder (approved):
Looks good! Approved in advance, however I am a bit confused about `parking_lot` sync Mutex usage in futures, could you elaborate on this approach?

crates/transport/src/pinnable.rs
CBenoit: It's not derivable because T does not necessarily implement Debug. Unless they improved that recently?

CBenoit:
> Looks good! Approved in advance, however I am a bit confused about `parking_lot` sync Mutex usage in futures, could you elaborate on this approach?
> 
> 

Of course! I should include that as a comment in the module: https://github.com/Devolutions/devolutions-gateway/pull/1202#discussion_r1933620799

crates/transport/src/shared.rs
pacmancoder: Thank you! Interesting, I'll take a read

crates/transport/src/pinnable.rs
pacmancoder: Huh,  it compiled successfully on playground at least https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=238783f816438c16430a790e0df7ddb4

crates/transport/src/shared.rs
CBenoit: I pushed a commit to improve the documentation directly in the code: https://github.com/Devolutions/devolutions-gateway/pull/1202/commits/9b424241a33504d97147e7c7e1bc4b297b81e7aa

CBenoit:
Also implemented the keepalive logic for Gateway -> Target server in /jet/fwd/tls route: https://github.com/Devolutions/devolutions-gateway/pull/1202/commits/17f0a955faf55dd6a84351cabc67adbe9f32ce3d

PR #1207, opened 2025-01-31
fix(agent): fix downgrade logic for unreleased Devolutions Gateway

pacmancoder:
Fix the updater corner case when the user is trying to downgrade to the latest specified version in `productinfo.htm` while a newer unreleased version is installed locally.


CBenoit (approved):
LGTM!

PR #1213, opened 2025-02-07
feat(installer): register PEDM shell extension instead of msix

thenextman:
Register the "Run Elevated" shell extension as a traditional context menu extension, via the registry. This will appear in the second level context menu (after clicking "Show more options") on a modern Win11 system.

For now, registration of the `IExplorerCommand` "modern" shell extension (to appear in the top-level context menu on a Win11 system) is removed. This is unstable: it's tricky to register the extension for multiple users, and it causes significant problems with upgrades or uninstall/reinstall, due to the DLL remaining locked by the shell. A future pull request will reinstate the .msix registration with a different approach.

Remove some unused code that was carried out needlessly from Devolutions Gateway.

### DIFF
--- a/package/AgentWindowsManaged/Actions/AgentActions.cs
+++ b/package/AgentWindowsManaged/Actions/AgentActions.cs
@@ -237,48 +237,47 @@ internal static class AgentActions
         Condition = Features.PEDM_FEATURE.BeingInstall(),
     };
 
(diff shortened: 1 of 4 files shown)


CBenoit (approved):
Some checks are failed, but LGTM! Thank you!

awakecoding (approved):

PR #1222, opened 2025-02-13
fix(dgw): notify all instead of notify one when recording ends

irvingoujAtDevolution:
Fix the issue where recording ended but the  streaming task keeps alive, this will result in streaming client not received correct update of the streaming state.


CBenoit (approved):
LGTM!

However, if this PR is fixing a big that could actually be observed, please describe that in a more concrete way. Someone unfamiliar with the codebase can’t understand what the current title is about, so it’s not good for the release notes.

I’ll let you merge with a better description that we can use in the release notes.

Issue #1223, opened 2025-02-14
Gateway service name is mistakenly changed to devolutions-agent


linuxgemini:
With 651d8cf89316ae694422b8b9e50bb93733aec236 from last month, the service name for Gateway became `devolutions-agent` for Linux and macOS:

https://github.com/Devolutions/devolutions-gateway/blob/404ba8aa11a33b3772ee359a76d0c911f84992f3/crates/devolutions-agent-shared/src/lib.rs#L16-L36

This is used when registering the service for SystemD etc.

https://github.com/Devolutions/devolutions-gateway/blob/404ba8aa11a33b3772ee359a76d0c911f84992f3/devolutions-gateway/src/service.rs#L18

---

However the Devolutions documentation references `devolutions-gateway` as the service name:

https://docs.devolutions.net/dgw/getting-started/configuration/linux-configuration/#restart-devolutions-gateway:

![Image](https://github.com/user-attachments/assets/1a8ea2fe-13d9-4a8b-a206-cd3f44a49969)

https://docs.devolutions.net/dgw/getting-started/configuration/linux-configuration/#connect-devolutions-gateway-to-devolutions-server:

![Image](https://github.com/user-attachments/assets/07a7015b-2593-4d47-aed7-0e7df3713ee9)

https://docs.devolutions.net/dgw/getting-started/configuration/linux-configuration/#enable-devolutions-gateway-web-interface:

![Image](https://github.com/user-attachments/assets/274a30d1-2213-448b-b0e0-c715f735b240)

CBenoit:
Hi!

Good catch! Actually, we were reported the same issue internally but didn’t follow up yet.
However, the name of the service was not changed deliberately.
We have two services: `devolutions-gateway` and `devolutions-agent`.
Most certainly, the file `devolutions-gateway/src/service.rs` was mistakenly modified instead of the file `devolutions-agent/src/service.rs`.

This is a bug.

(cc @allan2 )

linuxgemini:
Heya, I changed the title of the issue to reflect the issue properly.

PR #1224, opened 2025-02-14
fix(deb): improve deb package

linuxgemini:
~~This PR is somewhat related to #1223 as currently you cannot uninstall `devolutions-gateway` properly (on Ubuntu) without modifying `/var/lib/dpkg/info/devolutions-gateway.prerm` because of the service name change making the `systemctl stop` command.~~ Post-publication edit: #1223 is a bug, hence the modifications related to this are removed.

~~And although not utilized by the service file, this PR also adds the functionality to create `/etc/defaults/devolutions-gateway` file.~~ Post-publication edit: References to the defaults file are also removed.

Lastly, the PR adds an `rm -rf /etc/devolutions-gateway` command if `postrm` is called with `purge` flag.

Post-publication edit: PR also changes the systemd reload command order and also enables (and starts) `devolutions-gateway` service to start on boot.

### DIFF
--- a/package/Linux/gateway/debian/postinst
+++ b/package/Linux/gateway/debian/postinst
@@ -1,8 +1,5 @@
 #!/bin/sh
 
-# System-wide package configuration.
-DEFAULTS_FILE="/etc/default/devolutions-gateway"
-
 if [ ! -d /etc/devolutions-gateway ]; then
 	/bin/mkdir /etc/devolutions-gateway
 	/bin/chmod 655 /etc/devolutions-gateway
@@ -10,7 +7,7 @@ fi
 
 if [ -d /run/systemd/system ]; then
 	/usr/bin/devolutions-gateway service register >/dev/null
-	/usr/bin/devolutions-gateway service --config-init-only >/dev/null
-	systemctl restart devolutions-gateway >/dev/null 2>&1
 	systemctl daemon-reload
+	/usr/bin/devolutions-gateway --config-init-only >/dev/null
+	systemctl enable --now devolutions-gateway >/dev/null 2>&1
 fi
--- a/package/Linux/gateway/debian/postrm
+++ b/package/Linux/gateway/debian/postrm
@@ -9,9 +9,6 @@ if [ "$action" != "purge" ] ; then
   exit 0
 fi
 
-# System-wide package configuration.
-DEFAULTS_FILE="/etc/default/devolutions-gateway"
-
-if [ -s "$DEFAULTS_FILE" ]; then
-  rm "$DEFAULTS_FILE" || exit 1
+if [ -d /etc/devolutions-gateway ]; then
+  rm -rf /etc/devolutions-gateway
 fi
--- a/package/Linux/gateway/debian/prerm
+++ b/package/Linux/gateway/debian/prerm
@@ -8,7 +8,7 @@ if [ "$2" = "in-favour" ]; then
   action="upgrade"
 fi
 
-# Don't clean-up just for an upgrade.`
+# Don't clean-up just for an upgrade.
 if [ "$action" = "upgrade" ] ; then
   exit 0
 fi
@@ -16,4 +16,5 @@ fi
 if [ -d /run/systemd/system ]; then
   systemctl stop devolutions-gateway >/dev/null 2>&1
   /usr/bin/devolutions-gateway service unregister >/dev/null
+  systemctl daemon-reload
 fi


awakecoding:
Just to clarify, devolutions-agent is a different service, we haven't renamed devolutions-gateway to devolutions-agent. This being said, we may have accidentally renamed the service. @allan2 can you follow up on this?

package/Linux/gateway/debian/postinst
```diff
@@ -3,14 +3,18 @@
 # System-wide package configuration.
 DEFAULTS_FILE="/etc/default/devolutions-gateway"
 
+if [ ! -f $DEFAULTS_FILE ]; then
+	/bin/touch $DEFAULTS_FILE
+fi
```
CBenoit, line 8: > And although not utilized by the service file, this PR also adds the functionality to create `/etc/defaults/devolutions-gateway` file.

This is likely something that was overlooked in the past.
I don’t think the program is ever reading the file `/etc/default/devolutions-gateway`.
The system-wide configuration is found at `/etc/devolutions-gateway/gateway.json`, and is created by the command `devolutions-gateway --config-init-only`.

package/Linux/gateway/debian/postinst
```diff
@@ -3,14 +3,18 @@
...
 	/bin/chmod 655 /etc/devolutions-gateway
 fi
 
 if [ -d /run/systemd/system ]; then
 	/usr/bin/devolutions-gateway service register >/dev/null
-	/usr/bin/devolutions-gateway service --config-init-only >/dev/null
-	systemctl restart devolutions-gateway >/dev/null 2>&1
 	systemctl daemon-reload
+	/usr/bin/devolutions-gateway --config-init-only >/dev/null
+	systemctl restart devolutions-agent >/dev/null 2>&1
```
CBenoit, line 19: This is a pertinent improvement, the correct command is `/usr/bin/devolutions-gateway --config-init-only`.
(But we should not be using the name `devolutions-agent`, the fix is not to change that here.)

package/Linux/gateway/debian/postrm
```diff
@@ -12,6 +12,10 @@ fi
...
 
-if [ -s "$DEFAULTS_FILE" ]; then
-  rm "$DEFAULTS_FILE" || exit 1
+if [ -s $DEFAULTS_FILE ]; then
+  rm $DEFAULTS_FILE || exit 1
+fi
+
+if [ -d /etc/devolutions-gateway ]; then
+  rm -rf /etc/devolutions-gateway
 fi
```
CBenoit, line 21: To check.
Removing a package using `apt-get remove` should not purge the configuration. Configuration files found in `/etc`, are controlled by the system administrator, and could contain a significant amount of human effort.
However, `apt-get purge` should result in the configuration files to be deleted as well.

package/Linux/gateway/debian/prerm
```diff
@@ -8,12 +8,13 @@ if [ "$2" = "in-favour" ]; then
...
+# Don't clean-up just for an upgrade.
 if [ "$action" = "upgrade" ] ; then
   exit 0
 fi
 
 if [ -d /run/systemd/system ]; then
-  systemctl stop devolutions-gateway >/dev/null 2>&1
+  systemctl stop devolutions-agent >/dev/null 2>&1
   /usr/bin/devolutions-gateway service unregister >/dev/null
+  systemctl daemon-reload
```
CBenoit, line 19: Reloading at this point sounds reasonable. :+1:

package/Linux/gateway/debian/postinst
CBenoit, line 8: You mention that too in your original post, but I’m not sure we want to start creating this file at all given that we do not provide any option for configuring the service. We’re probably better off deleting the line above declaring `DEFAULTS_FILE`.

linuxgemini:
Heya @CBenoit !

Per your and @awakecoding's notes, I've removed the service renaming as its a bug (also renamed #1223 to reflect this).

I've also updated the code per your reviews. Thank you!

CBenoit (approved):
Everything is looking good to me. To my knowledge this is a net improvement, and I don’t see any issue. Approving and merging now to include in the next build.

@allan2 When you are back, please, double check in case I missed something. Otherwise, I think it would be good to apply a similar patch to the deb package for the agent too. Also, as discussed on Slack, it would be great to have a few sanity checks/smoke tests run on our Linux packages in the CI to prevent similar regressions in the future.

@linuxgemini Thank you for contributing this!

CBenoit:
@linuxgemini CI is currently broken for forks. This is fixed by https://github.com/Devolutions/devolutions-gateway/pull/1228

Do you think you could rebase your PR once https://github.com/Devolutions/devolutions-gateway/pull/1228 is merged?

Thank you!

linuxgemini:
@CBenoit Sure, I'll rebase when merged.

PR #1231, opened 2025-02-18
build: update Debian maintainer scripts

allan2:
This commit uses less overrides when building Linux packages with dpkg-buildpackage.

- uses a debian/install file to copy files into the package
- uses dh_installsystemd (which is on by default and built into dpkg-buildpackage)
  - removes prerm and postrm custom scripts in favour of the generated by dh_installsystemd, which handle systemd service setup and teardown
  - custom postinst script is retained, with the dh_installsystemd part pasted in
- moves previous postinst, prerm, postrm scripts into package/prod/rpm directory
  - fpm does not handle systemd like dh_installsystemd so we still pass these scripts in to fpm like before (https://github.com/jordansissel/fpm/issues/1163) 

- upgrades debhelper from 10 to 13
  - compat version is now specified in debian/control instead of a generated debian/compat file
  
Since the final scripts are not version-controlled, I have attached scripts extracted from this branch's Gateway build below. (GitHub required the .txt extension)

[postinst.txt](https://github.com/user-attachments/files/18855708/postinst.txt)
[postrm.txt](https://github.com/user-attachments/files/18855709/postrm.txt)
[prerm.txt](https://github.com/user-attachments/files/18855710/prerm.txt)


package/Linux/gateway/rpm/postinst
```diff
@@ -0,0 +1,14 @@
...
+	/bin/mkdir /etc/devolutions-gateway
+	/bin/chmod 655 /etc/devolutions-gateway
+fi
+
+if [ -d /run/systemd/system ]; then
+	/usr/bin/devolutions-gateway service register >/dev/null
+	systemctl daemon-reload
+	/usr/bin/devolutions-gateway --config-init-only >/dev/null
+	systemctl enable --now devolutions-gateway >/dev/null 2>&1
+    systemctl restart devolutions-gateway >/dev/null 2>&1
```
CBenoit, line 13:
```suggestion
   systemctl restart devolutions-gateway >/dev/null 2>&1
```

CBenoit (approved):
LGTM! Thank you.

PR #1233, opened 2025-02-19
ci(release): filter out agent Linux artifacts for GitHub releases

CBenoit:
Devolutions Agent for Linux is not ready yet.


awakecoding (approved):

CBenoit:
> Just to clarify, this removes the individual agent artifacts for linux but not the packages? We should still build and publish the packages even if not useful at this point. However, the executable alone will likely always be useless, only the package really matters

This modification only affects the artifacts attached to the GitHub release. The "official publishing" via OneDrive is untouched, and the artifacts are simply hidden from our website at this time.

PR #1235, opened 2025-02-19
feat(installer): make agent updater a selectable feature

thenextman:
Make the Gateway updater a selectable feature in the Agent MSI installer.

New installs will select the feature by default; upgrades will follow what is already installed. Note that since this a "new feature", upgrades from prior versions will _not_ automatically select the updater feature. I choose not to code around this because we don't have a public release yet.

I reworked the `StockIcon` helper but ended up not using the changes. However, I keep the improvements that may be useful in future.


CBenoit (approved):
LGTM!

PR #1237, opened 2025-02-20
fix(pedm): don't use Windows file namespace in paths

thenextman:
I've noticed that nearly every .NET application I elevate with PEDM, crashes (sometimes immediately, sometimes after running for a while).

The crashes seem centred around paths (for example, IISCrypto crashes on startup complaining of a invalid path; RDM crashes after some time when a third party component tries to interact with the shell by passing a path to the application, etc)

When elevating an executable, PEDM first expands environment variables in the executable path and then calls `fs::canonicalize` which has a side-effect of prefixing the canonical path with the Windows file namespace (`\\?\`).

It turns out, that not a single .NET application that I tested works when launched from a path with a Windows file namespace prefix.

This PR uses `dunce` in the same was as Devolutions Gateway to workaround the issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1255,6 +1255,7 @@ dependencies = [
  "devolutions-gateway-task",
  "devolutions-pedm-shared",
  "digest",
+ "dunce",
  "hyper 1.5.0",
  "hyper-util",
  "parking_lot",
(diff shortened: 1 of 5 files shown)


awakecoding (approved):

awakecoding:
Good catch!

PR #1240, opened 2025-02-21
build: split tlk.ps1 into separate scripts

allan2:
This commit splits _tlk.ps1_. It strives to make certain parts easier to run and to make the building and packaging process more readable.

The documentation of how to use the new scripts is in [_ci/README.md_](https://github.com/Devolutions/devolutions-gateway/blob/e98ef1a3368cffc352d9cae0183051adabd31945/ci/README.md). The existing _tlk.ps1_ is untouched as this PR does not include integration with GitHub Actions.

- usage is more strict; unsupported configurations (building on supported platforms, unused arguments) will throw an error
- _build.ps1_ now uses the `target/profile` build by default, without having to build into a separate target/arch folder
- architecture is an optional parameter for _build.ps1_, detecting your system architecture if not specified
- output directory structure now uses the full target triplet
  - this is to enable building for additional targets in the future such as musl
- removed `cargo test` functionality
- removed unused verbosed flag in `TlkRecipe`
- removed unused `Get-DotEnvFile` function
- removed DGATEWAY_PSMODULE_CLEAN functionality
- fail with `Stop` instead of `SilentlyContinue` when copying files
- use of named parameters for the PS scripts instead of environment variables
  - this makes it clear up-front what is required when building or packaging a given product
  - in the future, I would like to pass more parameters to `MSBuild` as properties instead of through environment variables
- functions are named using PowerShell-approved verbs
- adds documentation

This commit contains all scripts for building on all platforms and packaging on Windows.


awakecoding:
Here's the current plan:
- merge Windows tlk.ps1 split
- merge Linux tlk.ps1 split
- make use of new scripts in CI workflows, then remove old tlk.ps1 once everything is working fine

awakecoding (approved):

PR #1241, opened 2025-02-21
build: target Ubuntu 18.04

awakecoding:
Properly target ubuntu-18.04, including for the cadeau library which was just released with ubuntu-18.04 targeting as well. By targeting ubuntu-18.04, we are forward-compatible with ubuntu 20.04, 22.04, 24.04 but we are also finally compatible with RHEL8, which uses a version of glibc older than ubuntu 20.04.


thenextman (approved):
Excellent!!!

CBenoit (approved):

PR #1242, opened 2025-02-21
build: fix build.ps1 param parsing

allan2:
`$Target` being optional was causing a problem when running the script file like `build.ps1 agent`.

This adds a default `infer` value so that it can be omitted.


CBenoit (approved):
LGTM

PR #1244, opened 2025-02-24
refactor(pedm): refactor sid and acl modules

CBenoit:
Issue: DGW-223
Issue: DGW-239

### DIFF
--- a/crates/win-api-wrappers/src/token.rs
+++ b/crates/win-api-wrappers/src/token.rs
@@ -859,7 +859,7 @@ trait FromWin32<Win32Ty>: Sized {
     ///
     /// # Safety
     ///
-    /// The pointed Win32 struct must be a valid.
+    /// The pointed Win32 struct must be valid.
     ///
     /// Validity, in this context, means that it is possible to rely on the
     /// fields of the Win32 struct being initialized in a way that would not


crates/win-api-wrappers/src/identity/account.rs
```diff
@@ -213,3 +265,18 @@ impl Drop for ProfileInfo {
...
+
+    use crate::str::u16cstr;
+    use crate::token::Token;
+
+    #[test]
+    fn get_virtual_account_name() {
+        let token = Token::current_process_token();
+        let account_name = virtual_account_name(&token).unwrap();
+        assert_eq!(account_name, u16cstr!(""))
+    }
```
CBenoit, line 281: This fails with
```
The handle is invalid. (0x80070006)
```

Which makes sense given that `Token::current_process_token()` returns a pseudo-handle.

crates/win-api-wrappers/src/security/acl.rs
```diff
@@ -1,464 +1,265 @@
...
+        // The Windows heap managers have always performed heap allocations with a start address that is 8-byte aligned.
+        // (On 64-bit platforms the alignment is 16-bytes).
+        // Microsoft never documented it as a formal guarantee, but given its multi-decade consistency in practice,
+        // it’s essentially become a stable implementation detail.
+
+        // SAFETY: FFI call with no outstanding precondition.
+        let ptr = unsafe { Memory::LocalAlloc(Memory::LMEM_ZEROINIT, size_of::<Security::ACL>())? };
+
+        // SAFETY: The buffer is u32-aligned (= DWORD-aligned), since both 8-byte and 16-byte alignments are stricter.
+        unsafe { Security::InitializeAcl(ptr.0.cast(), u32size_of::<Security::ACL>(), Security::ACL_REVISION)? };
```
CBenoit, line 46: Does that sound good to you? @thenextman

crates/win-api-wrappers/src/security/acl.rs
```diff
@@ -1,464 +1,265 @@
...
 
-        Ok(match u32::from(kind) {
-            ACCESS_ALLOWED_ACE_TYPE => Self::AccessAllowed(Sid::try_from(raw_sid)?),
-            _ => bail!(Error::from_win32(ERROR_INVALID_VARIANT)),
+        Ok(Self {
+            // ACL structure initialized using InitializeAcl.
```
CBenoit, line 49:
```suggestion
            // ACL structure is properly initialized using InitializeAcl.
```

crates/devolutions-pedm/src/api/mod.rs
```diff
@@ -91,23 +91,23 @@ async fn named_pipe_middleware(
...
-            data: AceType::AccessAllowed(Token::current_process_token().sid_and_attributes()?.sid),
+        ExplicitAccess {
+            access_permissions: PIPE_ACCESS_FULL_CONTROL,
+            access_mode: GRANT_ACCESS,
+            inheritance: NO_INHERITANCE,
+            trustee: Trustee::Sid(Token::current_process_token().sid_and_attributes()?.sid),
         },
-    ])
-    .to_raw()?;
+    ])?;
```
CBenoit, line 107: @thenextman I moved away from building and adding ACE structures to the ACL, and instead `SetEntriesInAclW` is used along `EXPLICIT_ACCESS_W` structures. It seems to be equivalent.

crates/devolutions-pedm/src/elevator/virtual_account_elevator.rs
```diff
@@ -51,45 +53,49 @@ impl VirtualAccountElevator {
         }
     }
 
-    fn create_token(&self, token: &Token) -> Result<()> {
-        let virtual_account = create_virtual_account(self.rid, &self.domain, token)?;
+    fn create_token(&self, token: &Token) -> anyhow::Result<()> {
```
CBenoit, line 56: thought: I’m not sure the name is very good, as "create_token" suggests the output of the function would be a `Token`, but it’s not. Seems to be more about "upgrading" or "elevating" an existing token?

crates/devolutions-pedm/src/utils.rs
```diff
@@ -115,6 +116,7 @@ pub(crate) fn file_hash(path: &Path) -> anyhow::Result<Hash> {
 
 pub(crate) fn ensure_protected_directory(dir: &Path, _readers: Vec<Sid>) -> anyhow::Result<()> {
     // FIXME: Underlying behaviour of security primitives must be corrected before this can work
```
CBenoit, line 118: question: Is it something important we need to uncomment for the V1?

crates/win-api-wrappers/src/identity/account.rs
```diff
@@ -213,3 +265,19 @@ impl Drop for ProfileInfo {
...
+
+    use crate::str::u16cstr;
+    use crate::token::Token;
+
+    #[test]
+    #[cfg_attr(miri, ignore)]
+    fn get_virtual_account_name() {
+        let token = Token::current_process_token();
+        let account_name = virtual_account_name(&token).unwrap();
+        assert_eq!(account_name, u16cstr!(""))
```
CBenoit, line 281:
```suggestion
        assert_eq!(account_name, u16cstr!("TODO"))
```

crates/win-api-wrappers/src/str.rs
CBenoit, line 1: Please, use this module for UTF-16 string manipulations. The `WideString` type is flawed.

crates/win-api-wrappers/src/identity/account.rs
CBenoit: Modified into the following code that is working as expected:

```rust
// Retrieve a non-pseudo-token with TOKEN_QUERY | TOKEN_DUPLICATE | TOKEN_IMPERSONATE access rights.
// Without these access rights, Windows will return an "Access is denied." error.
let current_process = Process::current_process();
let token = current_process
    .token(Security::TOKEN_QUERY | Security::TOKEN_DUPLICATE | Security::TOKEN_IMPERSONATE)
    .unwrap();

let account_name = virtual_account_name(&token).unwrap();
let account_name = account_name.as_ucstr().to_string_lossy();

// Check that the UTF-16 substring substition logic is working as expected.
assert!(account_name.contains('_'));
assert!(!account_name.contains('\\'));
```

crates/win-api-wrappers/src/netmgmt.rs
```diff
@@ -1,49 +1,30 @@
...
-
-    if rc != NERR_Success {
-        bail!(Error::from_win32(WIN32_ERROR(rc)))
-    }
+    let group_name = U16CString::from_str(group_name).context("invalid group name")?;
 
-    Ok(())
-}
-
-pub fn remove_local_group_member(group_name: &str, security_identifier: &Sid) -> anyhow::Result<()> {
```
thenextman, line 33: Why remove this function? It's not currently used, but we would like to use that in future

If it's the case that we want to strip unused code for now; this whole module is technically not used by the application currently

crates/devolutions-pedm/src/api/mod.rs
thenextman: Excellent!

crates/devolutions-pedm/src/utils.rs
thenextman: Ideally we would uncomment this. I will try it on my side as this was a source of issues.

We are setting directory permissions in the installer, but I'd consider it safer to also use this function and enforce permissions on access.

crates/devolutions-pedm/src/elevator/virtual_account_elevator.rs
thenextman: The naming is bad; but it is creating a token. Let's leave for now as you suggest, because I don't have a better idea

crates/win-api-wrappers/src/security/acl.rs
thenextman: It matches my understanding, from the documentation 👍

thenextman (approved):

PR #1249, opened 2025-02-25
fix(installer): don't fail an uninstall if the shell extension can't be unregistered

thenextman:
An unexpected error unregistering the PEDM shell extension can cause an uninstall to fail; this leads to a bad posture on the user machine.

cc: @allan2


CBenoit (approved):
LGTM!

PR #1255, opened 2025-02-27
build: fix Agent MSI

allan2:
This builds a working Agent MSI.

- fixes MSI internal file names that were breaking install
  - PedmShellExt.dll must be named exactly that
- fixes inclusion of DesktopAgent
  - previously was only inlcuding source files
- adds MSBuild logs (excluded from version control)
- `OutputDir` changed to `Outfile` to allow custom output name
  - for example, one can now build an MSI with a timestamped filename


CBenoit (approved):
Thank you!

PR #1258, opened 2025-03-03
ci: fix configuration of arm64 linux runner

thenextman:
`apt` is failing to install packages for arm64; however as per @awakecoding we actually don't need these packages anymore since everything we want is inside our custom sysroot.


CBenoit (approved):
LGTM, thank you!

PR #1268, opened 2025-03-11
feat(dgw): add --config-path command-line option

awakecoding:
https://devolutions.atlassian.net/browse/ARC-307

image

### DIFF
--- a/devolutions-gateway/src/main.rs
+++ b/devolutions-gateway/src/main.rs
@@ -49,10 +49,32 @@ enum CliAction {
 
 fn main() -> anyhow::Result<()> {
     let mut args = std::env::args();
-
     let executable = args.next().context("executable name is missing from the environment")?;
 
-    let action = match args.next().as_deref() {
+    // Extract and remove --config-path argument if provided
+    let mut config_path: Option<String> = None;
+    let mut remaining_args = Vec::new();
+
+    while let Some(arg) = args.next() {
+        if arg == "--config-path" {
+            if let Some(path) = args.next() {
+                config_path = Some(path);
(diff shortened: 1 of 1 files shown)


devolutions-gateway/src/main.rs
```diff
@@ -49,10 +49,31 @@ enum CliAction {
...
+                return Err(anyhow::anyhow!("Missing value for --config-path"));
+            }
+        } else {
+            remaining_args.push(arg);
+        }
+    }
+
+    // Set the DGATEWAY_CONFIG_PATH if --config-path was provided
+    if let Some(path) = config_path {
+        std::env::set_var("DGATEWAY_CONFIG_PATH", &path);
```
CBenoit, line 72: note: This is an unsafe function in the 2024 edition, so I think we should refactor that out as a follow up before we migrate to the 2024 edition.

devolutions-gateway/src/main.rs
```diff
@@ -49,10 +49,31 @@ enum CliAction {
...
+    // Extract and remove --config-path argument if provided
+    let mut config_path = None;
+    let mut remaining_args = Vec::new();
+
+    while let Some(arg) = args.next() {
+        if arg == "--config-path" {
+            if let Some(path) = args.next() {
+                config_path = Some(path);
+            } else {
+                return Err(anyhow::anyhow!("Missing value for --config-path"));
```
CBenoit, line 63:
```suggestion
                return Err(anyhow::anyhow!("missing value for --config-path"));
```

CBenoit (approved):
LGTM!

PR #1270, opened 2025-03-13
refactor(dgw): rewrite recording-player-tester

irvingoujAtDevolution:
Rewrite recording-player-tester in React (TypeScript + TailwindCSS)
The old project became unmanageable, so I'm starting fresh with a simpler stack.
This is purely for testing, so React and TailwindCSS ensure faster development and easier maintenance.
Removed the old Node.js dummy server—now relying on the Gateway and token server for real scenario testing.


CBenoit (commented):
Thank you for this refactoring. It seems we have much more dependencies now, but it’s worth it if things are more maintainable. I’ll leave @kristahouse have a look too.

kristahouse (approved):
Nice. LGTM.

PR #1278, opened 2025-03-18
fix(installer): resolve potential null reference installing PEDM shell extension

thenextman:
The top-level file extension keys may not all exist (in the case of QA, `HKEY_CLASSES_ROOT\\.ps1` was not present). Although we check for an empty default _value_, we don't check for the presence of the top-level key which could cause a null-reference exception and force the installer to roll back.


CBenoit (approved):
Good catch

PR #1288, opened 2025-03-26
refactor(pedm): cleanup and clippy

allan2:
This resolves all clippy errors in the the devolutions-pedm crate.

### DIFF
--- a/crates/devolutions-pedm/src/api/elevate_session.rs
+++ b/crates/devolutions-pedm/src/api/elevate_session.rs
@@ -6,9 +6,7 @@ use crate::{elevations, policy};
 
 use super::NamedPipeConnectInfo;
 
-pub async fn post_elevate_session(
-    Extension(named_pipe_info): Extension<NamedPipeConnectInfo>,
-) -> Result<(), Error> {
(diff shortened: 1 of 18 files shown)


crates/devolutions-pedm/src/api/elevate_session.rs
```diff
@@ -6,7 +6,7 @@ use crate::{elevations, policy};
 
 use super::NamedPipeConnectInfo;
 
-pub(crate) async fn post_elevate_session(
+pub async fn post_elevate_session(
```
CBenoit, line 9: question: Why did you make everything public? Is it API we actually intended to expose from this crate? I believe not. I would rather not increase the API surface of any crate more than necessary.

CBenoit: I understand where you come from, but the problem is that there is no dead_code lint across crate boundaries. We are much more likely to keep around dead code if it’s publicly exposed from the crate. I also tend to prefer clear `pub(crate)` over `pub` when the item is never used outside of the crate, just so I have a clear idea on what is an API boundary (even if we are the only ones to use it).
That being said, the workspace unused pub items problem could be addressed with a tool such as https://github.com/cpg314/cargo-workspace-unused-pub
If we integrate it in our CI, this may become mostly a non issue.

CBenoit (approved):
LGTM!

PR #1289, opened 2025-03-26
ci: switch to nightly for cargo fmt

allan2:
I encountered a warning in the "Check formatting" step of the preflight job.

```
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
```

Switching to nightly resolves this issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,9 +108,9 @@ jobs:
 
       - name: Check formatting
         run: |
-          cargo fmt --all -- --check
+          cargo +nightly fmt --all -- --check
           if ! [ $? -eq 0 ] ; then
-              echo "::error::Bad formatting, please run 'cargo +stable fmt --all'"
+              echo "::error::Bad formatting, please run 'cargo +nightly fmt --all'"
               exit 1
           fi
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,9 @@ jobs:
           $Version = Get-Content VERSION -TotalCount 1
           echo "version=$Version" >> $Env:GITHUB_OUTPUT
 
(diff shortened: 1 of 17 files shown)


CBenoit:
> I encountered a warning in the "Check formatting" step of the preflight job.
> 
> ```
> Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
> ```
> 
> Switching to nightly resolves this issue.

Just to clarify here.

[The problem in your other PR](https://github.com/Devolutions/devolutions-gateway/actions/runs/14087621362/job/39455740362) is that you actually have some piece of code that is not properly formatted according to the _stable_ formater:

image

See what comes next to "Diff in…". The warning about the nightly option being ignored, is merely a warning, not an "issue" causing a CI failure. Switching to nightly will not "fix" this.

The original thinking for enabling this nightly option of rustfmt without enforcing in the CI:

- Do not enforce a nightly toolchain on the developers
- Benefit from the import granularity nightly feature, that is automatically organizing our imports such that std, external and internal imports are clearly separated.

It’s a "nice-to-have" dev QOL thing, more than anything else.

All that said, today, I think it’s fine to enforce that our developers run a nightly rustfmt.
Installing any toolchain is really straightforward with rustup, and it’s merely just formatting the code, it does not affect whatsoever what is shipped to the customer.

The small problem I think we may encounter with this change, is that nightly rustfmt may break our CI at some random timings (bugs may be introduced, changes may be introduced into nightly options, etc). But as of today, I am fairly confident that this would be a rare enough occurrence, especially given that rustfmt test suite was likely much improved, and we don’t use many nightly options in the first place.

CBenoit (approved):
Approving! You’ll need to run `cargo +nightly fmt --all` to reformat existing code before this can be merged.

PR #1290, opened 2025-03-31
fix(dgw): improve network interface information extraction

irvingoujAtDevolution:
Improved api/net/config end point, correctly return ip and routes for both windows and linux. Renamed struct field to better match it's semantices.

### Example output

From api request
```pwsh
id          : {B4488786-D294-448F-8D0A-D2573F2126C7}
description : Hyper-V Virtual Ethernet Adapter #3
mac_address : 00:15:5D:34:6E:07
routes      : {@{ip=10.10.0.0; prefixlen=24}, @{ip=10.10.0.1; prefixlen=32}, @{ip=10.10.0.255; 
              prefixlen=32}, @{ip=224.0.0.0; prefixlen=4}…}
is_up       : True
gateways    : {}
nameservers : {fec0:0:0:ffff::1, fec0:0:0:ffff::2, fec0:0:0:ffff::3}
name        : vEthernet (LAN Switch)
ip_adresses : {10.10.0.1}
```
from ipconfig
```
Ethernet adapter vEthernet (LAN Switch):

   Connection-specific DNS Suffix  . : 
   Link-local IPv6 Address . . . . . : fe80::f5f9:2d83:9342:2005%23
   IPv4 Address. . . . . . . . . . . : 10.10.0.1
   Subnet Mask . . . . . . . . . . . : 255.255.255.0
   Default Gateway . . . . . . . . . : 
```

### DIFF
--- a/crates/network-scanner/src/interfaces/filter.rs
+++ b/crates/network-scanner/src/interfaces/filter.rs
@@ -1,0 +1,42 @@
+use std::net::IpAddr;
+
+use super::NetworkInterface;
+
+#[derive(Debug, Clone)]
+pub struct Filter {
+    pub ignore_ipv6: bool,
+    pub include_loopback: bool,
+}
+
+impl Default for Filter {
+    fn default() -> Self {
+        Self {
+            ignore_ipv6: true,
+            include_loopback: false,
+        }
+    }
+}
+
+impl Filter {
+    pub fn matches(&self, interface: &NetworkInterface) -> bool {
+        // 1. Loopback exclusion
+        if !self.include_loopback {
+            for addr in &interface.addresses {
+                if let IpAddr::V4(ipv4) = addr.ip {
+                    if ipv4.octets()[0] == 127 {
+                        return false;
+                    }
+                }
+            }
+        }
+
+        true
+    }
+
+    pub fn clean(&self, interfaces: &mut NetworkInterface) {
+        if self.ignore_ipv6 {
+            interfaces.addresses.retain(|addr| matches!(addr.ip, IpAddr::V4(_)));
+            interfaces.ip_adresses.retain(|addr| matches!(addr, IpAddr::V4(_)));
+        }
+    }
+}
--- a/crates/network-scanner/src/interfaces/filter.rs
+++ b/crates/network-scanner/src/interfaces/filter.rs
@@ -21,7 +21,7 @@ impl Filter {
     pub fn matches(&self, interface: &NetworkInterface) -> bool {
         // 1. Loopback exclusion
         if !self.include_loopback {
-            for addr in &interface.addresses {
+            for addr in &interface.routes {
                 if let IpAddr::V4(ipv4) = addr.ip {
                     if ipv4.octets()[0] == 127 {
                         return false;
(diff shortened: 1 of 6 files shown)


crates/network-scanner/src/interfaces/filter.rs
```diff
@@ -0,0 +1,42 @@
...
+        Self {
+            ignore_ipv6: true,
+            include_loopback: false,
+        }
+    }
+}
+
+impl Filter {
+    pub fn matches(&self, interface: &NetworkInterface) -> bool {
+        // 1. Loopback exclusion
```
CBenoit, line 22: nitpick: I would prefer we not enumerate comments with `1.` and so on. This kind of comment becomes outdated easily, and there is not even a `2.`.

crates/network-scanner/src/interfaces/filter.rs
```diff
@@ -0,0 +1,42 @@
...
+
+        true
+    }
+
+    pub fn clean(&self, interfaces: &mut NetworkInterface) {
+        if self.ignore_ipv6 {
+            interfaces.routes.retain(|addr| matches!(addr.ip, IpAddr::V4(_)));
+            interfaces.ip_adresses.retain(|addr| matches!(addr, IpAddr::V4(_)));
+        }
+    }
```
CBenoit, line 41: naming: `apply`? As in "applying a filter"

crates/network-scanner/src/interfaces/filter.rs
```diff
@@ -0,0 +1,42 @@
+use std::net::IpAddr;
+
+use super::NetworkInterface;
+
+#[derive(Debug, Clone)]
+pub struct Filter {
+    pub ignore_ipv6: bool,
+    pub include_loopback: bool,
+}
```
CBenoit, line 9: issue: I think naming in this module should be improved. Or rather, it seems that the `Filter` struct is trying to achieve two different goals at once, and that’s why it resulted into a confusing naming. For instance, I didn’t expect `Filter::matches` to be used specifically for excluding loopback interfaces, and `Filter::clear` should be named something else, because `clear` is typically used to empty collections, and at least it’s not clear what it means for a "filter".

It may be simpler to use two free-standing functions instead.

Or maybe you want to keep `Filter` to use to pass the filter parameters to `get_network_interface`, but extract the parameters and pass them to functions instead of declaring the operations on `Filter`.

crates/network-scanner/src/interfaces/linux.rs
```diff
@@ -250,31 +242,14 @@ impl TryFrom<RouteMessage> for RouteInfo {
...
-        let prefix_len = value.header.prefix_len;
-
-        let address_info = AddressInfo {
-            address: *addr,
-            prefix_len,
-        };
-
-        Ok(address_info)
-    }
+fn try_to_ip(addr_msg: AddressMessage) -> Option<IpAddr> {
```
CBenoit, line 245: naming: `extract_ip`?

crates/network-scanner/src/interfaces/mod.rs
```diff
@@ -72,10 +88,13 @@ pub struct InterfaceAddress {
 
 #[derive(Debug, Clone)]
 pub struct NetworkInterface {
+    // Linux interfaces does not come with an id
+    pub id: Option<String>,
```
CBenoit, line 92: question: Should it be named `adapter_id` or `adapter_name` instead? Or is this naming used somewhere else?

crates/network-scanner/src/interfaces/filter.rs
irvingoujAtDevolution: updated

crates/network-scanner/src/interfaces/linux.rs
irvingoujAtDevolution: updated

CBenoit:
@irvingoujAtDevolution CI is red, can you check that?

CBenoit:
@irvingoujAtDevolution I think you can rebase on master. The problem appears to be fixed.

PR #1293, opened 2025-04-01
build(deps): bump libsql to 0.9

allan2:
0.9.x is still broken on Windows, but I will make a patch fix where needed, i.e., PEDM PR.

### DIFF
--- a/crates/job-queue-libsql/Cargo.toml
+++ b/crates/job-queue-libsql/Cargo.toml
@@ -19,6 +19,3 @@ tracing = "0.1"
 typed-builder = "0.19"
 ulid = { version = "1.1", features = ["uuid"] }
 uuid = "1.11"
-
-[patch.crates-io]
-libsql = { git = "https://github.com/allan2/libsql", branch = "ffi-windows-cp" }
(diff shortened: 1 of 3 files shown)


awakecoding:
if 0.9.x is still broken in Windows, then why does this pull request uses 0.9 instead of pinned commit or branch?

crates/job-queue-libsql/Cargo.toml
```diff
@@ -12,10 +12,13 @@ workspace = true
...
+libsql = { version = "0.9", default-features = false, features = ["core", "serde", "replication"] }
 serde = "1"
 time = { version = "0.3", default-features = false, features = ["std"] }
 tracing = "0.1"
 typed-builder = "0.19"
 ulid = { version = "1.1", features = ["uuid"] }
 uuid = "1.11"
+
+[patch.crates-io]
+libsql = { git = "https://github.com/allan2/libsql", branch = "ffi-windows-cp" }
```
CBenoit, line 24: issue: dependency patches not found in the workspace Cargo.toml manifest are ignored. This should be moved to the toplevel Cargo.toml

suggestion: Also include the revision (commit hash).

CBenoit (approved):
LGTM!

PR #1296, opened 2025-04-02
ci: run PEDM simulator as LocalSystem using psexec

CBenoit:


CBenoit:
Catching a use-after-free bug in the success path of the elevation logic:

image

awakecoding (approved):

PR #1301, opened 2025-04-08
refactor(pedm): modify `serve` arguments

allan2:
- `serve` now takes a `Config` struct rather than a path
  - pipe name is now part of `Config` rather than a direct argument
- `Config` struct is now exported as `devolutions_pedm::Config`
- `serve` is now exported as `devolutions_pedm::serve`

### DIFF
--- a/crates/devolutions-pedm/src/config.rs
+++ b/crates/devolutions-pedm/src/config.rs
@@ -7,32 +7,46 @@ use tracing::info;
 
 use crate::data_dir;
 
+/// Specifies the default pipe name.
+///
+/// This is a workaround for `serde(default)` not taking a raw string literal or escaped backslashes.
+fn default_pipe_name() -> String {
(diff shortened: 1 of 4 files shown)


crates/devolutions-pedm/src/config.rs
```diff
@@ -7,32 +7,46 @@ use tracing::info;
...
+    /// Backslashes must be escaped, like "\\\\.\\pipe\\foo".
+    /// This field is intentionally omitted from the example configuration.
+    #[serde(default = "default_pipe_name")]
+    pub pipe_name: String,
 }
 
 impl Config {
     /// Creates a new config with the default values for a new setup.
-    fn standard() -> Self {
+    pub fn standard() -> Self {
```
CBenoit, line 37: note: Sounds like it should be `Default::default()`

CBenoit (approved):
LGTM! It’s better that way. Thank you!

PR #1302, opened 2025-04-08
feature(pedm): SQL initialization

allan2:
- adds version tracking in the `pedm_schema_version` table
- adds table initialization
  - error checking is strict and initialization only happens if the     version table is not found
- adds SQLite error handling for microsecond timestamp conversion to/from
  Chrono
- fixes a bug in the libSQL schema where datetimes were stored   incorrectly
  - they were being stored fractionally, instead of the full     microseconds since epoch
- adds `/about` endpoint to get basic info about the running application
  - I used this to ensure that the libsql date handling is functional

A few other changes are included:
- libsql feature set is changed (default is now false, "stream" removed)

This PR includes all commits from PR #1301, so an option is to merge this one only instead of #1301 first.

### DIFF
--- a/crates/devolutions-pedm/schema/libsql.sql
+++ b/crates/devolutions-pedm/schema/libsql.sql
@@ -36,4 +36,4 @@ CREATE TABLE IF NOT EXISTS elevate_tmp_request
     seconds integer NOT NULL
 );
 
-INSERT INTO version (version) VALUES (1) ON CONFLICT DO NOTHING;
+INSERT INTO version (version) VALUES (0) ON CONFLICT DO NOTHING;
--- a/crates/devolutions-pedm/src/db/mod.rs
+++ b/crates/devolutions-pedm/src/db/mod.rs
@@ -31,7 +31,7 @@ use tokio_postgres::config::SslMode;
 #[cfg(feature = "postgres")]
 use tokio_postgres::NoTls;
 
-pub(crate) const CURRENT_SCHEMA_VERSION: i16 = 1;
+pub(crate) const CURRENT_SCHEMA_VERSION: i16 = 0;
 
 /// A wrapper around the database connection.
 #[derive(Clone)]
--- a/crates/devolutions-pedm/schema/pg.sql
+++ b/crates/devolutions-pedm/schema/pg.sql
@@ -28,4 +28,4 @@ CREATE TABLE IF NOT EXISTS elevate_tmp_request
     seconds int NOT NULL
 );
 
-INSERT INTO version (version) VALUES (1) ON CONFLICT DO NOTHING;
+INSERT INTO version (version) VALUES (0) ON CONFLICT DO NOTHING;


crates/devolutions-pedm/schema/libsql.sql
CBenoit, line 1: question: Is there some rationale for when the `pedm` prefix is used in table names?

CBenoit: I think it will be best to use a different database completely! :slightly_smiling_face: 
I agree with the way you renamed the tables!

crates/devolutions-pedm/src/api/mod.rs
```diff
@@ -53,6 +47,14 @@ mod revoke;
 pub(crate) mod state;
 mod status;
 
+use about::about;
+use elevate_session::elevate_session;
+use elevate_temporary::elevate_temporary;
+use launch::post_launch;
+use revoke::post_revoke;
+use state::{AppState, AppStateError};
+use status::get_status;
```
CBenoit, line 56: style: I recommend to use `self::` to make it clear we are importing from the current module, but it’s not a big deal.

crates/devolutions-pedm/src/db/mod.rs
```diff
@@ -1,7 +1,12 @@
+use core::fmt;
+use std::ops::Deref;
+use std::sync::Arc;
+
 use async_trait::async_trait;
 
 mod err;
 
+use chrono::{DateTime, Utc};
```
CBenoit, line 9: style: Inconsistent import order.

crates/devolutions-pedm/src/db/mod.rs
```diff
@@ -14,22 +19,172 @@ mod pg;
...
 /// All queries required by the application are defined here. They must be implemented by each backend.
 #[async_trait]
 pub(crate) trait Database: Send + Sync {
+    /// Returns the schema version from the `version` table.
+    async fn get_schema_version(&self) -> Result<i16, DbError>;
+
+    /// Initializes the database schema.
+    ///
+    /// This creates tables.
+    async fn init_schema(&self) -> Result<(), DbError>;
```
CBenoit, line 170: question: Should we really expose the schema version to the user of the `Database` trait?

I think it’s more appropriate to define a single method called `setup` that is performing any backend-dependent setup sequence, including initializing the schema and/or performing migrations if necessary.

See how it’s done in `job-queue`/`job-queue-libsql` crates: https://github.com/Devolutions/devolutions-gateway/blob/cdbe0b7a00e0ced13c90fa4198eabe173b272140/crates/job-queue-libsql/src/lib.rs#L174-L178

CBenoit (commented):
thought: It seems the `/about` route is close to something that is typically called `/health`. I’m curious if you have further motivations for the `/about` route?

crates/devolutions-pedm/schema/libsql.sql
```diff
@@ -1,20 +1,39 @@
...
+    status_code integer NOT NULL
 );
+
+CREATE TABLE IF NOT EXISTS elevate_tmp_request
+(
+    req_id  integer PRIMARY KEY,
+    seconds integer NOT NULL
+);
+
+INSERT INTO version (version) VALUES (1) ON CONFLICT DO NOTHING;
```
CBenoit, line 39: question: Any objection to starting with `0`? I think it can be useful in the future for indexing purposes.

crates/devolutions-pedm/src/db/mod.rs
CBenoit: Fair enough. My thinking was to keep the exact setup logic as an implementation detail to leave more leeway, but at this point it does not really matter so we may as well avoid the logic duplication as you suggest. Certainly a better abstraction will emerge when it becomes necessary.

PR #1310, opened 2025-04-15
fix(pedm): write default config when not present

allan2:
The wrong function was being called when initializing the config. No default config was actually being written to disk.

This commit fixes it.


CBenoit (approved):
LGTM!

PR #1311, opened 2025-04-17
feat(dgw): stabilize /jet/net/config

irvingoujAtDevolution:
Stable enough now, it's just one function call in the `interfaces::get_network_interfaces`


CBenoit (approved):

CBenoit:
> Stable enough now, it's just one function call in the `interfaces::get_network_interfaces`

Note that the most important part when stabilizing a route is first and foremost the API. Behaviors can be fixed, but API cannot be changed easily.

PR #1325, opened 2025-05-08
feat(pedm): explicitly support elevating .msi files

thenextman:
By inspecting elevated .msi files launched from Explorer, we see that Explorer invokes %systemroot%\system32\msiexec, with the command line `"%systemroot%\system32\msiexec" /i "{path-to-msi}"`.

We achieve the same in the PEDM module by using the same command if the file extension is .msi. The .msi extension is already being trapped by the shell extension, but previously we would call `CreateProcess` on the .msi causing "file is not a valid Win32 executable".

Additionally, I polished and added @awakecoding's original "ExplorerCommand.ps1" script. This is useful for debugging, where the installed shell extension can't be overwritten as it's in use by Explorer. Install Devolutions Agent with the PEDM feature; then use the PowerShell to unregistered the shell extension. Swap in your development or debug DLL, then re-register the shell extension. If you supply the .pdb file as well it's possible to attach the debugger to Explorer.exe and debug the shell extension code.

### DIFF
--- a/crates/devolutions-pedm-shell-ext/src/lib_win.rs
+++ b/crates/devolutions-pedm-shell-ext/src/lib_win.rs
@@ -12,6 +12,7 @@ use devolutions_pedm_shared::client::{self};
 use devolutions_pedm_shared::desktop;
 use parking_lot::{Mutex, RwLock};
 use tokio::sync::mpsc::{self, Receiver, Sender};
+use win_api_wrappers::fs::get_system32_path;
 use win_api_wrappers::process::{Module, Process};
 use win_api_wrappers::raw::core::{
     implement, interface, Error, IUnknown, IUnknown_Vtbl, Interface, Result, GUID, HRESULT, PWSTR,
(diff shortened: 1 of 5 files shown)


awakecoding (approved):
wooohoo, you're making my day! good job!

crates/win-api-wrappers/src/security/crypt.rs
```diff
@@ -172,7 +172,7 @@ impl CatalogAdminContext {
         let mut required_size = 0u32;
 
         // SAFETY: `hFile` must not be NULL and must be a valid file pointer. The `file` is not dropped so it should be valid.
-        let res = unsafe {
+        let _ = unsafe {
```
CBenoit, line 175:
```suggestion
        unsafe {
```

crates/devolutions-pedm-shell-ext/src/lib_win.rs
```diff
@@ -253,13 +275,16 @@ fn start_listener() {
...
-                            let mut payload = resolve_lnk(&path).unwrap_or_else(|| LaunchPayload {
-                                executable_path: path.as_os_str().to_str().map(str::to_owned),
-                                command_line: None,
-                                working_directory: None,
-                                creation_flags: 0,
-                                startup_info: None,
-                            });
+                            let mut payload =
+                                resolve_lnk(&path)
+                                    .or_else(|| resolve_msi(&path))
```
CBenoit, line 280: question: So this is where the magic happen? Do you think you could include an inline comment documenting what you wrote in your PR message?

>By inspecting elevated .msi files launched from Explorer, we see that Explorer invokes %systemroot%\system32\msiexec, with the command line "%systemroot%\system32\msiexec" /i "{path-to-msi}".
>
>We achieve the same in the PEDM module by using the same command if the file extension is .msi. The .msi extension is already being trapped by the shell extension, but previously we would call CreateProcess on the .msi causing "file is not a valid Win32 executable".

Thanks!

CBenoit (approved):
LGTM! Left a few minor comments, feel free to address and merge! :slightly_smiling_face:

crates/devolutions-pedm-shell-ext/src/lib_win.rs
thenextman: Thanks to you!

PR #1327, opened 2025-05-09
refactor: async bridge for database operations

CBenoit:


crates/devolutions-pedm/src/db/mod.rs
```diff
@@ -197,4 +199,77 @@ pub(crate) trait Database: Send + Sync {
...
+                    "DB sync bridge is down, failed to insert the JIT elevation result in the database"
+                );
+            }
+        }
+    }
+}
+
+enum DbRequest {
+    InsertJitElevationResult(ElevationRequest),
+}
```
CBenoit, line 230: @thenextman Add a new variant here if you need the bridge to support more operations (config, etc).

thenextman (approved):
💯

PR #1329, opened 2025-05-13
feat(webapp): integrate the RDP and VNC WASM packages

RRRadicalEdward:
This PR changes the Devolutions Gateway's webapp to integrate the new modular _VNC_ and _RDP_ packages.
Key changes:

- Refactored the webapp to use the updated API from `iron-remote-desktop`
- Update the _VNC_ connection form with new specific settings: `Enabled Encodings`, `Disable Cursor`, etc.
- Update the _ARD_ connection form with `Resolutions Quality` and `Quality Mode` settings.
- Enhanced the viewer toolbar; added `Toggle Cursor Style` button and `Unicode Keyboard Mode` checkbox.

Also, I've tested integration, and all 3 protocols are working as expected.
Here are some demos:

https://github.com/user-attachments/assets/372b7bce-a327-4778-adb0-63e206850600

https://github.com/user-attachments/assets/03fe3298-c0b8-4691-b5ab-0776e79f3563

https://github.com/user-attachments/assets/ed7d3b68-b6c8-4b8a-a337-e7c84265dde4


webapp/src/client/app/modules/web-client/ssh/web-client-ssh.component.ts
```diff
@@ -70,7 +70,7 @@ export class WebClientSshComponent extends WebClientBaseComponent implements Web
   }
 
   ngOnInit(): void {
-    sshLoggingService.setLevel(LoggingLevel.DEBUG);
+    // sshLoggingService.setLevel(LoggingLevel.DEBUG);
```
RRRadicalEdward, line 73: `LoggingLevel` is not exported from the `web-ssh-gui` module anymore. Could you fix it on your side?

webapp/src/client/app/modules/web-client/ard/web-client-ard.component.ts
```diff
@@ -73,6 +78,11 @@ export class WebClientArdComponent extends WebClientBaseComponent implements OnI
       icon: 'dvl-icon dvl-icon-screen',
       action: () => this.scaleTo(this.screenScale.Real),
     },
+    {
+      label: 'Toggle Cursor Kind',
+      icon: 'dvl-icon dvl-icon-toggle',
```
RRRadicalEdward, line 83: What icons are available?  I don't know which icon to choose (`dvl-icon-toggle` does not correspond to any icon)

RRRadicalEdward:
@CBenoit Please review when you get a chance

webapp/src/client/app/modules/web-client/ssh/web-client-ssh.component.ts
CBenoit: Sure, don’t worry about this, we’ll take care of that.

webapp/tsconfig.json
```diff
@@ -45,6 +45,7 @@
     //"enableI18nLegacyMessageIdFormat": false,
     //"strictInjectionParameters": true,
     //"strictInputAccessModifiers": true,
-    "strictTemplates": true
+    "strictTemplates": false,
```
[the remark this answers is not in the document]
CBenoit: I think it’s possible to tell the Angular compiler "trust me, it exists" using `$any(…)`:
```
[module]="$any(someModule)"
```

I also tried to ask ChatGPT (as I’m not super familiar with Angular myself :joy:) if it was possible to teach about the interface locally, and it appears to be possible:

> To inform the Angular compiler about the properties of a custom web component (e.g., a `module` property on `<my-custom-element>`), you can extend the global type definitions.
> 
> Create a file named `custom-elements.d.ts` (or similar) and add the following:
> 
> ```ts
> interface MyCustomElement extends HTMLElement {
>   module: any; // Replace `any` with a more specific type if available
> }
> 
> declare global {
>   interface HTMLElementTagNameMap {
>     'my-custom-element': MyCustomElement;
>   }
> }
> ```

There may be other ways as well. Could you consider that before disabling the checks?

CBenoit (commented):
This the PR is quite big, would you be okay with adding "Copilot" as a reviewer and let it perform a first pass? I think I’m not able to add it myself, could you try on your side?
https://docs.github.com/en/copilot/using-github-copilot/code-review/using-copilot-code-review

webapp/package-lock.json
```diff
@@ -54,10 +55,120 @@
         "@rollup/rollup-linux-x64-gnu": "^4.21.2"
       }
     },
+    "../../ironrdp/web-client/iron-remote-desktop": {
```
CBenoit, line 58: issue? Could you re-run npm install and ensure the lock is not referring to local paths?

RRRadicalEdward:
> This the PR is quite big, would you be okay with adding "Copilot" as a reviewer and let it perform a first pass? 

Yes. Sure

> I think I’m not able to add it myself, could you try on your side? https://docs.github.com/en/copilot/using-github-copilot/code-review/using-copilot-code-review

I'm afraid I can't do it either. I can't set a reviewer because of a lack of rights

CBenoit:
> > I think I’m not able to add it myself, could you try on your side? https://docs.github.com/en/copilot/using-github-copilot/code-review/using-copilot-code-review
> 
> I'm afraid I can't do it either. I can't set a reviewer because of a lack of rights

Maybe because you are not part of the organization. Since I can’t assign it on my side either, it’s likely not possible to request a review at all in this case.

package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "devolutions-gateway",
+  "name": "devolutions-jet",
```
CBenoit, line 2: question: Why is it renamed?

webapp/src/client/app/modules/web-client/form/form-controls/resolution-quality-control/resolution-quality-control.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  @Input() inputFormData;
+
+  resolutionQualityOptions: SelectItem[];
+
+  constructor(private formService: WebFormService) {
+    super();
+  }
+
+  ngOnInit(): void {
+    this.resolutionQualityOptions = this.formService.getResoltuionQualityOptions();
```
CBenoit, line 26: typo: ResoltuionQuality -> ResolutionQuality

webapp/src/client/app/modules/web-client/rdp/web-client-rdp.component.ts
```diff
@@ -53,13 +55,16 @@ export class WebClientRdpComponent extends WebClientBaseComponent implements OnI
   @Output() sizeChange: EventEmitter<void> = new EventEmitter<void>();
 
   @ViewChild('sessionRdpContainer') sessionContainerElement: ElementRef;
-  @ViewChild('ironGuiElement') ironGuiElement: ElementRef;
+  @ViewChild('ironRemoteDesktopElementRdp') ironRemoteDesktopElementRdp: ElementRef;
```
CBenoit, line 58: question: Is it necessary to name it ironRemoteDesktopElementRdp over ironRemoteDesktopElement (it’s part of a component called WebClient**Rdp**Component)

webapp/src/client/app/modules/web-client/rdp/web-client-rdp.component.ts
```diff
@@ -337,18 +369,37 @@ export class WebClientRdpComponent extends WebClientBaseComponent implements OnI
...
+
+    if (connectionParameters.domain != null) {
+      configBuilder.withServerDomain(connectionParameters.domain);
+    }
+
+    if (connectionParameters.screenSize != null) {
+      configBuilder.withDesktopSize(connectionParameters.screenSize);
+    }
+
+    configBuilder.withExtension(displayControl(true));
```
CBenoit, line 388: suggestion: Make this configurable via a checkbox in a follow up PR. (For clarity, the label must be more detailed that "Enable display control").

webapp/src/client/app/modules/web-client/rdp/web-client-rdp.component.ts
```diff
@@ -337,18 +369,37 @@ export class WebClientRdpComponent extends WebClientBaseComponent implements OnI
...
+      configBuilder.withExtension(preConnectionBlob(connectionParameters.preConnectionBlob));
+    }
+
+    if (connectionParameters.kdcProxyUrl != null) {
+      configBuilder.withExtension(kdcProxyUrl(connectionParameters.kdcProxyUrl));
+    }
+
+    const config = configBuilder.build();
+
+    this.setKeyboardUnicodeMode(true);
```
CBenoit, line 400: issue/question: The unicode mode is not supported by the RDP module (yet), and should not be default.

webapp/src/client/app/modules/web-client/telnet/web-client-telnet.component.ts
CBenoit, line 1: @irvingoujAtDevolution Can you verify this file?

webapp/src/client/app/modules/web-client/form/form-components/ssh/ssh-form.component.ts
```diff
@@ -64,14 +64,13 @@ export class SshFormComponent extends BaseComponent implements OnInit, OnDestroy
...
+        formGroup: this.form,
+        controlName: 'authMode',
         inputFormData,
-        true,
-        false,
-        SshAuthMode.Username_and_Password,
-      );
+        isRequired: true,
+        defaultValue: SshAuthMode.Username_and_Password,
+      });
```
CBenoit, line 73: praise: I think it’s more readable that way :+1:  (named parameters)

CBenoit (commented):
I’ve gone through the PR! Many improvements, good job! To be picky: I think it’s better to open smaller PRs for the small refactorings unrelated to the main change, and it’s easier to review all the things separately.

webapp/tsconfig.json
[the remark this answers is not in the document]
RRRadicalEdward: I got the following from ChatGPT on my side:

> That warning:
> `Property 'module' is not provided by any applicable directives nor by <my-custom-element> element`.
> ...is coming from Angular’s template type checking. It doesn’t know that module is a valid property on your custom element.
> Even though we've extended the TypeScript types with HTMLElementTagNameMap, Angular’s template compiler doesn’t to use that. It uses directive metadata and DOM schema info to validate templates
> 

We can enable the `strict template` and have the warning. It won't affect the code in any way. It would appear just this way:
image

package-lock.json
RRRadicalEdward: My bad. I will revert

webapp/tsconfig.json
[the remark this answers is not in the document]
CBenoit: Hmm… I think I still prefer the warning. And I guess `[module]="$any(someModule)"` will not help?

webapp/package-lock.json
RRRadicalEdward: I re-ran it. It should be fine now

RRRadicalEdward:
> * What does "Disable cursor" refers to?

It disables the `Cursor` pseudo-encoding which tells the server to draw the mouse cursor into the framebuffer.
The user may want to disable it to improve performance, because it slightly reduces the server's load.

> I would prefer we state the flags positively: "Enable cursor" (checked by default) instead of "Disable cursor" (unchecked by default). It’s easier to scan. (I think this is generally true and does not apply only to UI.)

Done.

webapp/tsconfig.json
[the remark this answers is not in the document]
RRRadicalEdward: Yeah, because it complains about `[module`] not `someModule`

RRRadicalEdward:
> * Do we support "non-Unicode input" at all for VNC? I’m pretty sure Unicode input should be hardcoded to "true" for VNC and ARD.

Yeah. I removed it for VNC and ARD

RRRadicalEdward:
> IIRC, we do not support Unicode input for RDP yet? This is planned, but until it’s actually implemented, could you hide the checkbox (comment it out with a comment explaining this should be enabled back later).
> 
> (Also, this should not be checked by default for RDP in my opinion.)

I commented it out

RRRadicalEdward:
> > ARD.mp4
> 
> As mentioned above, Unicode input should maybe be hardcoded to "true", with no checkbox.

Done

RRRadicalEdward:
> * What about enabling all of them by default? I think most people don’t really care about what is enabled until they want to disable some specific codecs for various reasons, so we should let the negotiation happens with all the codecs by default.

Done

webapp/tsconfig.json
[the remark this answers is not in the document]
RRRadicalEdward: I re-enabled it

webapp/src/client/app/modules/web-client/rdp/web-client-rdp.component.ts
RRRadicalEdward: I found the related code in IronRDP - https://github.com/Devolutions/IronRDP/blob/3029c8f9094094a3e0effa57a29d0df78c5e1823/crates/ironrdp-input/src/lib.rs#L307-L336, and I was able to trigger it in the webapp when the setting is enabled:
image.

Is there anything more to it that I don't know?

CBenoit:
> > IIRC, we do not support Unicode input for RDP yet? This is planned, but until it’s actually implemented, could you hide the checkbox (comment it out with a comment explaining this should be enabled back later).
> 
> > 
> 
> > (Also, this should not be checked by default for RDP in my opinion.)
> 
> 
> 
> I commented it out

I'm sorry, you were right: we support it already. My only suggestion is to uncheck it by default (this is the default behavior in the official client too I believe).

webapp/src/client/app/modules/web-client/ssh/web-client-ssh.component.ts
irvingoujAtDevolution: Hi! Logging level is now exposed from @devolutions/terminal-shared it suppose to be a peer dependency! We can fix it on ourside so npm add it altomatically, you can also try just install it manually.

package-lock.json
CBenoit: Wait, I just realized this file should not even exist. It seems it has been commit and merged by mistake :joy:

webapp/package.json
```diff
@@ -28,8 +28,9 @@
     "@angular/platform-browser-dynamic": "18.2.1",
     "@angular/router": "18.2.1",
     "@devolutions/icons": "4.0.8",
-    "@devolutions/iron-remote-gui": "^0.11.6",
-    "@devolutions/iron-remote-gui-vnc": "^0.2.2",
+    "@devolutions/iron-remote-desktop": "0.2.0",
+    "@devolutions/iron-remote-desktop-rdp": "0.1.1",
+    "@devolutions/iron-remote-desktop-vnc": "0.2.0",
```
CBenoit, line 33: suggestion: As long as we follow the semantic versioning, it’s fine to use `^` .
```suggestion
    "@devolutions/iron-remote-desktop": "^0.2.0",
    "@devolutions/iron-remote-desktop-rdp": "^0.1.1",
    "@devolutions/iron-remote-desktop-vnc": "^0.2.0",
```

webapp/src/client/app/modules/web-client/ssh/web-client-ssh.component.ts
CBenoit: @irvingoujAtDevolution Do you think you could take care updating the said packages and update code in the way this was intended?

@RRRadicalEdward For this PR, I think it’s best if you revert the update of the `@devolutions/web-ssh-gui` and `@devolutions/web-telnet-gui` packages, and associated changes.
It seems there was a breaking change in a patch number version bump, and you experienced it because the versions were not pinned and the packages got updated (as expected according to the semantic versioning).
To avoid the issue now, pin web-telnet-gui to 0.2.19 and @devolutions/web-ssh-gui to 0.3.1 (using =) in the manifest.

CBenoit:
> > * What does "Disable cursor" refers to?
> 
> It disables the `Cursor` pseudo-encoding which tells the server to draw the mouse cursor into the framebuffer. The user may want to disable it to improve performance, because it slightly reduces the server's load.

Got you! I think that if the support supports it, we may receive cursor images that we render ourselves, right? And I understand that in this case the performance is not badly impacted.

What about bringing the following changes?

- Checkbox label: `Enable cursor pseudo-encoding` (a bit, jargony, but specific :thinking:)
- Add a a tooltip: `When enabled, the server may render the mouse cursor into the framebuffer, which can affect performance depending on the server capabilities`

webapp/src/client/app/modules/web-client/ssh/web-client-ssh.component.ts
RRRadicalEdward: Reverted

webapp/src/client/app/modules/web-client/form/form-controls/resolution-quality-control/resolution-quality-control.component.ts
RRRadicalEdward: Done

webapp/package.json
RRRadicalEdward: Done

RRRadicalEdward:
> > > IIRC, we do not support Unicode input for RDP yet? This is planned, but until it’s actually implemented, could you hide the checkbox (comment it out with a comment explaining this should be enabled back later).
> > 
> > 
> > > 
> > 
> > 
> > > (Also, this should not be checked by default for RDP in my opinion.)
> > 
> > 
> > I commented it out
> 
> I'm sorry, you were right: we support it already. My only suggestion is to uncheck it by default (this is the default behavior in the official client too I believe).

Done

package-lock.json
RRRadicalEdward: Removed the file

webapp/src/client/app/modules/web-client/ard/web-client-ard.component.ts
RRRadicalEdward: I added usage of `dvl-icon-cursor`, but I'm not sure about "on" and "off" states. Can I address it in a follow-up?

webapp/src/client/app/modules/web-client/rdp/web-client-rdp.component.ts
RRRadicalEdward: I found that we use here - https://github.com/Devolutions/IronRDP/blob/3029c8f9094094a3e0effa57a29d0df78c5e1823/web-client/iron-remote-desktop/src/services/remote-desktop.service.ts#L318-L325

webapp/src/client/app/modules/web-client/telnet/web-client-telnet.component.ts
RRRadicalEdward: Since we reverted the changes in this file, there is nothing to review

webapp/src/client/app/modules/web-client/rdp/web-client-rdp.component.ts
RRRadicalEdward: Done in https://github.com/Devolutions/devolutions-gateway/pull/1333

webapp/src/client/app/shared/resolvers/wasm-init.resolver.ts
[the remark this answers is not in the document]
CBenoit: Oh… It appears less straightforward that I hoped then. On second thought, let’s keep things as is. I think the page is already loading at a very decent speed, so it’s not necessary to put too much effort into optimizing this.

webapp/src/client/app/modules/web-client/ard/web-client-ard.component.ts
CBenoit: Can be a follow up!

CBenoit (approved):
LGTM! Excellent work!

CBenoit:
Can you address this comment https://github.com/Devolutions/devolutions-gateway/pull/1329#issuecomment-2876913378 in a follow up PR? Thanks!

webapp/src/client/app/modules/web-client/ard/web-client-ard.component.ts
RRRadicalEdward: Addressed in https://github.com/Devolutions/devolutions-gateway/pull/1337

PR #1334, opened 2025-05-13
feat(pedm): log and query JIT elevations

thenextman:
JIT elevations (where the user right-clicks and chooses "Run elevated") are logged in the SQL datastore. The full elevation request is not logged - the current capabilities don't give any worth in logging anything about the asker application (beyond the path, maybe). Additionally we log the authenticate status of the target application but not the actual certificates used to sign it. This can be added later if needed, but that level of data is probably too high fidelity for now.

We need the user to exist in the database, which isn't happening currently; so we do an upsert of the user information at the time of logging.

A new route is added /log/jit to query the elevation logs from RDM.

The user must filter by start and end times, and optionally by user. If the user is not elevated, they can only query their own logs.

The user can specify a sort column and sort direction (currently what is available here is determined by the columns we show in RDM, it can be extended).

The user must specify a page size and offset; this facilitates paging the query results rather than sending back the whole data set on one call.

The data set returned is, again, driven by what we currently display in RDM. It can be extended easily.


CBenoit (approved):
LGETM :saluting_face:

PR #1337, opened 2025-05-14
refactor(webapp): add 'on' and 'off' states for `Toggle Cursor Button`

RRRadicalEdward:
Fix for https://github.com/Devolutions/devolutions-gateway/pull/1329#discussion_r2087157731


CBenoit (approved):
LGTM! Do you think you could you add a video? I’m curious to see how it looks like.

RRRadicalEdward:
> LGTM! Do you think you could you add a video? I’m curious to see how it looks like. Also, you need to run the formater: `npm run fmt:write`.

Sure. Here is a video: https://github.com/user-attachments/assets/07eb217b-3df5-4f5a-a91a-01be4334c089

PR #1338, opened 2025-05-15
feat(pedm): improvements for log querying

thenextman:
Improvements for querying PEDM logs, to support needed features in RDM

- `JitElevationLogRow` is expanded to include more fields
- The existing `get_jit_elevation_logs` returns a subset of fields
- Individual log entries can be queried with `get_jit_elevation_log(id)`, returning the full data
- Users can be queried


crates/devolutions-pedm/src/api/log.rs
```diff
@@ -1,11 +1,30 @@
...
-use axum::extract::State;
+use axum::extract::{Path, State};
 use axum::{Extension, Json};
 
+use super::policy::PathIntIdPath;
+
+async fn get_jit_elevation_log_id(
+    Path(id): Path<PathIntIdPath>,
+    Extension(named_pipe_info): Extension<NamedPipeConnectInfo>,
+    NoApi(State(_state)): NoApi<State<AppState>>,
```
CBenoit, line 14: question: This is not used, is there a reason to keep parameter?

crates/devolutions-pedm/src/api/log.rs
```diff
@@ -1,11 +1,30 @@
...
+use crate::log::{JitElevationLogPage, JitElevationLogQueryOptions, JitElevationLogRow};
 use aide::axum::ApiRouter;
 use aide::NoApi;
-use axum::extract::State;
+use axum::extract::{Path, State};
 use axum::{Extension, Json};
 
+use super::policy::PathIntIdPath;
+
+async fn get_jit_elevation_log_id(
```
CBenoit, line 11: naming: I think it’s easier to understand if a `by` is inserted
```suggestion
async fn get_jit_elevation_log_by_id(
```

CBenoit (approved):
LGTM!

crates/devolutions-pedm/src/api/log.rs
CBenoit: Okay! I don’t think it’s worth following this pattern, but I’ll rename in a separate PR (if I don’t forget :joy:)

PR #1340, opened 2025-05-15
fix(webapp): change the tooltip for enableCursor checkbox

RRRadicalEdward:


webapp/src/client/app/modules/web-client/form/form-controls/enable-cursor-control/enable-cursor-control.component.html
```diff
@@ -3,7 +3,7 @@
         <p-checkbox
                 formControlName="enableCursor"
                 label="Enable Cursor Pseudo-Encoding"
-                pTooltip="When enabled, the server may render the mouse cursor into the framebuffer, which can affect performance depending on the server capabilities"
+                pTooltip="When enabled, the server will not render the mouse cursor into the framebuffer, which can improve performance depending on the server capabilities. Instead, client will render cursor locally."
```
CBenoit, line 6: Oh, I think I misunderstood something when I suggested this text.

To keep it concise, what about this?

> When enabled, compatible servers will not render the mouse cursor into the framebuffer. The cursor is drawn locally, improving perceived performance over slow links and reducing server load.

RRRadicalEdward: I love it!

CBenoit (approved):

PR #1343, opened 2025-05-15
feat(pedm): add version check API and signature check API

thenextman:
Add a new policy option `target_must_be_signed`; this forces the target executable to have a valid auhenticode signature as per the policy.

Realistically, this should be a "rule" so I'm probably introducing some slight technical debt here but I believe it's a worthwhile feature and a low-hanging fruit we can squeeze in for the release.

Add the application (Agent) version to the "About" API endpoint. This is for future proofing: RDM needs a way to know what version of PEDM he's talking to, in case we add/remove/change APIs. 

The implementation is in Win32 and I believe it to be correct, but it could use some close scrutiny.

I generated the OpenAPI in a. separate commit for easier reviewing.


crates/win-api-wrappers/src/utils.rs
```diff
@@ -457,6 +457,69 @@ pub fn expand_environment_path(src: &Path, environment: &HashMap<String, String>
...
 }
 
+use windows::Win32::Foundation::{BOOL, HMODULE};
+use windows::Win32::Storage::FileSystem::{
+    GetFileVersionInfoSizeW, GetFileVersionInfoW, VerQueryValueW, VS_FIXEDFILEINFO,
+};
+use windows::Win32::System::LibraryLoader::{GetModuleFileNameW, GetModuleHandleW};
+
+pub fn get_exe_version() -> Result<String, anyhow::Error> {
+    // SAFETY: Passing None to GetModuleHandleW requests the handle to the current process main executable module
```
CBenoit, line 467: nitpick: This is not really a safety comment. A safety comment purpose is to _explain why the unsafe call is safe_. Essentially to make unsafe code easier to audit. In this case, it’s a trivial one.

```suggestion
    // Passing None to GetModuleHandleW requests the handle to the current process main executable module
    // SAFETY: FFI call with no outstanding preconditions.
```

CBenoit (approved):
By the way, in the Gateway we use the `CARGO_PKG_VERSION` env variable that is available at compile time:

https://github.com/Devolutions/devolutions-gateway/blob/bee85784ea1e734332f16807e22b3dc4c6c0e16f/devolutions-gateway/src/api/health.rs#L64

This is replaced by, e.g., "2025.1.5", at compile time. I think I prefer that, unless there is a reason to stick to the runtime extraction from the executable?

Approving now to smooth the process for you.

thenextman:
> By the way, in the Gateway we use the `CARGO_PKG_VERSION` env variable that is available at compile time:
> 
> https://github.com/Devolutions/devolutions-gateway/blob/bee85784ea1e734332f16807e22b3dc4c6c0e16f/devolutions-gateway/src/api/health.rs#L64
> 
> This is replaced by, e.g., "2025.1.5", at compile time. I think I prefer that, unless there is a reason to stick to the runtime extraction from the executable?
> 
> Approving now to smooth the process for you.

Yes, as I wrote on Slack because we are in a crate that gives use the crate version

I do understand the alternative approach of using the build.rs to properly embed the version number in the crate, as also proposed by @allan2. Honestly I'm not sure which way I prefer.

Part of this is my own background: coming from a C/C++/C# perspective, the proper way to do this is to query the main module version. i.e. In C# you'd do `GetEntryAssembly` or `GetExecutingAssembly` and obtain the version that way. I like the idea that if I want to get the Agent version from _another_ crate, I can call an existing function without having to modify the build system. It does (of course) rely on the developer knowing of the existence of said function.

_But_ I do understand it's probably more idiomatic in rust to use a CARGO variable.

For now, I leave this like it is; but I'm not opposed to someone changing that to the alternative approach.

CBenoit:
> Part of this is my own background: coming from a C/C++/C# perspective, the proper way to do this is to query the main module version. i.e. In C# you'd do `GetEntryAssembly` or `GetExecutingAssembly` and obtain the version that way. I like the idea that if I want to get the Agent version from _another_ crate, I can call an existing function without having to modify the build system. It does (of course) rely on the developer knowing of the existence of said function.

That’s a fair point. Anyway, now we have the function, we can use it :wink:

PR #1349, opened 2025-05-20
chore: use workspace version

allan2:
This sets the version in one place.

This will make endpoints that return the version, such as `/about` in PEDM, return a consistent version based on `CARGO_PKG_VERSION`.

A PR will follow to bump the `devolutions-pedm` version (currently 0.0.0).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,10 @@ exclude = [
     "crates/sogar-registry"
 ]
 
+[workspace.package]
+version = "2025.1.5"
+
+
 [profile.profiling]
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -20,6 +20,10 @@ exclude = [
     "crates/sogar-registry"
 ]
 
+[workspace.package]
+version = "2025.1.5"
+
```
CBenoit, line 25:
```suggestion
```

CBenoit (approved):
LGTM!

PR #1356, opened 2025-05-22
refactor: store PEDM configuration in SQLite instead of JSON files

thenextman:
Previously the PEDM profiles and assignments were wrapped up in a `Policy` object, which was loaded at startup and then maintained via an `IdList` structure. API calls would have the `Policy` in their context and interact with it directly. 

This PR:

- Strips out unused features from the API surface area (session / temporary elevations, revoke elevation). We don't support them right now and it just adds noise.
- The remaining `Policy` item needed for elevation is the current user selected profile. We load this from the DB at the API front -end and pass it back to where it's needed.
- The API now calls the database directly to query/update/delete. In most cases transactions are used where we need to enforce consistency, but we are not expecting a high volume of calls or needing a high level of atomicity.

The API and database design is not the most well thought-out, but it closely reflects the _existing_ APIs to minimize disruption elsewhere (API clients like RDM, as well as avoid significant behaviour changes that may introduce new bugs).

Concerns are:

- There's no effort to maintain compatibility here. Existing PEDM users will need to uninstall, wipe out their %programdata%\Agent\pedm directory, and reinstall. Internally I'll provide a script to do that. RDM will be blocked from using any version that doesn't include these changes.
- The handling of users is not optimal (see @allan2 related PR). This goes back to Kyle's original implementation and if we had made these changes earlier, we could've addressed it, but there is no time now. We'll have some technical debt here to correct in the future.


crates/devolutions-pedm-shared/src/policy.rs
```diff
@@ -311,23 +311,17 @@ pub trait Identifiable {
 #[derive(Serialize, Deserialize, JsonSchema, Debug, Clone, Hash, PartialEq, Eq)]
 #[serde(rename_all = "PascalCase")]
 pub struct Profile {
-    pub id: Uuid,
+    pub id: Option<i64>,
```
CBenoit, line 314: question: So we ended up with changing for `i64`? Is it really okay for it to be optional?

crates/devolutions-pedm/src/api/state.rs
```diff
@@ -65,11 +60,11 @@ where
...
-impl FromRef<AppState> for Arc<RwLock<Policy>> {
-    fn from_ref(state: &AppState) -> Self {
-        Arc::clone(&state.policy)
-    }
-}
+// impl FromRef<AppState> for Policy {
+//     fn from_ref(state: &AppState) -> Self {
+//         &state.policy
+//     }
+// }
```
CBenoit, line 67: Feel free to delete if it’s not useful, it’s not very tricky code either

crates/devolutions-pedm/src/db/libsql.rs
```diff
@@ -178,24 +203,400 @@ impl Database for LibsqlConn {
...
+            .await?;
+        }
 
         tx.commit().await?;
         Ok(())
     }
 
+    async fn set_user_profile(&self, user: &User, profile_id: Option<i64>) -> Result<(), DbError> {
+        // `unwrap` should be sound, the user should not be able to select a profile unless they already exist in the database
+        let user_id = self.get_user_id(user).await?.unwrap();
```
CBenoit, line 547: issue: This is true, but I also don’t really agree that it’s fine for the same reasons we can’t trust client-side forms to never send malformed data to the server in web pages. Granted here everything is local and the worst that would happen is the agent process crashing. If you did that to avoid dealing with the error handling, I think it’s fine for the V1, but add a `FIXME` comment so I can look into that later :slightly_smiling_face:

CBenoit (approved):
LGTM! I don’t see any blocking issue :saluting_face:

crates/devolutions-pedm/src/api/state.rs
thenextman: Refactoring leftovers, gone

crates/devolutions-pedm-shared/src/policy.rs
thenextman: Good catch, this was a leftover from refactoring.

I changed the IDs to i64 as it's nicer for the database. The ID start at "1", so "0" is used for "no ID". 

Originally I had this as `Option` because I thought it was cleaner, but actually OpenAPI doesn't generate a `long?` on the client, it simply sets the data contract to **not serialize** the default value. So in the end there was absolutely no benefit to using `Option` here, and it causes noise on the rust side. I simply forgot to remove the `Option` after figuring all that out :) 

Removed now, which makes the API cleaner, which I'm happy about. Thanks!

crates/devolutions-pedm/src/db/libsql.rs
thenextman: You're absolutely right and this was pure laziness on my part for not wanting to add a new error type. I've done that now.

PR #1360, opened 2025-05-24
feat(dgw): proxy-based credentials injection support for RDP

CBenoit:
## Consumer side

- Provide and associate the proxy-target credential mapping with the association token using a preflight API call.
- Connect using the fake (proxy) credentials to the Devolutions Gateway as usual, with a PCB containing the association token.

## How it works

- Perform two-way forwarding between the client and the target until the TLS security upgrade.
- Separately perform the TLS upgrade for both the client and the server, effectively acting as a man-in-the-middle.
  -  The client must trust the TLS certificate configured in the Devolutions Gateway.
- Separately perform CredSSP authentification as server with the client, and as client with the target.
  - The fake, proxy credentials are used with the client.
  - The real, target credentials are used with the target.
- Proceed with the usual two-way forwarding (expect we can actually see and inspect all the traffic)

## Demo

[proxy-based-credentials-injection-prototype.webm](https://github.com/user-attachments/assets/d5380053-810d-4529-b3f9-1ed84c2d77c4)

### DIFF
--- a/docs/COOKBOOK.md
+++ b/docs/COOKBOOK.md
@@ -6,6 +6,7 @@ Developer-oriented cookbook for testing purposes.
 - [WebSocket-to-TCP tunnel using jetsocat](#websocket-to-tcp-tunnel-using-jetsocat)
 - [Standalone web application custom authentication](#standalone-web-application-custom-authentication)
 - [Preflight API](#preflight-api)
+- [Proxy-based credentials injection for RDP](#proxy-based-credentials-injection-for-rdp)
 
 ## RDP routing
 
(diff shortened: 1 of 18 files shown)


docs/COOKBOOK.md
CBenoit, line 1: @TheBestTvarynka I updated the cookbook with instructions explaining how you can test the full thing using only curl and freerdp :slightly_smiling_face:

pacmancoder (approved):
Looks good to me! The code is well-written and well-documented as always, nice work! 🥇

docs/COOKBOOK.md
TheBestTvarynka: Thank you 🙏 !

PR #1366, opened 2025-05-28
fix(pedm): fix flawed expand_environment function

thenextman:
`expand_environment` is flawed; because it splits the input on "%" there can be improper boundary detection. For example, given input with back-to-back variables like "%HOMEDIR%%HOMEPATH%".

Instead, we walk the string one character at a time and properly tokenize the variables. This still account for the malformed example in Kyle's original comment ("For strings such as "%MyVar%MyVar%", only the first occurrence should be replaced").


CBenoit (approved):
LGTM! Thank you!

PR #1369, opened 2025-06-03
feat(webapp): allow dynamically adjust `wheel_speed_factor` for VNC and ARD sessions

RRRadicalEdward:
In this PR, I've moved the `wheel_speed_factor` setting to the toolbar for VNC and ARD sessions to dynamically adjust this setting.
It looks this way:
image


webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
```diff
@@ -118,6 +118,17 @@ export class WebClientVncComponent extends WebClientBaseComponent implements OnI
     },
   ];
 
+  sliders = [
+    {
+      label: 'Wheel Speed',
+      value: 1,
+      onChange: (value: number) => this.setWheelSpeedFactor(value),
+      min: 0.1,
+      max: 5,
```
CBenoit: question/suggestion: do you think a max of `2` or `3` would be reasonable? 5 times faster than the default sounds a bit much, but I didn’t try on my side.

RRRadicalEdward: Done. Set it to `3` for VNC and ARD

webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
```diff
@@ -412,12 +412,12 @@ export class WebClientVncComponent extends WebClientBaseComponent implements OnI
...
-    if (connectionParameters.disableCursor) {
-      configBuilder.withExtension(disableCursor(true));
+    if (connectionParameters.enableCursor) {
+      configBuilder.withExtension(enableCursor(true));
     }
 
-    if (connectionParameters.disableExtendedClipboard) {
-      configBuilder.withExtension(disableExtendedClipboard(true));
+    if (connectionParameters.enableExtendedClipboard) {
+      configBuilder.withExtension(enableExtendedClipboard(true));
```
CBenoit: nitpick: it seems reasonable to remove the `if`:
```ts
configBuilder.withExtension(enableExtendedClipboard(connectionParameters.enableExtendedClipboard));
```
And this can go in the call chain above

RRRadicalEdward: Done

CBenoit (approved):
LGTM!

PR #1372, opened 2025-06-04
feat(dgw): support VMConnect through RDCleanPath

irvingoujAtDevolution:
Support the explicit VMConnect RDCleanPath shape from Devolutions/IronRDP#1505.

Generic PCBs keep the ordinary X.224-first path. VMConnect requests carry a Unicode payload with no X.224; Gateway encodes the binary PCB, writes it before TLS (bounded by the MS-RDPEPS 10s deadline), then leaves CredSSP and X.224 to the client. Credential injection is rejected for this ordering.

Depends-on: Devolutions/IronRDP#1505

Issue: Devolutions/IronRDP#1505

### DIFF
--- a/devolutions-gateway/src/rd_clean_path.rs
+++ b/devolutions-gateway/src/rd_clean_path.rs
@@ -251,13 +251,39 @@ fn is_vmconnect_request(cleanpath_pdu: &RDCleanPathPdu) -> bool {
 /// Encode the Hyper-V PCB V2 that the proxy writes before TLS.
 ///
 /// `payload` is the opaque Unicode string from RDCleanPath (`GUID` or `GUID;EnhancedMode=1`).
+///
+/// Encoded locally rather than via `ironrdp-pdu` 0.9.0: that crates.io release counts `cchPCB`
+/// with `chars().count()`, which under-counts non-BMP code points (UTF-16 surrogates). IronRDP
+/// master fixed this to `encode_utf16().count()` but has not published a crates.io bump yet, and
+/// its MSRV is ahead of Gateway. Match the fixed wire shape here so opaque Unicode payloads stay
+/// well-formed.
 fn encode_vmconnect_pcb_v2(payload: String) -> anyhow::Result<Vec<u8>> {
-    let pcb = ironrdp_pdu::pcb::PreconnectionBlob {
-        id: 0,
-        version: ironrdp_pdu::pcb::PcbVersion::V2,
-        v2_payload: Some(payload),
-    };
-    ironrdp_core::encode_vec(&pcb).context("encode VMConnect preconnection blob")
+    // PCB V2 layout (little-endian):
+    // cbSize u32 | flags u32 | version u32 | id u32 | cchPCB u16 | wszPCB UTF-16LE + NUL
+    const FIXED_PART_SIZE: usize =
+        4 /* cbSize */ + 4 /* flags */ + 4 /* version */ + 4 /* id */;
+    const VERSION_V2: u32 = 2;
+
+    let utf16: Vec<u16> = payload.encode_utf16().chain(core::iter::once(0)).collect();
+    let cch_pcb = u16::try_from(utf16.len()).context("VMConnect PCB payload too long")?;
+    let utf16_byte_len = utf16.len().checked_mul(2).context("VMConnect PCB payload too long")?;
+    let total_size = FIXED_PART_SIZE
+        .checked_add(2 /* cchPCB */)
+        .and_then(|n| n.checked_add(utf16_byte_len))
+        .context("VMConnect PCB payload too long")?;
+    let cb_size = u32::try_from(total_size).context("VMConnect PCB payload too long")?;
+
+    let mut out = Vec::with_capacity(total_size);
+    out.extend_from_slice(&cb_size.to_le_bytes());
+    out.extend_from_slice(&0u32.to_le_bytes()); // flags
+    out.extend_from_slice(&VERSION_V2.to_le_bytes());
+    out.extend_from_slice(&0u32.to_le_bytes()); // id
+    out.extend_from_slice(&cch_pcb.to_le_bytes());
+    for unit in utf16 {
+        out.extend_from_slice(&unit.to_le_bytes());
+    }
+    debug_assert_eq!(out.len(), total_size);
+    Ok(out)
 }
 
 /// Cert-chain-only success response after VMConnect PCB + TLS (no X.224).
@@ -989,6 +1015,23 @@ mod tests {
         let payload = "21c82e1f-2368-43d5-9cb6-a7c99c449bba".to_owned();
         let bytes = encode_vmconnect_pcb_v2(payload).expect("encode");
         assert_eq!(bytes.len(), 92);
+    }
+
+    #[test]
+    fn encodes_non_bmp_payload_with_utf16_code_unit_cch() {
+        // U+1F600 needs a UTF-16 surrogate pair (2 code units, 1 scalar).
+        // crates.io ironrdp-pdu 0.9.0 would set cchPCB = chars+NUL = 5; correct is 6.
+        let payload = "vm-\u{1F600}".to_owned();
+        assert_eq!(payload.chars().count(), 4);
+        assert_eq!(payload.encode_utf16().count(), 5);
+
+        let bytes = encode_vmconnect_pcb_v2(payload.clone()).expect("encode");
+        let cch = u16::from_le_bytes([bytes[16], bytes[17]]);
+        assert_eq!(cch, 6, "cchPCB must count UTF-16 code units including NUL");
+        assert_eq!(bytes.len(), 16 + 2 + usize::from(cch) * 2);
+
+        let decoded: ironrdp_pdu::pcb::PreconnectionBlob = ironrdp_core::decode(&bytes).expect("decode round-trip");
+        assert_eq!(decoded.v2_payload.as_deref(), Some(payload.as_str()));
     }
 
     #[test]
--- a/devolutions-gateway/src/rd_clean_path.rs
+++ b/devolutions-gateway/src/rd_clean_path.rs
@@ -1,6 +1,7 @@
 use std::io::{self, ErrorKind};
 use std::net::SocketAddr;
 use std::sync::Arc;
+use std::time::Duration;
 
 use anyhow::Context as _;
 use ironrdp_pdu::nego;
@@ -10,6 +11,9 @@ use tap::prelude::*;
 use thiserror::Error;
 use tokio::io::{AsyncRead, AsyncReadExt as _, AsyncWrite, AsyncWriteExt as _};
 use tracing::field;
+
+/// MS-RDPEPS upper bound for transmitting the complete Preconnection Blob after TCP connect.
+const PCB_TRANSMIT_DEADLINE: Duration = Duration::from_secs(10);
 
 use crate::config::Conf;
 use crate::credential_injection_kdc::{CredentialInjectionKdc, CredentialService};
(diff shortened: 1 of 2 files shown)


devolutions-gateway/src/rd_clean_path.rs
```diff
@@ -212,45 +212,88 @@ async fn process_cleanpath(
...
+    // │                                                             │    │       client         │
+    // └─────────────────────────────────────────────────────────────┘    └──────────────────────┘
+    // │X224 connection request → X224 connection response → TLS hs  │ →  │CredSSP → ...         │
+    // └─────────────────────────────────────────────────────────────┘    └──────────────────────┘
+    //
+    // Summary:
+    // - With PCB: Gateway handles (1) sending PCB, (2) TLS handshake, then leaves CredSSP
+    //   and X224 connection request/response to IronRDP client
+    // - Without PCB: Gateway handles (1) X224 connection request, (2) X224 connection response,
+    //   then leaves TLS handshake and CredSSP to IronRDP client
```
CBenoit: praise: Good summary and explanation.

issue: However, this implementation assumes that the presence of a Preconnection Blob (PCB) implies the use of the Hyper-V VMConnect flow, where X.224 messages are exchanged after the CredSSP sequence. However, this assumption is overly restrictive.
While the PCB is indeed utilized in Hyper-V scenarios, particularly with VMConnect, which employs a unique connection sequence involving CredSSP prior to X.224 negotiation, it's important to recognize that the PCB is a general-purpose mechanism that does not necessarily imply an altered connection sequence. RDP clients, such as MSTSC and FreeRDP, support specifying a PCB without altering the conventional connection sequence, where X.224 negotiation precedes CredSSP authentication.

suggestion: The implementation should not default to the Hyper-V VMConnect flow solely based on the presence of a PCB. For instance, you could verify whether `cleanpath_pdu.x224_connection_pdu` is `Some` or `None`. If it’s `None`, it seems fair to assume the Hyper-V VMConnect flow.

devolutions-gateway/src/rd_clean_path.rs
```diff
@@ -234,45 +234,89 @@ async fn process_cleanpath(
...
+    // Send preconnection blob and/or X224 connection request
+    match (&cleanpath_pdu.preconnection_blob, &cleanpath_pdu.x224_connection_pdu) {
+        (None, None) => {
+            return Err(CleanPathError::BadRequest(anyhow::anyhow!(
+                "RDCleanPath PDU is missing both preconnection blob and X224 connection PDU"
+            )));
+        }
+        (Some(general_pcb), Some(_)) => {
+            server_stream.write_all(general_pcb.as_bytes()).await?;
```
CBenoit: issue: The logic for sending the PCB in the general case is the same: we need to put the string into a PCB PDU and re-encode it.

devolutions-gateway/src/rd_clean_path.rs
```diff
@@ -336,8 +380,12 @@ pub async fn handle(
...
     trace!("Sending RDCleanPath response");
 
-    let rdcleanpath_rsp = RDCleanPathPdu::new_response(server_addr.to_string(), x224_rsp, x509_chain)
-        .map_err(|e| anyhow::anyhow!("couldn’t build RDCleanPath response: {e}"))?;
+    let rdcleanpath_rsp = RDCleanPathPdu::new_response(
+        server_addr.to_string(),
+        x224_rsp,
+        x509_chain,
+    )
+    .map_err(|e| anyhow::anyhow!("couldn’t build RDCleanPath response: {e}"))?;
```
CBenoit: suggestion: Use `context` or `with_context` + `format!`

devolutions-gateway/src/rd_clean_path.rs
```diff
@@ -234,45 +234,89 @@ async fn process_cleanpath(
...
 
-    Ok(CleanPathResult {
+    return Ok(CleanPathResult {
         destination: selected_target.to_owned(),
         claims,
         server_addr,
         server_stream,
         x224_rsp,
-    })
+    });
```
CBenoit: style: return is not required

devolutions-gateway/Cargo.toml
```diff
@@ -28,8 +28,8 @@ job-queue = { path = "../crates/job-queue" }
 job-queue-libsql = { path = "../crates/job-queue-libsql" }
 ironrdp-pdu = { version = "0.5", features = ["std"] }
 ironrdp-core = { version = "0.1", features = ["std"] }
-ironrdp-rdcleanpath = "0.1"
 ironrdp-tokio = "0.5"
+ironrdp-rdcleanpath = { path = "C:\\dev\\IronRDP\\crates\\ironrdp-rdcleanpath" }
```
irvingoujAtDevolution: will replace with commit/new version later

devolutions-gateway/src/rd_clean_path.rs
irvingoujAtDevolution: updated!

devolutions-gateway/src/rd_clean_path.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/rd_clean_path.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/rd_clean_path.rs
irvingoujAtDevolution: updated

devolutions-gateway/Cargo.toml
irvingoujAtDevolution: Staying on crates.io `ironrdp-rdcleanpath 0.2.2` on purpose.

IronRDP master shipped the VMConnect helpers in 0.2.3 source, but crates.io is still 0.2.2, so Gateway keeps VERSION_1 field detection and hand-builds the cert-only response. No dep bump until 0.2.3 is published — then we can switch to `new_vmconnect_response` / pin `>=0.2.3`.

irvingoujAtDevolution:
follow-up in https://github.com/Devolutions/devolutions-gateway/commit/d1f7e396

- VMConnect only when no X.224 + non-empty PCB payload
- proxy encodes PCB V2, 10s write bound + flush, then TLS
- cred injection rejected with RDCleanPath 400 for this path
- unit tests for the fork; still on crates.io rdcleanpath 0.2.2 until 0.2.3 is published

lab on NESTED-TGT (`10.10.0.3:2179`) with IronRDP master `5c181624`: direct + RDCleanPath both got HYBRID and frames. gateway log showed `pcb_len=122`.

all review threads resolved.

devolutions-gateway/src/rd_clean_path.rs
```diff
@@ -229,14 +234,61 @@ struct ConnectedRdpServer {
...
+/// Encode the Hyper-V PCB V2 that the proxy writes before TLS.
+///
+/// `payload` is the opaque Unicode string from RDCleanPath (`GUID` or `GUID;EnhancedMode=1`).
+fn encode_vmconnect_pcb_v2(payload: String) -> anyhow::Result<Vec<u8>> {
+    let pcb = ironrdp_pdu::pcb::PreconnectionBlob {
+        id: 0,
+        version: ironrdp_pdu::pcb::PcbVersion::V2,
+        v2_payload: Some(payload),
+    };
+    ironrdp_core::encode_vec(&pcb).context("encode VMConnect preconnection blob")
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Fixed in https://github.com/Devolutions/devolutions-gateway/commit/5a6940d49e6ab68a6f980e206cbef3e6a9ee17ce

crates.io is still on ironrdp-pdu 0.9.0 (chars().count()), and IronRDP master (with the fix) is MSRV 1.94 while Gateway is on 1.90, so I encode PCB V2 locally with UTF-16 code-unit cchPCB. Added a non-BMP unit test that asserts cchPCB == 6.

devolutions-gateway/src/rd_clean_path.rs
```diff
@@ -616,8 +717,14 @@ pub async fn handle(
...
 
-    let rdcleanpath_rsp = RDCleanPathPdu::new_response(server_addr.to_string(), x224_rsp, x509_chain)
-        .context("build RDCleanPath response")?;
+    // Ordinary responses include X.224 CC. VMConnect responses are cert-chain only; the client
+    // runs CredSSP then X.224 on the upgraded path.
+    let rdcleanpath_rsp = if let Some(x224_rsp) = x224_rsp {
+        RDCleanPathPdu::new_response(server_addr.to_string(), x224_rsp, x509_chain)
+            .context("build RDCleanPath response")?
+    } else {
+        build_vmconnect_response(server_addr.to_string(), x509_chain).context("build VMConnect RDCleanPath response")?
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Ran `cargo +nightly fmt --all` in https://github.com/Devolutions/devolutions-gateway/commit/5a6940d49e6ab68a6f980e206cbef3e6a9ee17ce

devolutions-gateway/src/rd_clean_path.rs
```diff
@@ -816,3 +923,83 @@ impl From<&io::Error> for WsaError {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Leaving these as in-module unit tests on purpose.

`testsuite/` is the integration/CLI harness. These checks target private helpers (`is_vmconnect_request`, `encode_vmconnect_pcb_v2`, `build_vmconnect_response`), and the rest of `devolutions-gateway` already keeps that kind of coverage next to the code (`#[cfg(test)]` in ~15 modules). Exposing the helpers just to move the tests would be worse signal/noise.

mamoreau-devolutions (approved):

CBenoit (commented):
The protocol branching is valid, but the implementation carries more machinery than necessary:

1. Replace `is_vmconnect_request` plus later `Option` extraction with one `match` over `(x224_connection_pdu, preconnection_blob)`. This removes impossible error paths and keeps request classification beside handling.
2. Remove the redundant X.224 check after `connect_rdp_server`; the credential-injection path already rejects requests without X.224.
3. Keep the 10-second timeout only on the VMConnect PCB write. Applying it to the legacy path broadens this PR's behavior and duplicates error handling.
4. Remove both `flush()` calls and the buffering comment. Neither `TcpStream` nor the current tunnel stream gains ordering guarantees from them.
5. Construct the response once, mapping `Option<Vec<u8>>` into the optional X.224 field, rather than maintaining a separate VMConnect response builder and branch.
6. Collapse the request-shape tests into a table-driven test. Keep the wire-format regression tests because the pinned IronRDP encoder mishandles non-BMP UTF-16 lengths.

Longer term, upgrade to an IronRDP release exposing `RDCleanPathMessage`; that should eliminate the local classifier, response builder, and PCB encoder. Until then, retaining the corrected local encoder is safer than reverting to the known-buggy dependency implementation.

devolutions-gateway/src/rd_clean_path.rs
```diff
@@ -229,14 +234,87 @@ struct ConnectedRdpServer {
...
+/// Encode the Hyper-V PCB V2 that the proxy writes before TLS.
+///
+/// `payload` is the opaque Unicode string from RDCleanPath (`GUID` or `GUID;EnhancedMode=1`).
+///
+/// Encoded locally rather than via `ironrdp-pdu` 0.9.0: that crates.io release counts `cchPCB`
+/// with `chars().count()`, which under-counts non-BMP code points (UTF-16 surrogates). IronRDP
+/// master fixed this to `encode_utf16().count()` but has not published a crates.io bump yet, and
+/// its MSRV is ahead of Gateway. Match the fixed wire shape here so opaque Unicode payloads stay
+/// well-formed.
+fn encode_vmconnect_pcb_v2(payload: String) -> anyhow::Result<Vec<u8>> {
```
CBenoit, line 260: This local encoder is justified as a temporary workaround: `ironrdp-pdu` 0.9.0 counts Unicode scalar values instead of UTF-16 code units for `cchPCB`, while the upstream fix is not yet available in a compatible release. Please replace it with `PreconnectionBlob` and `ironrdp_core::encode_vec` once Gateway can upgrade to the fixed IronRDP version.

PR #1376, opened 2025-06-11
feat(dgw): allow session tokens to be reused in a limited way

CBenoit:
Add support a configurable reconnection window, allowing token reuse for reconnections after disconnection within a defined period. This feature will enable more robust handling of temporary disconnections, minimizing session disruptions.

The token can be reused under three conditions:

- The associated session was not killed.
- The reconnection window since last disconnection is not exceeded.
- The number of connections during the reconnection window does not exceed 10 (hardcoded value).

For the last point, this is the maximum number of reconnections allowed during the reconnection window. If the reconnection window (e.g.: 30 seconds) is over while the connection is still alive, the counter is reset, and it’s possible to reconnect up to 10 times again. This prevents brute force attacks in the situation where the token is stolen, although that is tricky to exploit in the first place.

Issue: DGW-238


pacmancoder (approved):
LGTM! 🎉

PR #1380, opened 2025-06-16
ci(dependabot): group IronVNC bumps with IronRDP

CBenoit:
IronVNC also depends on a few IronRDP crates, so we may as well group them together.

cc @MathieuMorrissette


MathieuMorrissette (approved):
lgtm

PR #1383, opened 2025-06-16
fix(webapp): incorrect desktop size when a predefined resolution is selected

RRRadicalEdward:
Fix for https://github.com/Devolutions/IronVNC/issues/861.


CBenoit (approved):
LGTM!

You need to run the formater though (see the scripts in the package.json).

PR #1390, opened 2025-06-20
fix(webapp): change toolbar activation height

RRRadicalEdward:
When the webapp is in fullscreen mode, the session toolbar doesn't appear when ф user moves the cursor to the top of the screen. It turned out that for some reason, the minimum cursor height is not 0px, but somewhere around 6px. So, I changed the conditions, and now we're checking that the height is less than 10. No,w the session toolbar appears when the user moves the cursor to the top of the screen.


CBenoit (approved):
Good catch! :100:

PR #1391, opened 2025-06-20
fix(dgw)!: fail-fast on improper certificate for TLS

CBenoit:
Certificates missing the auth extended key usage, or missing a subject alternative name are now rejected:

- immediately fail on startup for certificates from filesystem, and
- fail on certificate resolution for system certificate store.

Issue: DGW-286

### DIFF
--- a/devolutions-gateway/src/tls.rs
+++ b/devolutions-gateway/src/tls.rs
@@ -269,7 +269,7 @@ pub mod windows {
                                     %idx,
                                     serial_number = %report.serial_number,
                                     issues = %report.issues,
-                                    "Filtered out certificate because it has issues"
+                                    "Filtered out certificate because it has significant issues"
                                 );
                                 return None;
                             }


devolutions-gateway/src/tls.rs
```diff
@@ -212,52 +230,54 @@ pub mod windows {
...
+                            );
 
-                            if !has_server_auth_key_purpose {
+                            if skip {
                                 debug!(
-                                    %idx, %serial_number, "Filtered out certificate because it does not have the server auth extended usage"
+                                    %idx,
+                                    serial_number = %report.serial_number,
+                                    issues = %report.issues,
+                                    "Filtered out certificate because it has issues"
```
CBenoit:
```suggestion
                                    "Filtered out certificate because it has significant issues"
```

devolutions-gateway/src/tls.rs
```diff
@@ -343,6 +365,94 @@ pub mod windows {
...
+    let subject = cert.subject_name();
+    let issuer = cert.issuer_name();
+    let not_before = cert.valid_not_before();
+    let not_after = cert.valid_not_after();
+
+    if at < not_before {
+        issues.insert(CertIssues::NOT_YET_VALID);
+    }
+
+    if not_after < at {
```
CBenoit:
```suggestion
    } else if not_after < at {
```

pacmancoder (approved):
LGTM! 🚀

PR #1402, opened 2025-07-01
fix(webapp): update RDP and VNC web clients

RRRadicalEdward:
- Fix an issue in the VNC client where display scaling was not correctly updated after a server-initiated resize.


CBenoit (commented):
To help with tracking the improvements in the changelog, could you list the user-facing changes in the PR body, and change the title as appropriate to match what I did in this PR, please?

https://github.com/Devolutions/devolutions-gateway/pull/1393

(Feel free to use `fix` if it only includes fixes. This is a best effort.)

In this case, we are only interested into user-facing changes, so if nothing is changed in RDP, it’s fine to not mention it (or just qualify it of "minor update").

In this case, I think we added a "resized callback", that is used to fix a bug occurring when resizing the canvas for the VNC client, is that correct?

CBenoit (approved):
LGTM!

CBenoit:
> > In this case, I think we added a "resized callback", that is used to fix a bug occurring when resizing the canvas for the VNC client, is that correct?
> 
> Yes. I improved the PR description. Did I get it right?

Thank you!
I adjusted a little bit: the conventional comment stuff can be used only in the commit summary itself, so the PR title. I used `fix` there, since this bump only includes a bug fix.
I prefer using a list for the details, kind of like a sub-changelog, but without section.

PR #1409, opened 2025-07-01
fix(pedm): add additional context to virtual account code paths

thenextman:
Some users are experiencing an error with the virtual account elevator that I cannot reproduce, and we do not have the necessary contextual information. This PR adds additional context to the virtual account elevator code paths.


crates/devolutions-pedm/src/elevator/virtual_account_elevator.rs
```diff
@@ -26,6 +26,8 @@ use win_api_wrappers::Error;
 
 use super::Elevator;
 
+use anyhow::Context;
```
CBenoit:
```suggestion
use anyhow::Context as _;
```

PR #1423, opened 2025-07-10
fix(ci): include psd1 when zipping the PS module

allan2:
The "Publish PowerShell module" step of _release.yml_ is failing due to

```
Run $Archive = Get-ChildItem -Recurse -Filter "*-ps-*.zip" -File
Remove-Item: Cannot find path
'/home/runner/work/devolutions-gateway/devolutions-gateway/PowerShell/Devolution
sGateway/DevolutionsGateway.psd1' because it does not exist.
Error: Process completed with exit code 1.
```

This PR runs `Compress-Archive` from the parent folder. This is an attempt to include the psd1 in the "Sign PowerShell module contents" step.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -245,9 +245,13 @@ jobs:
           }
 
           Remove-Item $PSModuleZipFilePath -ErrorAction SilentlyContinue | Out-Null
-          Compress-Archive -Path $DGatewayPSModulePath -Destination $PSModuleZipFilePath -CompressionLevel Optimal
 
           $PSModuleParentPath = Split-Path $DGatewayPSModulePath -Parent
+
+          Push-Location $PSModuleParentPath
+          Compress-Archive -Path $DGatewayPSModulePath -Destination $PSModuleZipFilePath -CompressionLevel Optimal
+          Pop-Location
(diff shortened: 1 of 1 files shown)


awakecoding:
hum... how did it work before? can you confirm the resulting module is correct after this fix?

awakecoding:
in any case I downloaded the artifacts from the corresponding workflow run, and the powershell module is complete and loads properly

CBenoit:
Where did you find such a workflow run? I don’t see it

CBenoit:
@allan2 For this kind of changes, you should use the `ci` commit type. [Example](https://github.com/Devolutions/devolutions-gateway/commit/6c414ec2d01f862fc85b80f6a931268b3f881059).

awakecoding:
@CBenoit https://github.com/Devolutions/devolutions-gateway/actions/runs/16184976472

CBenoit:
@awakecoding I think this is the CI workflow run, which is unrelated to the package workflow. I’m double checking the PR on my side right now, but I think it didn’t fix the problem: https://github.com/Devolutions/devolutions-gateway/actions/runs/16198679043

 I’ll look into it.

PR #1438, opened 2025-07-22
minor container fixes

awakecoding:


package/Linux/entrypoint.ps1
```diff
@@ -192,9 +192,10 @@ if ($WebScheme -eq 'https' -and
     $Arguments = @(
         "req", "-x509", "-nodes",
         "-newkey", "rsa:2048",
-        "-keyout", $PrivateKeyFile,
-        "-out", $CertificateFile,
+        "-keyout", $TlsPrivateKeyFile,
+        "-out", $TlsCertificateFile,
         "-subj", "/CN=$Hostname",
+        "-addext", "subjectAltName=DNS:$Hostname",
```
CBenoit, line 198: question: Shouldn’t we also add this?
```
-addext "extendedKeyUsage = serverAuth"
```

PR #1446, opened 2025-07-30
feat(dgw): basic network monitoring

philippelatulippe:
Introduces an endpoint monitor that checks the uptime of a list of hosts provided through a remotely-submitted configuration file. The monitor results are saved to a temporary in-memory buffer and can be fetched by means of a REST endpoint. 

The monitor system is structured as an agent, meant to be driven by a third party (for example our DVLS), so the configuration is ephemeral and monitor results are deleted from the buffer after being fetched. Gateway is not itself the source of truth for the monitor configuration, and it does not persist the log entries.

Two authenticated endpoints are introduced:

* POST /jet/net/monitor/config
* POST /jet/net/monitor/log/drain 

Example requests for testing:

```bash
curl -v http://127.0.0.1:7171/jet/net/monitor/config \
    --json '{"monitors":[{"id":"monitor1","probe":"ping","address":"localhost","interval":10,"timeout":10}]}' \
    -H "Authorization: Bearer $dgwkey"
```
```bash
curl  http://127.0.0.1:7171/jet/net/monitor/log/drain -X POST -H "Authorization: Bearer $dgwkey"
```

Remaining tasks for a basic scanner:
* We should parse the ICMP responses
* Better structuring of errors so that we can have error codes in the JSON, instead of arbitrary strings
* Upper bound on the number of log entries stored in the VecQueue

Issue: DGW-302


devolutions-gateway/src/lib.rs
```diff
@@ -54,6 +57,7 @@ pub struct DgwState {
     pub recordings: recording::RecordingMessageSender,
     pub job_queue_handle: job_queue::JobQueueHandle,
     pub credential_store: credential::CredentialStoreHandle,
+    pub monitoring_state: Arc<network_monitor::State>
```
philippelatulippe: I avoided my reflex to implement all functionality inside of an object, and instead have an opaque state object that is passed to my functions, since that seemed to be what was done here. Does that make sense?

crates/network-monitor/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+
+    let (new_cancellation_tokens, new_monitors): (
+        Vec<(String, CancellationToken)>,
+        Vec<Pin<Box<dyn Future<Output = ()> + Send>>>,
+    ) = added
+        .map(|definition| {
+            let cancellation_token = CancellationToken::new();
+            let cancellation_monitor = cancellation_token.clone();
+
+            let definition_clone = (*definition).clone(); // TODO: is there a nicer way to do this?
```
philippelatulippe: If I remember right, I have to clone `definition` here because its owner will not outlive the end of the function. But I suspect there might be a way to do `map` so that `definition` is already a clone?

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -0,0 +1,49 @@
...
+
+pub fn make_router<S>(state: DgwState) -> Router<S> {
+    let router = Router::new()
+        .route("/config", routing::post(handle_set_monitoring_config))
+        .route("/log/drain", routing::post(handle_drain_log));
+
+    router.with_state(state)
+}
+
+#[debug_handler]
```
CBenoit: issue: This macro is only for debugging purposes, please remove it and the `macros` feature

devolutions-gateway/Cargo.toml
```diff
@@ -115,6 +116,7 @@ etherparse = "0.15"
 
 # For KDC proxy
 portpicker = "0.1"
+network-monitor = { version = "0.1.0", path = "../crates/network-monitor" }
```
CBenoit: style: Move to the "In-house" section (this is unrelated to the KDC proxy)

devolutions-gateway/Cargo.toml
```diff
@@ -86,7 +87,7 @@ ngrok = "0.13"
 # HTTP
 hyper = "1.6"
 hyper-util = { version = "0.1", features = ["tokio", "server", "server-auto"] }
-axum = { version = "0.8", default-features = false, features = ["http1", "json", "ws", "query", "tracing", "tower-log"] }
+axum = { version = "0.8", default-features = false, features = ["http1", "json", "ws", "query", "tracing", "tower-log", "macros"] }
```
CBenoit: issue: This feature is only for development purposes, let’s not include it in prod.
```suggestion
axum = { version = "0.8", default-features = false, features = ["http1", "json", "ws", "query", "tracing", "tower-log"] }
```

crates/network-monitor/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+
+            let monitor = async move {
+                loop {
+                    let start_time = Utc::now();
+
+                    let monitor_result = match definition_clone.probe {
+                        ProbeType::Ping => do_ping_monitor(&definition_clone).await,
+                        ProbeType::TcpOpen => do_tcpopen_monitor(&definition_clone).await,
+                        ProbeType::Unknown(_) => return, // TODO: shouldn't happen, they should be filtered out. Create a separate ProbeType enum without Unknown?
+                    };
```
philippelatulippe: When parsing the config, I have an `Unknown` case in the enum for the sake of forward-compatibility. I'd like to filter these out at the start of this function, in order to avoid dealing with the Unknown case here. Can I do that without having to manually write a duplicate of the `ProbeType` enum? Can I start with an unknown with no unknown case, and then derive a version that has it?

devolutions-gateway/src/lib.rs
```diff
@@ -77,6 +81,13 @@ impl DgwState {
         let (shutdown_handle, shutdown_signal) = devolutions_gateway_task::ShutdownHandle::new();
         let (job_queue_handle, job_queue_rx) = job_queue::JobQueueHandle::new();
         let credential_store = credential::CredentialStoreHandle::new();
+        
+        let temp_dir = TempDir::new("dgw-network-monitor-test").expect("Could not create temp dir");
+        let temp_path: Utf8PathBuf = Utf8Path::from_path(temp_dir.path())
+            .expect("TempDir gave us a garbage path")
+            .to_path_buf();
+        
+        let monitoring_state = Arc::new(network_monitor::State::mock(temp_path));
```
CBenoit: thought: I would prefer we do not include an extra dependency (tempdir) just for mocking (testing). I’ll come back on this later. (EDIT: it would be fine as a "dev-dependency", but it’s not one right now)

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -0,0 +1,49 @@
...
+async fn handle_drain_log(
+    extract::State(DgwState { monitoring_state, .. }): extract::State<DgwState>
+) -> Result<Json<MonitoringLogResponse>, http::StatusCode> {
+    Ok(Json(
+        MonitoringLogResponse {
+            entries: network_monitor::drain_log(monitoring_state).into()
+        }
+    ))
```
CBenoit: suggestion: Not faillible, so you don’t need to return a `Result`

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -0,0 +1,49 @@
...
+    Json(config): Json<network_monitor::MonitorsConfig>,
+) -> Result<(), http::StatusCode> {
+    network_monitor::set_config(config, monitoring_state).await.map_err (| err |  {
+        // TODO: no side effects in a map please. move the match to a From impl?
+        error!(error = format!("{err:#}"), "Failed to set up network monitoring");
+        match err {
+            SetConfigError::Io(_) => http::StatusCode::INTERNAL_SERVER_ERROR,
+            SetConfigError::Serde(_) => http::StatusCode::INTERNAL_SERVER_ERROR
+        }
+    })
```
CBenoit: suggestion: Use `crate::http::HttpError` instead.
https://github.com/Devolutions/devolutions-gateway/blob/25213d1e7d384df7b3551abb0f67f4b5809d6b74/devolutions-gateway/src/api/jrl.rs#L37-L41

This will solve your TODO as well, the returned error is already logged automatically in the `IntoResponse` of `HttpError`

devolutions-gateway/src/api/monitoring.rs
philippelatulippe: Guides for axum recommended always adding this macro to get proper compiler diagnostic messages. It seems convenient to keep those messages on?

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -0,0 +1,49 @@
...
+pub fn make_router<S>(state: DgwState) -> Router<S> {
+    let router = Router::new()
+        .route("/config", routing::post(handle_set_monitoring_config))
+        .route("/log/drain", routing::post(handle_drain_log));
+
+    router.with_state(state)
+}
+
+#[debug_handler]
+async fn handle_set_monitoring_config(
```
CBenoit, line 36: suggestion: Please, add the openapi/utoipa documentation. You can check the `crate::api::session` module or another one, for reference.

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -0,0 +1,49 @@
...
+) -> Result<Json<MonitoringLogResponse>, http::StatusCode> {
+    Ok(Json(
+        MonitoringLogResponse {
+            entries: network_monitor::drain_log(monitoring_state).into()
+        }
+    ))
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct MonitoringLogResponse {
```
CBenoit: issue: Should not be part of the public API.

```suggestion
pub(crate) struct MonitoringLogResponse {
```

(Or private, but you may need that in the openapi declaration.)

CBenoit (commented):
Good job! I just did a very high level review, I’ll look at the details of your crate later 🙂

crates/network-monitor/src/state.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub fn new(cache_path: Utf8PathBuf) -> State {
+        State {
+            cache_path: cache_path,
+            log: LogQueue::new(),
+            config: RwLock::new(MonitorsConfig::empty()),
+            cancellation_tokens: Mutex::new(HashMap::new()),
+        }
+    }
+
+    pub fn mock(cache_path: Utf8PathBuf) -> State {
```
allan2, line 30: Add `#[cfg(test)]`.

crates/network-monitor/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+pub struct MonitorsConfig {
+    monitors: Vec<MonitorDefinition>,
+}
+
+impl MonitorsConfig {
+    fn empty() -> MonitorsConfig {
+        MonitorsConfig { monitors: Vec::new() }
+    }
+
+    fn mock() -> MonitorsConfig {
```
allan2, line 182: Add `#[cfg(test)]`.

crates/network-monitor/src/lib.rs
allan2: You can clone before collecting.

```rs
pub async fn set_config(config: MonitorsConfig, state: Arc<State>) -> Result<(), SetConfigError> {
    let file = fs::OpenOptions::new()
        .create(true)
        .write(true)
        .truncate(true)
        .open(get_monitor_config_path(&state.cache_path))?;

    let mut config_write = state.config.write().await;

    serde_json::to_writer_pretty(&file, &config)?;

    let old_config = mem::replace(&mut *config_write, config);

    // clone here
    let new_config_set = config_write.monitors.clone().into_iter().collect::<HashSet<_>>();
    let old_config_set = old_config.monitors.clone().into_iter().collect::<HashSet<_>>();

    // `Iterator::cloned` here
    let added = new_config_set.difference(&old_config_set).cloned();
    let deleted = old_config_set.difference(&new_config_set);

    let (new_cancellation_tokens, new_monitors): (Vec<_>, Vec<_>) = added
        .into_iter()
        .map(|definition| {
            let cancellation_token = CancellationToken::new();
            let cancellation_monitor = cancellation_token.clone();

            // my personal preference is to use `Arc::clone`
            // https://rust-lang.github.io/rust-clippy/stable/index.html#/clone_on_ref_ptr
            let state = Arc::clone(&state);
            // clone id before moving definition into the async block
            let definition_id = definition.id.clone();
            let monitor = async move {
                loop {
                    let start_time = Utc::now();

                    let monitor_result = match &definition.probe {
                        ProbeType::Ping => do_ping_monitor(&definition).await,
                        ProbeType::TcpOpen => do_tcpopen_monitor(&definition).await,
                        ProbeType::Unknown(_) => return, // TODO: shouldn't happen, they should be filtered out. Create a separate ProbeType enum without Unknown?
                    };

                    state.log.write(monitor_result);

                    let elapsed = Utc::now() - start_time;
                    let next_run_in = (definition.interval as f64 - elapsed.as_seconds_f64()).clamp(1.0, f64::INFINITY);
                    select! {
                        _ = cancellation_monitor.cancelled() => { () }
                        _ = tokio::time::sleep(Duration::from_secs_f64(next_run_in)) => { () }
                    }
                }
            };
            (
                (definition_id, cancellation_token),
                Box::pin(monitor) as Pin<Box<dyn Future<Output = ()> + Send>>,
            )
        })
        .unzip();

    let mut cancellation_tokens_write = state.cancellation_tokens.lock().await;

    for definition in deleted {
        cancellation_tokens_write[&definition.id].cancel();
        cancellation_tokens_write.remove(&definition.id);
    }

    for (monitor_id, cancellation_token) in new_cancellation_tokens {
        cancellation_tokens_write.insert(monitor_id, cancellation_token);
    }

    for monitoring_task in new_monitors {
        tokio::spawn(monitoring_task);
    }

    Ok(())
}
```

crates/network-monitor/src/lib.rs
allan2: Maybe remove the `Unknown` variant?

If it shouldn't happen, you can handle the error when deserializing.

devolutions-gateway/src/lib.rs
allan2: It seems ok to me. Should the network monitor be feature gated?

crates/network-monitor/src/log_queue.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    pub(crate) fn drain(&self) -> VecDeque<T> {
+        let mut entries = self.entries.lock().expect("Couldn't get LogQueue lock");
+
+        return mem::replace(&mut entries, VecDeque::new());
+    }
+}
+
+#[cfg(test)]
+mod test {
```
allan2: Typically `tests`

crates/network-monitor/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "network-monitor"
+version = "0.1.0"
+edition = "2021"
```
allan2: You can copy the Cargo.toml of other crates.

Version `0.0.0`, `publish = false`, `edition = 2024`, and other fields.

crates/network-monitor/src/lib.rs
philippelatulippe: It will happen as we introduce new monitor types. In the long-term I'd like gateway to return some json-structured warning the client that some if its monitor types weren't supported, so I do want them to be represented in the array of monitors. But after filtering them out I feel like the unknown case should be gone so that I don't have to handle it in my `match`. Maybe I can compose my ProbeType enum somehow?

devolutions-gateway/src/api/monitoring.rs
CBenoit: I remember that it was stated to be only intended for debugging and should not be used in production to avoid the compile-time overhead, and losing other optimizations opportunities.
It is nice to get better diagnostics, but it’s also a typical write once thing, and we’ll be compiling again and again, so I still stand for removing it once you get the correct signature.

CBenoit: I see in the doc that now it does not affect the runtime performance when built with --release, so my info may be outdated, but I would still prefer we don’t make the builds heavier

crates/network-monitor/src/lib.rs
philippelatulippe: Hm, what is the magic that would make this work? I've called `cloned()` with a _d_, which I hadn't noticed existed probably due to how close it looks like just plain `clone`, but once I access `definition` in the `async move` closure, then I get a plethora of borrow errors. Is there a change in your code that I'm missing?

allan2: The function in my last comment should work as is.

crates/network-monitor/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    Ok(())
+}
+
+async fn do_ping_monitor(definition: &MonitorDefinition) -> MonitorResult {
+    let start_time = Utc::now();
+
+    let ping_result = async || -> anyhow::Result<TimeDelta> {
+        let runtime = network_scanner_net::runtime::Socket2Runtime::new(None)?;
```
irvingoujAtDevolution: Start this runtime is a bit expensive, I think it's better to put it outside of a single request, but keep it alive when a monitoring is requested, and reuse it from place to place. It's an Arc, so you can safely pass it around as a parameter with `Arc::clone` if necessary

devolutions-gateway/src/lib.rs
CBenoit: If you mean adding a compile-time feature flag, then I would suggest we don’t. If you suggest to hide the feature behind the unstable flag, then I agree we could do that so we can tweak the API across the minor releases.

allan2: I did mean compile-time feature flag. I agree with what you are saying now.

crates/network-scanner-proto/src/icmp_v6.rs
```diff
@@ -0,0 +1,139 @@
...
+        payload: Vec<u8>,
+    },
+    EchoReply {
+        identifier: u16,
+        sequence_number: u16,
+        payload: Vec<u8>,
+    },
+}
+
+impl Into<Vec<u8>> for Icmpv6Message {
```
CBenoit: suggestion: Prefer implementing `From<Icmpv6Message> for Vec<u8>` when feasible (as in this case). This provides a more idiomatic interface and enables `.into()` automatically via the blanket `Into` implementation.

issue/style: However, using `From`/`Into` for binary encoding or parsing is considered unidiomatic. These traits are meant to express *type conversions*, not serialization or protocol-specific encodings. Encoding logic belongs in explicitly-named methods or dedicated traits (e.g., `encode()` or `to_bytes()`), which make the semantics clearer.

CBenoit (commented):
note: While `.into()` is concise, I strongly discourage its use in non-generic, concrete code because it obscures intent. Prefer `TargetType::from(source_value)`:

* The destination type is explicit at the call site.
* It immediately signals that a conversion is taking place and to which type.
* It removes the need for type inference and reduces coupling to surrounding context, aiding readability and future refactoring.

`.into()` is better reserved for generic contexts or builder-style code where the target type is already constrained. Outside of those, the clarity benefit of `from()` is worth the extra characters.

CBenoit (commented):
Highlighting this comment: https://github.com/Devolutions/devolutions-gateway/pull/1446/files#r2242524595

suggestion: Look in the `config` module, there is a `enable_unstable` option. For now, I suggest we gate this feature behind this option, so you can freely change the API as necessary until the 2025.3 release.

crates/network-monitor/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+use std::future::Future;
+use std::io;
+use std::mem;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::time::Duration;
+
+use camino::*;
+use chrono::{DateTime, TimeDelta, Utc};
+use serde::*;
```
CBenoit: issue: I highly discourage using glob imports. When done without precise conventions it leads to code harder to review and explore. There are some exceptions, but we should not decide in an ad-hoc way.

crates/network-monitor/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+#[derive(PartialEq, Clone, Serialize, Deserialize, Debug)]
+pub struct MonitorResult {
+    monitor_id: String,
+    request_start_time: DateTime<Utc>,
+    response_success: bool,
+    response_messages: Option<String>,
+    response_time: f64,
+}
```
CBenoit: note: So far, we’ve been separating serialization (models) from the actual business logic. In practice, this means that we don’t use `serde` in the "library" crates, and we have separate types for the internal representations, and the "Gateway API boundary representation" ("DTO"). The idea is that the crate should be used like any other library, and should not take decision on how the data is represented between the systems (e.g.: DVLS <-> Gateway) at this level.

crates/network-monitor/src/log_queue.rs
```diff
@@ -0,0 +1,60 @@
...
+
+impl<T> LogQueue<T> {
+    pub(crate) fn new() -> Self {
+        LogQueue {
+            entries: Mutex::new(VecDeque::new()),
+        }
+    }
+
+    pub(crate) fn write(&self, data: T) {
+        let mut entries = self.entries.lock().expect("Couldn't get LogQueue lock");
```
CBenoit: nitpick: Rust errors should not be capitalized. Also, I suggest to stay concise for this specific failure point, explaining we got a poisoned mutex (highly unlikely, since we should not panic while holding the lock as far as I can see).

```suggestion
        let mut entries = self.entries.lock().expect("poisoned");
```

crates/network-monitor/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+}
+
+pub fn drain_log(state: Arc<State>) -> VecDeque<MonitorResult> {
+    return state.log.drain();
+}
+
+fn get_monitor_config_path(cache_path: &Utf8PathBuf) -> Utf8PathBuf {
+    cache_path.join("monitors_cache.json")
+}
```
CBenoit: issue: The current logic assumes that cache_path is a directory and appends a hardcoded filename internally. This creates ambiguity and limits flexibility, as the consumer cannot control the final file path or name. It also introduces hidden behavior that is not obvious to user of the API.

suggestion: This logic should be delegated to the consumer. The library should treat cache_path as the full path to the cache file, allowing consumers to specify both the location and the name.

rationale: Since the library only manages a single cache file, there's no strong reason to impose a fixed internal filename. This change would improve API clarity and make the library more adaptable to various use cases.

crates/network-monitor/src/lib.rs
CBenoit: Yes, I think you could have two different enums for that…!

```rust
#[derive(Deserialize)]
#[serde(untagged)]
enum ProbeType {
    Known(KnownProbeType),
    Unknown(String),
}

#[derive(Deserialize)]
#[serde(rename_all = "snake_case")]
enum KnownProbeType {
    Ping,
    TcpOpen,
}
```

This cleanly separates the supported values (`KnownProbeType`) from fallbacks (`Unknown(String)`), letting `serde` match any unrecognized string to the `Unknown` variant.

I’ve used a similar pattern in the token module: https://github.com/Devolutions/devolutions-gateway/blob/25213d1e7d384df7b3551abb0f67f4b5809d6b74/devolutions-gateway/src/token.rs#L141-L146

But, for maximum flexibility, I think you should **defer the parsing**:

- Deserialize into a more permissive intermediate type (e.g. `String` or `serde_json::Value`)
- Convert to a strongly typed value using serde_json::from_value (or a manual match)
- Gracefully handle invalid or unknown values (log them at debug! level and proceed)

This pattern helps keep your data models lenient at the edges while maintaining strong typing internally, which seems to be exactly what you need.

In fact, I’ve preferred this approach in similar contexts, I think it often results in cleaner separation between raw input and trusted data.
You might find some inspiration in how the preflight API module handles this:
https://github.com/Devolutions/devolutions-gateway/blob/master/devolutions-gateway/src/api/preflight.rs

CBenoit: > It will happen as we introduce new monitor types. In the long-term I'd like gateway to return some json-structured warning the client that some if its monitor types weren't supported, so I do want them to be represented in the array of monitors.

By the way, the preflight API does that too.

Please have a look at the cookbook as well: https://github.com/Devolutions/devolutions-gateway/blob/master/docs/COOKBOOK.md#preflight-api

(I would appreciate if you could add a section for the monitor API there as well!)

crates/network-monitor/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+devolutions-agent-shared = { path = "../devolutions-agent-shared" }
+network-scanner = { path = "../network-scanner" }
+network-scanner-net = { path = "../network-scanner-net" }
+serde = { version = "1", features = ["derive"] }
+camino = { version = "1", features = ["serde1"] }
+serde_json = "1"
+thiserror = "1"
+chrono = { version = "0.4", features = ["serde"] }
+tokio = { version = "1.45", features= ["full"] }
+tokio-util = { version = "0.7", features = ["full"] }
```
CBenoit: issue: Do not use `"full"`, especially in libraries, this brings in new dependencies and code we don’t need.

devolutions-gateway/src/api/monitoring.rs
philippelatulippe: That's very cool! ~~Some of my structs are in my network-monitor crate, so they need to have the `ToSchema` trait. How should I proceed, should my crate also depend on utoipa? Should that dependency be gated behind a feature? If yes then how is the flag passed from the main application to the crates?~~ I've just read some of your other comments which render this issue moot

crates/network-monitor/src/state.rs
philippelatulippe: This mock is created by the mock function in `DgwState`, which is not gated in this way. I'm not sure why though, maybe it should be? @CBenoit

crates/network-monitor/Cargo.toml
philippelatulippe: I did notice that `"full"` is avoided in the codebase, but I wanted to ask: it feels tedious to toggle features with such granularity, why can't the compiler figure out what we're not using? I'm genuinely asking, probably I'm about to find out that the compiler is not that smart and that by doing this I'm going to murder our compile time due to monomorphisation.

philippelatulippe: But in the case of tokio, maybe we can enable "full"? Across our crates we currently enable:

fs, io-std, io-util, macros, net, parking_lot, process, rt, rt-multi-thread, signal, sync, time

I think that's every feature enabled by "full"!

allan2: Consumers won't be including every crate in the workspace though.

I think you only need `macros` for `network-monitor`.

crates/network-monitor/src/lib.rs
philippelatulippe: I agree completely! My original code was reading the config dir path from the existing env variable, but I didn't rethink the implementation when I switched to passing the path through the state struct.

crates/network-monitor/src/lib.rs
philippelatulippe: That sounds good. My thought process was to keep everything tightly contained in my monitoring crate, but I see how keeping the DTOs in my crate goes against the current architecture. However I'm not sure how to proceed other than duplicating the structs and enums, would that be okay?

crates/network-monitor/src/state.rs
CBenoit: The one in Devolutions Gateway is #[doc(hidden)] to express a similar intention, but if we gate with #[cfg(test)], then the function is not available for the integration tests…
It’s not ideal though, I don’t really like that. But for the purpose of getting stuff done, I would say that it’s not a big deal, and we can surely do something about that later

crates/network-monitor/Cargo.toml
CBenoit: That’s a fair point for tokio, but as a good practice I still encourage you to specify only what is needed.
Allan point also holds true: all the features are enabled for Jetsocat I think, but not necessarily for all the other binaries.

Also, for tokio-util you are adding extra features which _are_ bringing extra crates into the dependency graph.

Of course, the compiler, and more specifically the linker is able to remove most of the dead code.
Typically, it’s not really a problem about what is in the final artifacts we deliver.

However:
- I’m pretty sure the code still gets compiled first, before its removed by the linker
- For each build from scratch, we need to download more and more crates
- From scratch builds are done way more often than you may think (especially in the CI)
- Our from scratch build time is indeed already not ideal today, and adding a few seconds to the mess may seem like not much, but that’s part of the reason why the builds are getting significant longer on the long term
- I still have hopes to improve that someday, so this is saving time for future me
- As an aside, we also track ALL the dependencies for each releases of the Gateway (SBOM), because we care about security issues found in the dependencies, and even though the situation can already be considered not ideal (many, many dependencies in there), I still prefer minimizing as much as possible the risk of being alerted about a vulnerability or a supply chain attack in the one crate we didn’t actually need

I do think that the feature "full" is very useful though, especially when you’re prototyping or exploring and you don’t know exactly the feature set you need yet.

crates/network-monitor/src/lib.rs
CBenoit: I think that’s okay! They may or may not diverge at some point in the future based on our needs; the benefit is that you can refactor the internal structs without affecting the system API by mistake, and you can modify the system API without altering the internal details. IIRC, the network scanner went through something like this.

devolutions-gateway/openapi/gateway-api.yaml
CBenoit, line 1: Something seems wrong here. Let’s update the OpenAPI stuff in a separate PR.

devolutions-gateway/src/api/mod.rs
```diff
@@ -42,5 +41,9 @@ pub fn make_router<S>(state: crate::DgwState) -> axum::Router<S> {
         );
     }
 
+    if state.conf_handle.get_conf().enable_network_monitoring {
```
CBenoit: suggestion: Instead of adding a new configuration key, please use `get_conf().debug.enable_unstable`. This is the generic flag we use for unstable stuff. For stabilizing, we just drop the related if blocks.

devolutions-gateway/src/api/mod.rs
```diff
@@ -42,5 +41,9 @@ pub fn make_router<S>(state: crate::DgwState) -> axum::Router<S> {
...
+        router = router.nest("/jet/monitoring", monitoring::make_router(state.clone()));
```
CBenoit: naming: For the API naming, what about `/jet/monitor`? This seems more aligned with the rest of the routes.

question: Would it make sense to put that under the `/jet/net` endpoint? E.g.: `/jet/net/monitor`? Or is it also intended to do something else outside of network monitoring?

devolutions-gateway/src/openapi.rs
CBenoit, line 1: note: I’m okay with adjusting the small details with OpenAPI in a follow up PR. It’s fine to just brush up the documentation. It’s something hard to get right on the first try.

CBenoit:
note: If you have a Jira ticket, add a `Issue: DGW-XXX` footer in the main comment that we are using for changelog generation :slightly_smiling_face:

crates/network-monitor/src/lib.rs
philippelatulippe: Thanks, I got it working. I had missed that you were using the clone**d** function, which is the key here

crates/network-monitor/src/lib.rs
philippelatulippe: I've went ahead and separated the internal structs from the DTO, and now my issues are resolved. I also added a cookbook section, check it out!

Edit: I didn't defer the parsing of the enum, since I still wanted the type to express exactly which values are supported. I think that's important since these end up in our openapi doc

crates/network-scanner-proto/src/icmp_v6.rs
philippelatulippe: Thanks for the advice. I'd seen some libraries making maximal use of From/Into and I thought it was a bit odd, but I wanted to join in on the fun! ;) I switched it to a simple `encode()` method without a trait.

devolutions-gateway/openapi/gateway-api.yaml
philippelatulippe: Should I just revert the yaml file, or should I also comment out my changes that causes these to be included in the output?

devolutions-gateway/src/api/mod.rs
philippelatulippe: Sorry I forgot about your previous comment, the new configuration key is gone now

devolutions-gateway/src/api/mod.rs
philippelatulippe: They are indeed closely related, I moved the path to `jet/net/monitor`

crates/network-monitor/src/lib.rs
philippelatulippe: @irvingoujAtDevolution Thanks for the type, I now create the runtime only once by using a `LazyLock`

devolutions-gateway/openapi/gateway-api.yaml
CBenoit: You can just revert the yaml and keep the rest :+1:

docs/COOKBOOK.md
```diff
@@ -270,6 +270,36 @@ And here is how the response may look like:
 ]
 ```
 
+## Server monitoring API
```
CBenoit: suggestion (non-blocking): Maybe easier to understand with this wording?
```suggestion
## Network monitoring API
```

docs/COOKBOOK.md
```diff
@@ -270,6 +270,36 @@ And here is how the response may look like:
...
+```
+
+The monitor will start immediately. Calling this API again will overwrite the configuration, stopping any 
+monitors no longer present. A body is returned which may contain a list of monitors that could not be started 
+due their type (set in the field `probe`) being unsupported.
+
+Retrieve the logs:
+
+```shell
+curl -v http://127.0.0.1:7171/jet/net/monitor/log/drain -X POST -H "Authorization: Bearer $dgwkey"
```
CBenoit, line 291: praise: I see you are using the correct methods 💯

docs/COOKBOOK.md
```diff
@@ -270,6 +270,36 @@ And here is how the response may look like:
...
+```
+
+The response will look similar to this:
+
+```json
+{"entries":[{"monitor_id":"monitor1","request_start_time":"2025-08-22T17:07:34.3370521Z","response_success":true,"response_messages":null,"response_time":0.0585181}]}
```
CBenoit: suggestion: Maybe use skip_serializing_if = Option::is_none in the serde annotations, we typically don’t bother returning null values in the rest of the API

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -0,0 +1,222 @@
...
+        .route("/log/drain", routing::post(handle_drain_log));
+
+    router.with_state(state)
+}
+
+/// Replaces the existing monitoring config with the one provided in the body.
+/// This request will immediately start any new monitors, and will stop
+/// currently active monitors that are no longer in the config.
+///
+/// The configuration is not persisted across restarts.
```
CBenoit: suggestion: Here is a wording I find slightly easier to understand. (It’s on me, but I was confused about "will immediately start any new monitors" on first read.

```suggestion
/// Replace the current monitoring configuration with the configuration in the request body.
///
/// Changes take effect immediately:
/// - Starts any monitors newly defined in the payload.
/// - Stops any currently running monitors that are omitted from the payload.
///
/// Note: The configuration is not persisted across process restarts.
```

CBenoit: question: By the way, I kind of remember you talked about that in the past, but I’m not entirely sure: is it really fine to not persist the configuration? I guess the idea is that if the Gateway is restarted, the monitor order must be reissued by DVLS?

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -0,0 +1,222 @@
...
+#[cfg_attr(feature = "openapi", utoipa::path(
+    post,
+    operation_id = "SetMonitoringConfig",
+    tag = "NetworkMonitoring",
+    path = "/jet/net/monitor//config",
```
CBenoit: nitpick:
```suggestion
    path = "/jet/net/monitor/config",
```

devolutions-gateway/src/api/monitoring.rs
philippelatulippe: Good point, the original paragraph mentions "new monitors" without being clear that we are talking about the changes in the config. Thanks for the new wording!

philippelatulippe: That's the idea, given that the controlling server needs to be in regular contact with gateway, we might as well keep it stateless. But I've been meaning to ask you: does DVLS have an explicit way to know when the gateway comes online? It's not clear to me how these "Subscribers" work. If not I would have to rethink the approach.

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -0,0 +1,224 @@
...
+async fn handle_set_monitoring_config(
+    extract::State(DgwState { monitoring_state, .. }): extract::State<DgwState>,
+    Json(config): Json<MonitorsConfig>,
+) -> Result<Json<SetConfigResponse>, HttpError> {
+    let (processed_config, probe_type_errors) = config.lossy_into();
+
+    network_monitor::set_config(processed_config, monitoring_state)
+        .await
+        .map(|_| Json(SetConfigResponse::new(probe_type_errors)))
+        .map_err(
```
CBenoit, line 45: style: https://github.com/Devolutions/IronRDP/blob/6fddcaae386cddb32c4bcc3a349bc07308c65a59/STYLE.md#formatting-of-error-messages
```suggestion
                .with_msg("failed to set up network monitoring")
```

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -0,0 +1,224 @@
...
+            ),
+            |mut partitions, conversion_result| {
+                match conversion_result {
+                    Ok(value) => partitions.0.push(value),
+                    Err(error) => partitions.1.push(error),
+                };
+
+                return partitions;
+            },
+        );
```
CBenoit, line 102: style: Probably linted by clippy

```suggestion
        let config = network_monitor::MonitorsConfig { monitors };
```

devolutions-gateway/src/api/monitoring.rs
CBenoit: The subscriber API is only used by Hub Business, so you can ignore all of that when it comes to DVLS. DVLS performs active polling in order to maintain a "one-directed communication" (Gateway inbound and DVLS outbound only). DVLS may never know that a Devolutions Gateway was restarted. Does that change the design in your mind?

philippelatulippe: Thanks for the deets, I'll bring back my code to save the config to disk in a follow-up PR then.

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -0,0 +1,226 @@
...
+
+        let config = network_monitor::MonitorsConfig { monitors };
+
+        return (config, errors);
+    }
+}
+
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
+#[derive(Eq, PartialEq, Hash, Clone, Serialize, Deserialize, Debug)]
```
CBenoit, line 111: suggestion: Everything in there can probably be `pub(crate)`

devolutions-gateway/openapi/gateway-api.yaml
CBenoit: Don’t forget to revert before we merge 🙂

crates/network-monitor/src/state.rs
```diff
@@ -0,0 +1,38 @@
...
+        State {
+            cache_path: cache_path,
+            log: LogQueue::new(),
+            config: RwLock::new(MonitorsConfig::empty()),
+            cancellation_tokens: Mutex::new(HashMap::new()),
+            scanner_runtime: LazyLock::new(|| Socket2Runtime::new(None)),
+        }
+    }
+
+    pub fn mock(cache_path: Utf8PathBuf) -> State {
```
CBenoit, line 30: Add `#[doc(hidden)]`

crates/network-monitor/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+            Duration::from_secs(definition.timeout),
+        )
+        .await?;
+        // TODO: send more than 1 ping packet
+
+        Ok(UtcDateTime::now() - start_time)
+    }()
+    .await;
+
+    return match ping_result {
```
CBenoit: style: unneeded return

crates/network-monitor/src/state.rs
```diff
@@ -0,0 +1,38 @@
+use std::sync::LazyLock;
+
+use tokio::sync::{Mutex, RwLock};
```
CBenoit: suggestion: Use either `std::sync::{Mutex, RwLock}` or `parking_lot::{Mutex, RwLock}` (better, but does not change much in this low contention context)

rationale: Don’t use async Mutex when
- The lock does not need to be held across await points.
- The duration of any single lock is near-instant.

CBenoit: Going further: do you really need to use a RwLock? Under low contention, it’s better to stick to a plain Mutex (and I don’t see where you use it read-only)

crates/network-monitor/src/lib.rs
philippelatulippe: Could I keep the `return` for larger methods and larger return statements? I feel that if you remove the statement here, then it's not clear you've reached the end of the method until you scroll down to the its last line

crates/network-monitor/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+        .open(&state.cache_path)?;
+
+    let mut config_write = state.config.write().await;
+
+    serde_json::to_writer_pretty(&file, &config)?;
+
+    let old_config = mem::replace(&mut *config_write, config);
+
+    let new_config_set: HashSet<MonitorDefinition> = config_write.monitors.clone().into_iter().collect();
+    let old_config_set: HashSet<MonitorDefinition> = old_config.monitors.clone().into_iter().collect();
```
CBenoit: issue: Drop the lock as soon as you are done with it, so that it is released sooner.

crates/network-monitor/src/lib.rs
CBenoit: I totally get the desire for an explicit `return` as a visual "we’re done" marker, especially coming from C/C++/C#/Java. In Rust though, the common convention is to reserve `return` for early exits and let the last expression be the value. Using a trailing expression tends to read well once you’re used to it, and tools like Clippy even warn on a final `return` as needless by default.

For long tail expressions, if you prefer, you can still keep the end obvious without adding `return` by binding then tailing:

```rust
let result = match ping_result {
    Ok(time) => MonitorResult { /* ... */ },
    Err(error) => MonitorResult { /* ... */ },
};

result
```

In fact, since this is the convention in Rust, I’m afraid going against could harm clarity more than the opposite (expecting an early exit).

crates/network-monitor/src/state.rs
philippelatulippe: I'm curious about this, the [tokio docs for their mutex](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html) at once states that it's fine to use the std mutex in an async context, but also states that the std mutex will block threads (which I assume could lead to threadpool exhaustion), and that awaiting while holding the std mutex's lock can easily lead to deadlocks. Should I not prefer correctness over a slight performance improvement?

CBenoit: That’s exactly my point, but did I miss a place where you are using it across await points? I don’t think you really need to hold the lock across an await point ever though. Do you think the lock is held for a prolonged period too?

CBenoit: Note that a task should not be running for more than approximately 200ms (rule of thumb) before it yields back to the executor, anything running faster is considered "nearly instant" enough. I don’t see any heavy processing on the Mutex in your code, so I’m pretty sure that all your stuff is running in a blink of an eye.

philippelatulippe: It's true, everything done while holding the lock here is fast and there shouldn't be any contention unless something inexplicably spams the set_config endpoint. I was asking because the documentation had surprised me. Shouldn't the first choice be safe code, rather than a tiny performance gain? I can imagine in a more complex example, a developer might add an await point, or add different functions that lock the mutex, and suddenly these problems show up, undetected by the compiler.

CBenoit: Actually, you are not the only one who had this thought, and there is a lint for that (clippy::await_holding_lock), so you would get a warning if that were to happen :slightly_smiling_face: 
The Tokio docs recommend the std mutex first not to chase a micro-optimization, but because the async mutex buys you one specific semantic: the ability to hold the lock across an .await. If you don’t need that capability, the std mutex is simpler and cheaper, and tends to produce clearer, more predictable code. It’s about choosing the simpler tool unless you truly need to await while holding the lock (but such code is also prone to deadlocks, in its own way!). If that need appears (now or in a future refactor), the general approach is to then switch the guard to tokio::sync::Mutex or move to an actor pattern.

So, my recommendation here is to use the std or parking_lot one, but:
- Keep the critical section as tiny as possible.
- Maybe wrap it into non-async helper methods as suggested in Tokio docs, this ensures you never run into this problem, and I used this pattern in a lot of places already.

philippelatulippe: Alright, I switched to the std lock now

crates/network-monitor/src/lib.rs
philippelatulippe: Done

crates/network-monitor/src/lib.rs
philippelatulippe: Done

crates/network-monitor/src/state.rs
philippelatulippe: Thanks for reminding me, I remember you'd suggested it in a previous discussion

devolutions-gateway/src/api/monitoring.rs
philippelatulippe: Done

devolutions-gateway/src/api/monitoring.rs
philippelatulippe: Fixed. Github is look at the wrong line so it can't see the fix

devolutions-gateway/src/api/monitoring.rs
philippelatulippe: fixed

PR #1448, opened 2025-07-31
chore(webapp): update to Angular 20 and PrimeNG 20

allan2:
This updates all deps in the webapp.

- Sass imports are refactored to `@use` due to `@import` deprecation in dart-sass. Namespaces are explicitly set, .e.g., `variables.$borderRadius` 
- Angular components are all standalone
- PrimeNG components are replaced with v20 versions using the [migration guide](https://primeng.org/migration/v20)
  - `Defer`, `TriStateCheckbox`, and `VirtualScroller` have no replacements but are unused, so they are removed
- all other deps are updated to latest
- TypeScript kept at 5.8.2 instead of 5.9 for Angular 20 compatibility
- control flow switched to `@if` and `@for`

I did not re-alphabetize imports so it is easier for reviewers to see which components were replaced.

### DIFF
--- a/webapp/player-project/tsconfig.json
+++ b/webapp/player-project/tsconfig.json
@@ -1,9 +1,9 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
+    "target": "ES2022",
     "useDefineForClassFields": true,
     "module": "ESNext",
-    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "lib": ["ES2022", "DOM", "DOM.Iterable"],
     "skipLibCheck": true,
 
     /* Bundler mode */
--- a/webapp/recording-player-tester/tsconfig.json
+++ b/webapp/recording-player-tester/tsconfig.json
@@ -1,9 +1,9 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
+    "target": "ES2022",
     "useDefineForClassFields": true,
     "module": "ESNext",
-    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "lib": ["ES2022", "DOM", "DOM.Iterable"],
     "skipLibCheck": true,
 
     /* Bundler mode */
--- a/webapp/video-player/tsconfig.json
+++ b/webapp/video-player/tsconfig.json
@@ -1,9 +1,9 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
+    "target": "ES2022",
     "useDefineForClassFields": true,
     "module": "ESNext",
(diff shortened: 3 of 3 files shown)


CBenoit (commented):
> This updates all deps in the webapp.
> 
> * Sass imports are refactored to `@use` due to `@import` deprecation in dart-sass. Namespaces are explicitly set, .i.e., `variables.$borderRadius`
> 
> * Angular components are all standalone
> 
> * PrimeNG components are replaced with v20 versions using the [migration guide](https://primeng.org/migration/v20)
>       
> * `Defer`, `TriStateCheckbox`, and `VirtualScroller` have no replacements but are unused, so they are removed
> 
> * all other deps are updated to latest
> 
> * TypeScript kept at 5.8.2 instead of 5.9 for Angular 20 compatibility

Thank you for the methodical and well-documented migration.
Have you tested the login form functionality specifically, given the p-messages deprecation?

> Areas to check (and marked with a TODO comment):
> 
> * login.component.html: The `p-checkbox` had a bound label. The `label` attribute has been deprecated so the formatting may have changed when we add the label separately
> 
> * session-toolbar.component.html: `p-messages` is deprecated and replaced with `p-message` in the `login-form-message`

Is it something you will look into, or we should ask help to Irving and Krista?

> There are still warnings for the `ngIf` control flow syntax. Full conversion to the new `@if` syntax can be done after this PR. I did not re-alphabetize imports so it is easier for reviewers to see which components were replaced.

Sounds good.
Did you also check if there were any runtime warnings or errors in the console after this upgrade?

webapp/src/client/app/modules/base/menu/menu-group-list-item/menu-group-list-item.component.html
```diff
@@ -1,30 +1,30 @@
...
+  [class.isSlim]="isMenuSlim" [class.closed]="isClosed">
 
-  <div
-    *ngIf="hasHeader"
-    class="menu-group-list-header"
-    [class.menu-group-list-header--slim]="isMenuSlim"
-    (click)="onToggleMenuGroupList($event)"
-    (mouseover)="onMenuGroupMouseOver()"
-    (mouseleave)="onMenuGroupMouseLeave()">
+  @if (hasHeader) {
```
irvingoujAtDevolution, line 4: Wow! This @if syntax is amazing!

irvingoujAtDevolution (commented):
LGTM! Thank you!

kristahouse (commented):
Wow. That's a lot of changes! Great progress with v 20! Well done 👍

CBenoit (approved):
Approving as agreed on Slack. Merge when ready 👍

PR #1454, opened 2025-08-21
revert: update to Angular 20 and PrimeNG 20

CBenoit:
Will be re-applied later.

Refs: 966ba8cfd1234b18032672906d56b2a5c4618104


CBenoit:
As discussed on Slack

PR #1460, opened 2025-08-29
refactor: follow up to PR 1446

CBenoit:
cc @philippelatulippe

### DIFF
--- a/crates/network-monitor/src/lib.rs
+++ b/crates/network-monitor/src/lib.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use network_scanner::ping;
+use network_scanner_net::runtime::Socket2Runtime;
 use thiserror::Error;
 use time::UtcDateTime;
 use tokio_util::sync::CancellationToken;
@@ -12,7 +13,6 @@ mod log_queue;
 mod state;
 
 pub use crate::state::{ConfigCache, State};
-pub use network_scanner_net::runtime::Socket2Runtime;
 
 #[derive(Error, Debug)]
 pub enum SetConfigError {
--- a/crates/network-monitor/src/state.rs
+++ b/crates/network-monitor/src/state.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex, RwLock};
 
+use anyhow::Context as _;
 use network_scanner_net::runtime::Socket2Runtime;
 use tokio_util::sync::CancellationToken;
 
@@ -33,14 +34,18 @@ pub struct State {
 }
 
 impl State {
-    pub fn new(config_cache: Arc<dyn ConfigCache>, scanner_runtime: Arc<Socket2Runtime>) -> State {
-        State {
+    pub fn new(config_cache: Arc<dyn ConfigCache>) -> anyhow::Result<State> {
+        let scanner_runtime = Socket2Runtime::new(None).context("create socket2 runtime")?;
+
+        let state = State {
             config_cache,
             log: LogQueue::new(),
             config: RwLock::new(MonitorsConfig::empty()),
             cancellation_tokens: Mutex::new(HashMap::new()),
             scanner_runtime,
             set_config_permit: tokio::sync::Semaphore::new(1),
-        }
+        };
+
+        Ok(state)
     }
 }
--- a/devolutions-gateway/src/lib.rs
+++ b/devolutions-gateway/src/lib.rs
@@ -79,8 +79,7 @@ impl DgwState {
         let (job_queue_handle, job_queue_rx) = job_queue::JobQueueHandle::new();
         let credential_store = credential::CredentialStoreHandle::new();
 
-        let socket2runtime = network_monitor::Socket2Runtime::new(None)?;
-        let monitoring_state = Arc::new(network_monitor::State::new(Arc::new(MockMonitorsCache), socket2runtime));
+        let monitoring_state = Arc::new(network_monitor::State::new(Arc::new(MockMonitorsCache))?);
 
         let state = Self {
             conf_handle,
--- a/crates/network-scanner/src/ping.rs
+++ b/crates/network-scanner/src/ping.rs
@@ -1,4 +1,3 @@
-use std::default;
 use std::mem::MaybeUninit;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr, ToSocketAddrs};
 use std::sync::Arc;
--- a/devolutions-gateway/src/service.rs
+++ b/devolutions-gateway/src/service.rs
@@ -246,11 +246,7 @@ async fn spawn_tasks(conf_handle: ConfHandle) -> anyhow::Result<Tasks> {
     let filesystem_monitor_config_cache = devolutions_gateway::api::monitoring::FilesystemConfigCache::new(
         config::get_data_dir().join("monitors_cache.json"),
     );
-    let socket2runtime = network_monitor::Socket2Runtime::new(None).context("init Socket2Runtime")?;
-    let monitoring_state = Arc::new(network_monitor::State::new(
-        Arc::new(filesystem_monitor_config_cache),
-        socket2runtime,
-    ));
+    let monitoring_state = Arc::new(network_monitor::State::new(Arc::new(filesystem_monitor_config_cache))?);
 
     let state = DgwState {
         conf_handle: conf_handle.clone(),


crates/network-monitor/src/lib.rs
```diff
@@ -1,155 +1,133 @@
...
         Err(error) => MonitorResult {
             monitor_id: definition.id.clone(),
-            request_start_time: start_time,
+            request_start_time,
             response_success: false,
-            response_messages: Some(format!("{error:#}")),
-            response_time: f64::INFINITY,
+            response_message: Some(format!("{error:#}")),
+            response_time: None,
         },
```
CBenoit, line 134: For the internal model (not the DTO), using an Option, so that when there is no proper value for response time, we set it to None enabling more robust type checking

crates/network-monitor/src/lib.rs
```diff
@@ -177,44 +155,29 @@ impl MonitorsConfig {
...
 pub struct MonitorDefinition {
     pub id: String,
     pub probe: ProbeType,
     pub address: String,
-    pub interval: u64,
-    pub timeout: u64,
-    pub port: Option<i16>,
+    pub interval: Duration,
+    pub timeout: Duration,
+    pub port: Option<u16>,
```
CBenoit, line 176: Using stronger types

crates/network-monitor/src/lib.rs
```diff
@@ -177,44 +155,52 @@ impl MonitorsConfig {
...
     pub request_start_time: UtcDateTime,
     pub response_success: bool,
-    pub response_messages: Option<String>,
-    pub response_time: f64,
+    pub response_message: Option<String>,
+    pub response_time: Option<Duration>,
+}
+
+#[derive(Eq, PartialEq, Hash, Clone, Debug)]
+pub struct MonitorId(String);
```
CBenoit, line 189: NewType wrapper around String to get stronger type checking for monitor IDs

crates/network-monitor/src/state.rs
```diff
@@ -1,39 +1,46 @@
...
 
-use crate::log_queue::*;
-use crate::*;
+pub trait ConfigCache: Send + Sync {
+    fn store(&self, new_config: &MonitorsConfig) -> anyhow::Result<()>;
+}
 
 pub struct State {
-    pub(crate) cache_path: Utf8PathBuf,
+    pub(crate) config_cache: Arc<dyn ConfigCache>,
```
CBenoit, line 16: Offloaded the concern of cache storage out of network-monitor crate

crates/network-monitor/src/lib.rs
```diff
@@ -1,155 +1,133 @@
...
-        .write(true)
-        .truncate(true)
-        .open(&state.cache_path)?;
+pub async fn set_config(new_config: MonitorsConfig, state: Arc<State>) -> Result<(), SetConfigError> {
+    // Ensure set_config is never run concurrently using a 1-permit semaphore.
+    let _guard = state
+        .set_config_permit
+        .acquire()
+        .await
+        .expect("as per invariant, semaphore is never closed");
```
CBenoit, line 29: Using a semaphore to ensure set_config is never executed concurrently

crates/network-monitor/Cargo.toml
```diff
@@ -6,21 +6,14 @@ authors = ["Devolutions Inc. <infos@devolutions.net>"]
 publish = false
 
 [dependencies]
-network-scanner = { path = "../network-scanner" }
+anyhow = "1"
 network-scanner-net = { path = "../network-scanner-net" }
-serde = { version = "1", features = ["derive"] }
-camino = { version = "1", features = ["serde1"] }
-serde_json = "1"
```
CBenoit, line 13: No serialization concern in network-monitor

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -140,7 +143,7 @@ pub(crate) struct MonitorDefinition {
     address: String,
     interval: u64,
     timeout: u64,
-    port: Option<i16>,
+    port: Option<u16>,
```
CBenoit, line 146: The correct type for ports is u16

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -209,18 +212,76 @@ pub(crate) struct MonitorResult {
     request_start_time: OffsetDateTime,
     response_success: bool,
     #[serde(skip_serializing_if = "Option::is_none")]
-    response_messages: Option<String>,
+    response_message: Option<String>,
+    /// Duration it took for the response to be received, in seconds.
+    ///
+    /// On error, this value is set to f64::INFINITY.
     response_time: f64,
```
CBenoit, line 219: @philippelatulippe What about using `Option` and returning `None` when there is no response time? The problem with infinity is that at the type level it looks like a regular value, and this tends to lead to subtle logic bugs.

devolutions-gateway/src/lib.rs
```diff
@@ -111,7 +103,15 @@ impl DgwState {
...
-        Ok((state, handles))
+        return Ok((state, handles));
+
+        struct MockMonitorsCache;
+
+        impl network_monitor::ConfigCache for MockMonitorsCache {
+            fn store(&self, _: &network_monitor::MonitorsConfig) -> anyhow::Result<()> {
+                Ok(())
+            }
+        }
```
CBenoit, line 113: Using a dummy cache for the test state

devolutions-gateway/Cargo.toml
```diff
@@ -64,7 +64,6 @@ backoff = "0.4"
 sysinfo = { version = "0.35", default-features = false, features = ["disk"] }
 dunce = "1.0"
 bitflags = "2.9"
-tempdir = "0.3"
```
CBenoit, line 67: Removed this should-have-been-dev-dependency from our regular dependency graph

crates/network-monitor/src/state.rs
```diff
@@ -1,39 +1,46 @@
...
     pub(crate) log: LogQueue<MonitorResult>,
     pub(crate) config: RwLock<MonitorsConfig>,
-    pub(crate) cancellation_tokens: Mutex<HashMap<String, CancellationToken>>,
-    pub(crate) scanner_runtime: LazyLock<anyhow::Result<Arc<Socket2Runtime>>>,
+    pub(crate) cancellation_tokens: Mutex<HashMap<MonitorId, CancellationToken>>,
+    pub(crate) scanner_runtime: Arc<Socket2Runtime>,
```
CBenoit, line 20: The LazyLock idea was nice, but we end up with an anyhow::Result<Socket2Runtime> instead of a Socket2Runtime, forcing us to check for errors inside the monitoring logic. It’s better to build the runtime upfront and verify the error once, so we can have a cleaner logic in the rest of the code.

crates/network-monitor/src/lib.rs
```diff
@@ -1,155 +1,133 @@
...
-                        _ = cancellation_monitor.cancelled() => { return }
-                        _ = tokio::time::sleep(Duration::from_secs_f64(next_run_in)) => { }
-                    };
-                }
-            };
-
-            (
-                (definition_id, cancellation_token),
-                Box::pin(monitor) as Pin<Box<dyn Future<Output = ()> + Send>>,
-            )
```
CBenoit, line 98: Previous logic

crates/network-monitor/src/lib.rs
```diff
@@ -1,155 +1,133 @@
...
+                    ProbeType::TcpOpen => do_tcpopen_monitor(&definition).await,
+                };
+
+                state.log.write(monitor_result);
+            }
+        };
+
+        tokio::spawn(monitor_task);
+
+        (definition_id, cancellation_token)
```
CBenoit, line 100: New logic

devolutions-gateway/src/api/monitoring.rs
```diff
@@ -209,18 +212,76 @@ pub(crate) struct MonitorResult {
     request_start_time: OffsetDateTime,
     response_success: bool,
     #[serde(skip_serializing_if = "Option::is_none")]
-    response_messages: Option<String>,
+    response_message: Option<String>,
```
CBenoit, line 215: @philippelatulippe I renamed to remove the `s`, since there can be only one message at most. Does that sound good?

CBenoit, line 215: @philippelatulippe Another question: is it really necessary to have a `response_` prefix? `MonitorResult::success` and `MonitorResult::message` seems clear enough to me.

crates/network-monitor/src/lib.rs
```diff
@@ -1,155 +1,133 @@
...
-                        _ = tokio::time::sleep(Duration::from_secs_f64(next_run_in)) => { }
-                    };
-                }
-            };
-
-            (
-                (definition_id, cancellation_token),
-                Box::pin(monitor) as Pin<Box<dyn Future<Output = ()> + Send>>,
-            )
-        })
```
CBenoit, line 99: Previous logic

crates/network-monitor/src/lib.rs
```diff
@@ -1,155 +1,133 @@
...
+                };
+
+                state.log.write(monitor_result);
+            }
+        };
+
+        tokio::spawn(monitor_task);
+
+        (definition_id, cancellation_token)
+    }
```
CBenoit, line 101: New logic

crates/network-monitor/src/lib.rs
```diff
@@ -1,155 +1,136 @@
...
+            let monitor_id = definition.id.clone();
+            let cancellation_token = spawn_monitor(Arc::clone(&state), definition);
+            cancellation_tokens.insert(monitor_id, cancellation_token);
+        }
+
+        for definition in deleted {
+            match cancellation_tokens.get(&definition.id) {
+                Some(token) => token.cancel(),
+                None => warn!("cancellation token for monitor {} not found", definition.id),
+            }
```
CBenoit, line 64: Panic resistant code

devolutions-gateway/src/api/monitoring.rs
philippelatulippe: Yup

philippelatulippe: I was planning to replace this field with something more structured than a string, this field was just provisional until I implement more monitor types and have a better idea of what info should be returned. "message" probably isn't the right term in the first place, since it's not a message coming from the network, it's rather information coming from the monitor itself

devolutions-gateway/src/api/monitoring.rs
philippelatulippe: I was on the fence about it, but you're right that Option would be better so that it doesn't get treated as a regular value. My reasoning was that the duration of a response that never arrives can indeed be considered infinite, but it's true that practical considerations would favour Option.

crates/network-monitor/src/state.rs
philippelatulippe: I went for this approach because it seemed strange that the gateway crate would concern itself with creating the Socket2Runtime, something which is totally internal to the monitoring crate. I don't think it should be exposed at all since I can't see the need to inject it. Maybe `State::new()` could return a Result instead?

devolutions-gateway/src/api/monitoring.rs
philippelatulippe: Oops, glad you spotted that!

crates/network-monitor/src/state.rs
CBenoit: Works too! I was not sure which one you would prefer

devolutions-gateway/src/api/monitoring.rs
CBenoit: Yeah, I thought it was making sense too, but re-thinking about it, it’s probably safer to consider "no-value" when we don’t want to process it at all

devolutions-gateway/src/api/monitoring.rs
CBenoit: Sounds good!

crates/network-monitor/src/state.rs
CBenoit: I updated the PR accordingly

philippelatulippe (approved):

PR #1459, opened 2025-08-28
refactor(webapp): replace the current clipboard with auto and manual modes

RRRadicalEdward:
This PR refactors the clipboard part of the Devolutions Gateway Standalone to follow the new API(see https://github.com/Devolutions/IronRDP/pull/935/) of the _iron-remote-desktop_. The changes:
 * Introduce _autoclipboard_ mode, which the user can enable/disable using the connection form. However, it's available only for browsers based on _Blink_  engine. For others, _autoclipboard_ mode is always false.
 * Add two new buttons to the toolbar that are available when _autoclipboard_ is false: _Save Clipboard_ and _Send Clipboard_.
 * Add notification for manual clipboard actions. For example, when the user clicks the _Save Clipboard_ button, an error or success notification will appear, informing the user about the operation status.
 
Demos:

- Chrome auto clipboard mode

https://github.com/user-attachments/assets/001d150a-982a-430d-84ba-fd29b5ea1258

- Chrome manual clipboard mode

https://github.com/user-attachments/assets/957d9832-80af-4c95-9559-a11ab8eeba12

- Firefox (Safari behaves the same way) clipboard

https://github.com/user-attachments/assets/7842e5ec-c83e-40a6-9d81-2b09f99fa4e8

PR #1464, opened 2025-09-01
fix(webapp): improve clipboard interaction for outdated versions of the Firefox browser

RRRadicalEdward:
Follow-up to https://github.com/Devolutions/IronRDP/pull/951.

Before merging this PR, we have to release https://github.com/Devolutions/IronRDP/pull/951(https://github.com/Devolutions/IronRDP/pull/957). Also, we should merge it not before https://github.com/Devolutions/devolutions-gateway/pull/1459.

### DIFF
--- a/webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
+++ b/webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
@@ -161,22 +161,25 @@ export class WebClientVncComponent extends WebClientBaseComponent implements OnI
       return;
     }
 
-    this.clipboardActionButtons.push(
-      {
-        label: 'Save Clipboard',
-        tooltip: 'Copy received clipboard content to your local clipboard.',
-        icon: 'dvl-icon dvl-icon-save',
-        action: () => this.saveRemoteClipboard(),
-        enabled: () => this.saveRemoteClipboardButtonEnabled,
-      },
-      {
+    // We don't check for clipboard write support, as all recent browser versions support it.
+    this.clipboardActionButtons.push({
+      label: 'Save Clipboard',
+      tooltip: 'Copy received clipboard content to your local clipboard.',
+      icon: 'dvl-icon dvl-icon-save',
+      action: () => this.saveRemoteClipboard(),
+      enabled: () => this.saveRemoteClipboardButtonEnabled,
+    });
+
+    // Check if the browser supports reading local clipboard.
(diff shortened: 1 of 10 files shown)


CBenoit (approved):
LGTM! Need rebase, but enabling auto-merge 👍

webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
```diff
@@ -464,7 +467,7 @@ export class WebClientVncComponent extends WebClientBaseComponent implements OnI
       sessionId,
       enabledEncodings: enabledEncodings.join(','),
       enableCursor,
-      enableExtendedClipboard,
+      enableExtendedClipboard: enableExtendedClipboard ?? false,
```
CBenoit, line 470: question: Why is it necessary to use add `?? false`?

webapp/src/client/app/shared/interfaces/forms.interfaces.ts
```diff
@@ -40,7 +40,7 @@ export interface VncFormDataInput {
   password: string;
   protocol: number;
   enableCursor: boolean;
-  enableExtendedClipboard: boolean;
+  enableExtendedClipboard?: boolean;
```
CBenoit, line 43: question: I understand this is unrelated to the main changes? I was left wondering why it was necessary to change this?

RRRadicalEdward: If the clipboard is not supported in outdated Firefox versions, the `extendedClipboard` checkbox is hidden and the form will have an `undefined` value for the `enableExtendedClipbord`. To show it explicitly, I added a `?` to the value.

webapp/src/client/app/modules/web-client/vnc/web-client-vnc.component.ts
RRRadicalEdward: Because now the `enableExtendedClipboard` can be undefined (see the comment below).

webapp/src/client/app/shared/interfaces/forms.interfaces.ts
CBenoit: Do you think you could add an inline comment explaining this when you get a chance? Thanks!

RRRadicalEdward: Done in https://github.com/Devolutions/devolutions-gateway/pull/1473

PR #1475, opened 2025-09-05
feat(webapp): add color format selector for VNC session

RRRadicalEdward:
This way, a user can tune the session for their environment (e.g, lower image quality if the network is slow).

Closes https://github.com/Devolutions/IronVNC/issues/908.

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -32,7 +32,7 @@
     "@devolutions/icons": "4.0.8",
     "@devolutions/iron-remote-desktop": "0.9.0",
     "@devolutions/iron-remote-desktop-rdp": "^0.6.0",
-    "@devolutions/iron-remote-desktop-vnc": "^0.5.0",
+    "@devolutions/iron-remote-desktop-vnc": "../../ironvnc/web-client/iron-remote-desktop-vnc/dist",
     "@devolutions/terminal-shared": "^1.3.1",
     "@devolutions/web-ssh-gui": "^0.4.0",
     "@devolutions/web-telnet-gui": "^0.3.0",
--- a/webapp/src/client/app/shared/enums/color-format.enum.ts
+++ b/webapp/src/client/app/shared/enums/color-format.enum.ts
@@ -1,0 +1,36 @@
+import { SelectItemWithTooltip } from '@shared/interfaces/select-item-tooltip.interface';
+
+enum ColorFormat {
+  Default = '',
+  Full = 'rgb888le',
+  High = 'rgb655le',
+  Medium = 'rgb332le',
+  Low = 'rgb222le',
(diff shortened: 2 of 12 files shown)


webapp/package.json
```diff
@@ -32,7 +32,7 @@
     "@devolutions/icons": "4.0.8",
     "@devolutions/iron-remote-desktop": "0.9.0",
     "@devolutions/iron-remote-desktop-rdp": "^0.6.0",
-    "@devolutions/iron-remote-desktop-vnc": "^0.5.0",
+    "@devolutions/iron-remote-desktop-vnc": "../../ironvnc/web-client/iron-remote-desktop-vnc/dist",
```
RRRadicalEdward: https://github.com/Devolutions/IronVNC/pull/955 has to be merged and released first.

webapp/src/client/app/shared/enums/color-format.enum.ts
```diff
@@ -0,0 +1,36 @@
+import { SelectItemWithTooltip } from '@shared/interfaces/select-item-tooltip.interface';
+
+enum ColorFormat {
+  Default = '',
+  Full = 'rgb888le',
```
RRRadicalEdward, line 5: Here are the available options for color formats

webapp/src/client/app/shared/enums/color-format.enum.ts
```diff
@@ -0,0 +1,36 @@
...
+  'Very Low' = 'rgb111le',
+}
+
+enum Tooltips {
+  Default = 'Color format advertised by a server',
+  Full = 'True 24-bit color',
+  High = 'High 16-bit color',
+  Medium = '256 colors',
+  Low = '64 colors',
+  'Very Low' = '8 colors',
```
CBenoit: question: Is it really advertised _by_ the server?

RRRadicalEdward: What do you mean by "advertised by the server"?

RRRadicalEdward: It's the client's ability to be able to force a specific color format. This one might not be the most commonly used, but it's present in UltraVNC and other clients, so it makes sense to have it as well.

CBenoit: Oh sorry, I just got what you meant now. Maybe word it like `Use the color format advertised by the server`? If I understood you correctly.

RRRadicalEdward: Done

webapp/package.json
RRRadicalEdward: Done

CBenoit (approved):

PR #1491, opened 2025-09-11
feat(dgw): emit syslogs and Windows events for important events

CBenoit:
Easier auditability of Devolutions Gateway service by emitting system wide logs.

Issue: DGW-63
Security: yes

### DIFF
--- a/package/WindowsManaged/Program.cs
+++ b/package/WindowsManaged/Program.cs
@@ -358,7 +358,7 @@ internal class Program
             {
                 AttributesDefinition = "Type=string",
                 Win64 = project.Platform == Platform.x64,
-                RegistryKeyAction = RegistryKeyAction.create,
+                RegistryKeyAction = RegistryKeyAction.createAndRemoveOnUninstall,
             }
         };
         project.Properties = GatewayProperties.Properties.Select(x => x.ToWixSharpProperty()).ToArray();


package/WindowsManaged/Program.cs
```diff
@@ -353,6 +353,12 @@ static void Main()
                 AttributesDefinition = "Type=string; Component:Permanent=yes",
                 Win64 = project.Platform == Platform.x64,
                 RegistryKeyAction = RegistryKeyAction.create,
+            },
+            new (RegistryHive.LocalMachine, $"SYSTEM\\CurrentControlSet\\Services\\EventLog\\Application\\{Includes.PRODUCT_NAME}", "EventMessageFile", $"[{GatewayProperties.InstallDir}]{Includes.EXECUTABLE_NAME}")
+            {
+                AttributesDefinition = "Type=string",
+                Win64 = project.Platform == Platform.x64,
+                RegistryKeyAction = RegistryKeyAction.create,
             }
```
CBenoit: @thenextman I added a event source "Devolutions Gateway". To go along, here is a new registry key registering the event source. Does that look good to you?

thenextman: It looks fine to me; but we may want to check: does the key get removed at uninstall time? `RegistryKeyAction.create` implies that it **won't**, but on the `InstallDir` key I also set a custom attribute to mark the component permanent.

If the key is removed, does it break something in the event log? I never dealt with a custom .mc file before (.NET provides one for .NET applications).

CBenoit: Good catch. If this key is removed we lose the formatting and localization in the Event Viewer, but we would lose that anyway when the resources are removed along the executable so it’s best to remove the registry key I think

package/WindowsManaged/Program.cs
```diff
@@ -353,6 +353,12 @@ static void Main()
                 AttributesDefinition = "Type=string; Component:Permanent=yes",
                 Win64 = project.Platform == Platform.x64,
                 RegistryKeyAction = RegistryKeyAction.create,
+            },
+            new (RegistryHive.LocalMachine, $"SYSTEM\\CurrentControlSet\\Services\\EventLog\\Application\\{Includes.PRODUCT_NAME}", "EventMessageFile", $"[{GatewayProperties.InstallDir}]{Includes.EXECUTABLE_NAME}")
+            {
+                AttributesDefinition = "Type=string",
+                Win64 = project.Platform == Platform.x64,
+                RegistryKeyAction = RegistryKeyAction.createAndRemoveOnUninstall,
```
CBenoit, line 361: @thenextman I changed to createAndRemoveOnUninstall

PR #1499, opened 2025-09-12
refactor(webapp): refactor VNC encoding selection settings

RRRadicalEdward:
* refactor the encoding selector to select one encoding at a time (we have the same in _RDM_)
* add `Default` encoding to the encoding selector list, which defaults to sending all encodings to the server, and let it select one (we have the same in _RDM_)
* added logic to hide the `pixel format` option if _Tight JPEG_ or _Tight PNG_ is selected. These don't work well together, as the https://github.com/Devolutions/IronVNC/issues/924 bug happens.
* removed _Tight PNG_ encoding from the encoding selector and added a checkbox with the same appearing only when `Tight` encoding is selected. _Tight PNG_ is more like an extension for _Tight_ (like `JPEG`), rather than a separate thing. So, it makes sense to keep it as a setting for _Tight_ encoding.
* added hiding `JPEG` option for other encodings than _Tight_, as it's only matters when _Tight_ is enabled.

Showcase: 

https://github.com/user-attachments/assets/7019761b-d811-467e-8296-eec8875766b7


CBenoit:
In RDM, we have a list of "preferred encodings"

image

Also, here is what @thenextman said on this topic:

> I would think hard about giving the user a list of “preferred encodings” again. Because it’s just technobabble to most people and rarely used. Something I’d discussed with David in the past was presenting it more like “Automatic”, “High quality / slower”, “medium”, “low quality / faster”. That kind of thing. With preconfigured preferences underneath.

Given this feedback, I think its not really worth supporting the "preferred encodings" of RDM, and instead provide an extension for the "Encoding Profile". Instead of disabling / enabling, we should adapt the ordering.

What do you think? Maybe we can talk more deeply on Slack.

RRRadicalEdward:
In `IronEncodingSet`, we also have pseudo-encoding, and with this change, we need to pay attention to keep sending them for all possible encoding sets. If we refactored `IronEncodingSet` to have a few smaller sets, we would have to copy-paste the pseudo-encodings to each set or come up with a way to separate them from the no-pseudo-encodings. 

> Alternatively, we extend the ironvnc-session crate and offer an API to assign weights to Encodings as a way to change the ordering when sending the SetEncodings packet.

It might be a better solution here, as we wouldn't have to touch the pseudo-encodings.

RRRadicalEdward:
> > In `IronEncodingSet`, we also have pseudo-encoding, and with this change, we need to pay attention to keep sending them for all possible encoding sets. If we refactored `IronEncodingSet` to have a few smaller sets, we would have to copy-paste the pseudo-encodings to each set or come up with a way to separate them from the no-pseudo-encodings.
> 
> I think we can factor that using a declarative macro.
> 
> > > Alternatively, we extend the ironvnc-session crate and offer an API to assign weights to Encodings as a way to change the ordering when sending the SetEncodings packet.
> > 
> > 
> > It might be a better solution here, as we wouldn't have to touch the pseudo-encodings.
> 
> But yes, I have a small preference for this solution as well. I think it’s better than completely relying on the ordering in the enum.

Another solution can be to have the pseudo-encoding in a separate set(it might be hardcoded, and user code won't be able to change it)

CBenoit:
> Another solution can be to have the pseudo-encoding in a separate set (it might be hardcoded, and user code won't be able to change it)

Makes sense too! :thinking:

RRRadicalEdward:
> I will come back to you with the brainstormed list of encoding profiles and what encoding to include where, as I see it.

Hi @CBenoit @thenextman. 
I prepared a list of VNC encoding profiles (similar to ARD quality modes in FreeVNC/IronVNC) that make sense to show to a user instead of having a "preferred encodings" list. 

At first, I separated the encodings in terms of _CPU_ and _bandwidth_ usage, and got this table(_CPU_ is vertical, and _bandwidth_ is horizontal). Note that we don't have _H.264_, _CoRRE_, _RRE_, and _ZRLE_ implemented in IronVNC yet.  

| _CPU_/_bandwidth_| High | Middle | Low |
| :------- | :------: | :-------: | -------: |
| **High** |   | <ul><li>ZlibHextile</li><li>ZLib</li></ul> | H.264  |
| **Middle** | | TightPNG   |  <ul><li>TightJPEG</li><li>Tight</li><li>ZRLE</li></ul>   |
| **Low** | <ul><li>Hextile</li><li>Raw</li></ul> | <ul><li>CoRRE</li><li>RRE</li></ul> | CopyRect |

Using this table, I got the following encoding profiles (the higher the better). 
**High _CPU_**, **low _bandwidth_**
- H.264 (lossy)
- TightJPEG (lossy)
- Tight
- ZRLE
- CopyRect

**High _CPU_**, **middle _bandwidth_**
- ZlibHextile
- Zlib
- H.264
- TightPNG
- TightJPEG (lossy)
- TIght
- ZRLE
- CoRRE
- RRE
- CopyRect

**Low _CPU_**, **high _bandwidth_**
- Hextile
- CoRRE
- RRE
- Raw
- CopyRect

**Low _CPU_**, **middle _bandwidth_**
- CoRRE
- RRE
- CopyRect

**Middle _CPU_**, **middle _bandwidth_**
- TightJPEG (lossy)
- TightPNG
- Tight
- ZRLE
- CoRRE
- RRE
- CopyRect

I think allowing the user to select the desired _CPU_ consumption, and specify the network _bandwidth_, is better than selecting preferred encoding, and should give the user a clear explanation of what to expect. 
Let me know what you think about the encoding sets 🙂.

PR #1514, opened 2025-09-19
fix(jetsocat): make the MCP proxy non intrusive

CBenoit:
The MCP proxy was previously handling the initialize request itself and returning a response instead of forwarding it to the target MCP server. It also intercepted notifications/initialized and logging/ setLevel without passing them through.

Additionally, any unknown methods were being answered with an error response, even though the actual server might support them.

This change forwards all requests and notifications directly to the server to ensure correct behavior and compatibility.

Issue: DGW-308


crates/mcp-proxy/src/lib.rs
```diff
@@ -123,516 +103,321 @@ impl McpProxy {
...
-            if let Some(result) = result_obj.get_mut("result") {
-                if let Some(result_inner) = result.get_mut::<HashMap<String, tinyjson::JsonValue>>() {
-                    if let Some(content) = result_inner.get_mut("content") {
-                        if let Some(arr) = content.get_mut::<Vec<tinyjson::JsonValue>>() {
-                            for item in arr.iter_mut() {
-                                if let Some(item_obj) = item.get_mut::<HashMap<String, tinyjson::JsonValue>>() {
-                                    if let Some(text) = item_obj.get_mut("text") {
-                                        if let Some(s) = text.get::<String>() {
-                                            let decoded = s
-                                                .replace("\\u0027", "'")
```
pacmancoder, line 611: That may be the biggest `if` chain I've ever seen 😄

testsuite/Cargo.toml
```diff
@@ -16,16 +16,24 @@ path = "tests/main.rs"
 harness = true
 
 [dependencies]
+anyhow = "1.0"
 assert_cmd = "2.0"
+dynosaur = "0.3"
```
pacmancoder, line 21: Wow, didn't know about that cool crate!

pacmancoder (approved):
LGTM!

crates/mcp-proxy/src/lib.rs
CBenoit: Yeah right! :joy: 
This was a remaining of the AI-generated stuff from Marc-André’s repo: https://github.com/awakecoding/mcp-proxy-tool/blob/5c8e1664c06c2e0fea2c5f5e861788bd188adfe6/src/main.rs#L285-L306
But got worse when I removed serde-json

testsuite/Cargo.toml
CBenoit: It’s actually recommended over `async-trait` now! The implementer does not need to know about the crate too :slightly_smiling_face:

PR #1519, opened 2025-09-29
build(package): remove unnecessary lines

allan2:
- removes unnecessary clear lines (following typical convention in /usr/lib/systemd/system/foo.service)
- removes extra leading whitespace


CBenoit (approved):
LGTM!

PR #1535, opened 2025-10-10
refactor(jetsocat): implement better coloring control for logging

RRRadicalEdward:
This PR improves coloring control as requested in https://github.com/Devolutions/IronVNC/pull/988#pullrequestreview-3303353388.

### DIFF
--- a/testsuite/tests/cli/jetsocat.rs
+++ b/testsuite/tests/cli/jetsocat.rs
@@ -37,15 +37,15 @@ fn all_subcommands() {
 }
 
 #[rstest]
-#[case::default(&[], &[], true)]
+#[case::default(&[], &[], false)] // is_terminal = false
 #[case::cli_always(&["--color=always"], &[], true)]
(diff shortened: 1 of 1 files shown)


testsuite/tests/cli/jetsocat.rs
```diff
@@ -37,15 +37,15 @@ fn all_subcommands() {
 }
 
 #[rstest]
-#[case::default(&[], &[], true)]
+#[case::default(&[], &[], false)] // is_terminal = false
```
RRRadicalEdward, line 40: In tests, `assert_cmd` is used to run _jetsocat_ in a child process, where _stderr_ is piped. This means that we always have `is_terminal = false`. That's why, for example, `log_term_coloring` with the default case will return _false_.

CBenoit: I’ll investigate if we can still check the TTY case properly.

PR #1539, opened 2025-10-20
Windows installer UX improvements

thenextman:
Several improvements for the Devolutions Gateway Windows installer.

- If the user is installing Gateway to interoperate with another server (versus the standalone web application), they'll see a screen allowing them to enter their Devolutions Server URL. The URL must be validated before being allowed to proceed.
    - If they elect to provide a URL; the screen where they must provide a copy of the server's public key is skipped. At install time, the public key is downloaded directly from the specified Devolutions Server.
    - The user may choose to proceed manually, and specify the public key file themself (as before)
    - Since this is a public property, it can be configured on the command line with `P.DEVOLUTIONSSERVERURL="..."`

Screenshot 2025-10-20 at 13 27 35

- During listener configuration, we perform a basic port check on the values the user provided. The check is debounced, but it had a bug which could _sometimes_ occur if the user switched to another page while the debouncer was running. I rewrote the debouncer to be more robust.
    - Further, the result of the port check was not especially legible (the port number was coloured black, red or green appropriately). Instead, I switched it to use a WinForms error provider which is much more clear.

Screenshot 2025-10-20 at 13 27 59

- Disable all file picker dialogs when running on Windows Server Core. These would cause a crash. We _could_ fallback to a different file picker, but since this is a niche case; let's just avoid the crash. The user will have to type or paste file paths themselves.
- Fixed some broken UI layouts on extremely high scaling factors (250% or greater)
- Fixed broken hyperlinks to Devolutions Server and Hub help topics
- Cleanup .designer files for consistency; notably some forms has a bottom black border above the action buttons and some didn't
- Fix a non-handled NRE that was annoying during debugging
- Allow CLI generation on release builds. By passing the property `P.ENABLECLIGENERATION="True"` to msiexec and then stepping through the installer GUI, the last page will have a link button to show the CLI command that can be used to invoke msiexec with the given configuration. This is useful in dev, it may be useful in the real world too.

### DIFF
--- a/package/WindowsManaged/Actions/CustomActions.cs
+++ b/package/WindowsManaged/Actions/CustomActions.cs
@@ -363,7 +363,7 @@ namespace DevolutionsGateway.Actions
 
                     if (!string.IsNullOrEmpty(publicKeyFile))
                     {
-                        command += $" -PublicKeyFile '{session.Get(GatewayProperties.publicKeyFile)}'";
+                        command += $" -PublicKeyFile '{publicKeyFile}'";
                     }
 
                     if (!string.IsNullOrEmpty(session.Get(GatewayProperties.privateKeyFile)))


CBenoit (approved):
LGTM! Nice improvements! Since you separated everything into clean commits, feel free to "rebase merge" instead of squashing 🙂

PR #1751, opened 2026-04-13
fix(dgw): resolve WER APIs dynamically to support Windows Server 2016

thenextman:
`WerRegisterExcludedMemoryBlock` is absent from `wer.dll` on Windows Server 2016 RTM (NT 10.0.14393). The static import introduced in v2026.1.1 caused the Windows loader to refuse starting DevolutionsGateway.exe on those hosts with a "procedure entry point not found" error.

Replace the static import with runtime resolution via `GetModuleHandleW` and `GetProcAddress`, cached in a `OnceLock`. The binary now starts on any Windows 10/Server 2016 build and silently skips WER dump exclusion when the API is unavailable; all other protections (guard pages, `VirtualLock`, `PAGE_READONLY`) still apply.

⚠️ Authored with Claude, but seems sane to me

Issue: DGW-368

### DIFF
--- a/crates/secure-memory/Cargo.toml
+++ b/crates/secure-memory/Cargo.toml
@@ -21,7 +21,7 @@ libc = "0.2"
 version = "0.61"
 features = [
     "Win32_Foundation",
-    "Win32_System_ErrorReporting",
+    "Win32_System_LibraryLoader",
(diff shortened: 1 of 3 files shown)


CBenoit (approved):
LGTM!

PR #1768, opened 2026-04-28
feat(dgw): Kerberos credential injection via explicit jet_cred_id

irvingoujAtDevolution:
## Summary

Implements [DGW-378](https://devolutions.atlassian.net/browse/DGW-378): proxy-based Kerberos credential injection for Web Access. A new `cred_injection_id` (UUID) is minted at `/jet/preflight provision-credentials` and propagated through DVLS-issued association and KDC tokens as the JWT claim `jet_cred_id` — the gateway uses it as the primary credential-store key for `/rdp` and `/jet/KdcProxy`. Replaces the heuristic AS-REQ router from the experimental `dgw-378-session-redesign` branch.

Coupled with [DVLS-13821](https://devolutions.atlassian.net/browse/DVLS-13821); old DVLS still works for NTLM injection via the JTI fallback (no regression).

## Highlights

- Per-session `SessionKerberos` (krbtgt + service long-term keys, `secrecy`-wrapped) with a `CRED-{uuid}.INVALID` synthetic realm for bare-UUID proxy usernames.
- Gateway-as-CredSSP-server self-AS-REQ uses an in-process loopback at `http://cred.invalid/{cred_injection_id}` (RFC 6761 reserved TLD) — no HTTP, no token, no middleware.
- SPN derived from `dst_hst` (`TERMSRV/<target>`) so client AP-REQ tickets validate against the impersonated server identity. Required at insert time — missing/malformed `dst_hst` is rejected at preflight, not mid-CredSSP.

## Verification

- 5 unit tests in `credential/tests.rs` (insert / lookup / eviction / synthetic realm).
- E2E against an AD-joined RDP server via DVLS web → Gateway → target: full CredSSP MITM + KdcProxy completes in ~1.4s, no errors.

## Test plan

- [ ] CI green
- [ ] DVLS-side PR (DVLS-13821, RDM repo) reviewed in lockstep
- [ ] Release pipeline regenerates Gateway.Client SDK from OpenAPI (currently hand-edited to match)

[DGW-378]: https://devolutions.atlassian.net/browse/DGW-378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DVLS-13821]: https://devolutions.atlassian.net/browse/DVLS-13821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TheBestTvarynka (approved):
Wow, amazing feature 🤩. _Per-session Kerberos material_ is a great improvement.

I reviewed the PR. LGTM.

devolutions-gateway/src/credential/mod.rs
```diff
@@ -108,14 +114,120 @@ pub struct CleartextAppCredentialMapping {
...
+}
+
+/// Per-session Kerberos material for proxy-based credential injection.
+///
+/// The two AES-256 keys and the service user's PA_ENC_TIMESTAMP password are wrapped in
+/// [`SecretBox`] / [`SecretString`] so they cannot be accidentally written to logs through
+/// `tracing::debug!(?kerberos, ...)`. Access requires an explicit `expose_secret()` call,
+/// which is greppable and reviewable. The custom [`fmt::Debug`] adds a second layer that
+/// stays correct even if a future refactor changes the field types.
+pub struct SessionKerberos {
```
CBenoit: **Move all of this out of `credential/`.** `SessionKerberos`, `build_session_kdc_config` (line 179), `principal_for_realm`, `kerberos_salt`, `synthetic_realm`, `encrypt_with_kerberos`, `random_32_bytes`, plus the hardcoded `service_user_name: "jet"` and `max_time_skew: 300` — all of this is Kerberos/CredSSP session glue that has no business inside what used to be a protocol-neutral credential store.

Suggested layout: a new `rdp_proxy::kerberos_session` (or `credssp_session`) submodule owns the `SessionKerberos` type and the KDC config builder. `InjectionState` in `credential/` becomes just `{ mapping, target_hostname }`. The `Arc<SessionKerberos>` is either lazily derived on first CredSSP use, or stored in a parallel `HashMap<Uuid, Arc<SessionKerberos>>` owned by the RDP layer (keyed by the same UUID — see point 2 in the summary, that UUID should be the association-token JTI).

This also lets `CleartextAppCredentialMapping::encrypt_with_kerberos` go back to a plain `encrypt`, so non-Kerberos use of the store stops minting AES-256 keys it never uses.

devolutions-gateway/src/credential/mod.rs
```diff
@@ -166,35 +304,121 @@ impl CredentialStore {
...
+        let injection = mapping_and_kerberos
+            .map(|(mapping, kerberos)| -> Result<InjectionState, InsertError> {
+                // Credential injection only fires for RDP (see `GenericClient::serve` and
+                // `rd_clean_path::handle`), and the fake-KDC's TGS-REQ sname check requires
+                // `TERMSRV/<target_hostname>`. A bare hostname or `host:port` form is valid;
+                // matches how the association-token deserializer resolves `dst_hst`
+                // (`token.rs::parse_target_address`). Missing/malformed `dst_hst` is fatal
+                // here so the failure surfaces at preflight rather than mid-CredSSP.
+                const DEFAULT_DST_PORT: u16 = 3389;
+                let raw = crate::token::extract_dst_hst(&token)
```
CBenoit: Extracting `dst_hst` from the JWT inside the credential store is a layering violation. The preflight handler in `api/preflight.rs` already has the validated token in hand and could resolve `target_hostname: String` itself, then pass it to `insert` as a typed parameter. That removes the credential module's dependency on JWT shape beyond `jti`, and removes the second `discard_signature()` call on a token whose validation invariant is documented elsewhere.

A short doc comment on `CredentialStore::insert` stating the upstream-validation invariant for the existing JTI extraction would also be welcome, regardless.

devolutions-gateway/src/token.rs
```diff
@@ -431,6 +431,9 @@ pub struct AssociationTokenClaims {
     /// When set alongside `ConnectionMode::Fwd`, the Gateway will proxy the connection
     /// through the specified agent instead of connecting directly to the target.
     pub jet_agent_id: Option<Uuid>,
+
+    /// Reference to a credential-injection record provisioned via /jet/preflight.
+    pub jet_cred_id: Option<Uuid>,
```
CBenoit: **Please drop `jet_cred_id` from `AssociationTokenClaims`.** The association token already has a `jti`, and that JTI uniquely identifies the credential-store entry — every existing call site in this PR (`generic_client.rs`, `rd_clean_path.rs`) already falls back to `credential_store.get_by_token(token)` for old DVLS, which proves a JTI-keyed lookup suffices.

Keep `jet_cred_id` only on `KdcTokenClaims` (line 624), defined as "the JTI of the associated session token." That preserves your invariant that JTIs are not reused across tokens — the KDC token has its own `jti`, and `jet_cred_id` is a *reference* to a different token's ID. The KDC handler then does `credential_store.get(claims.jet_cred_id)` against the same JTI-keyed store.

This change cascades into removing `extract_jet_cred_id`/`extract_optional_uuid` callers on the association path, the preflight request field, the `provisioned-credentials` output kind, the `Conflict` error, the reverse `association_token_index`, and the `get_by_token` helper.

devolutions-gateway/src/api/preflight.rs
```diff
@@ -46,6 +46,8 @@ struct ProvisionCredentialsParams {
     token: String,
     #[serde(flatten)]
     mapping: crate::credential::CleartextAppCredentialMapping,
+    #[serde(default)]
+    cred_injection_id: Option<Uuid>,
```
CBenoit: Revert this addition (and the `PreflightOutputKind::ProvisionedCredentials` variant at line 120, plus the `provisioned_cred_injection_id` plumbing at line 348 and the response shaping at line 372). With the JTI-keyed approach, the preflight contract is unchanged: `Ack` stays, no new request field, no new output kind. DVLS already knows the association-token JTI it minted; that's the identifier.

devolutions-gateway/openapi/gateway-api.yaml
```diff
@@ -1300,6 +1300,14 @@ components:
           allOf:
           - $ref: '#/components/schemas/AppCredential'
           nullable: true
+        cred_injection_id:
```
CBenoit: Revert this `cred_injection_id` field (and the matching one at line 1361 on `PreflightOutput`, plus the `provisioned-credentials` enum value). Falls out of the preflight revert. The .NET client regen will follow automatically — please don't hand-edit it.

devolutions-gateway/src/rdp_proxy.rs
```diff
@@ -393,6 +363,43 @@ where
...
+        return Ok(None);
+    };
+
+    let user = sspi::CredentialsBuffers::AuthIdentity(sspi::AuthIdentityBuffers::from_utf8(
+        &kerberos.service_user_name,
+        &kerberos.realm,
+        kerberos.service_user_password.expose_secret(),
+    ));
+
+    let kdc_url = Url::parse(&format!("http://cred.invalid/{}", credential_entry.cred_injection_id))
```
CBenoit: About the `http://cred.invalid/<id>` trampoline: I see why you reached for it (sspi-rs's CredSSP server config exposes `kdc_url: Option<Url>`, and you needed a way to dispatch a synthetic in-process "KDC"), but the implementation is heavier than the idea. Three concrete asks:

1. **Verify whether sspi-rs offers a custom KDC dispatcher API** (trait object, async closure, `KdcClient`-like abstraction). If yes — use it; the URL trick goes away and `CredentialStoreHandle` no longer needs to be threaded through `perform_credssp_with_client` → `process_authentication_with_client` → `resolve_server_generator` → `send_network_request`.

2. **If no clean API exists in sspi-rs**, that's fine for now (sspi-rs is ours — Devolutions/sspi-rs), but please **open a GitHub issue against sspi-rs** asking for a pluggable KDC dispatcher and link it from a `// TODO(sspi-rs#NNN)` comment here. We can keep a workaround in this PR to unblock; the cleanup lands when sspi-rs ships the API.

3. **Even with the workaround, please localize the interception.** Right now `send_network_request` is a generic Kerberos-network helper that knows about `"cred.invalid"`. Push the loopback dispatch into a non-generic function (closed over `Arc<SessionKerberos>` + `mapping` + `target_hostname` already in scope at `build_credential_injection_server_kerberos_config`), and keep the generic helper unaware of the magic hostname.

Also related: `kdc_proxy_http_client()` (line 653) and the `"http" | "https"` non-loopback branch (line 681) appear to have no real caller in this PR — sspi-rs only emits `tcp://`/`udp://` in the non-injection path and `http://cred.invalid/...` in the injection path. Either delete this branch or document the external use case it's meant to serve (an undocumented "session token can point at arbitrary HTTP host" capability would be a security concern).

devolutions-gateway/src/rdp_proxy.rs
```diff
@@ -112,49 +122,8 @@ where
...
-                user,
-                std::time::Duration::from_secs(*max_time_skew),
-                ticket_decryption_key.clone(),
-            )?,
-        })
-    } else {
-        None
-    };
+    let krb_server_config =
+        build_credential_injection_server_kerberos_config(&credential_entry, &server_dns_name, client_addr)?;
```
CBenoit: The previous code in both `rdp_proxy.rs` and `rd_clean_path.rs` only built the Kerberos server config when `conf.debug.enable_unstable && conf.debug.kerberos.is_some()`. The new `build_credential_injection_server_kerberos_config` is unconditional — every `provision-credentials` session now goes through the fake-KDC path.

That's a real product-surface change not called out in the PR description. Please keep the `enable_unstable`. Promotion to stable should be a separate, deliberate PR with its own CHANGELOG entry under `feat(dgw)`.

devolutions-gateway/src/credential/mod.rs
```diff
@@ -134,30 +246,56 @@ impl CredentialStoreHandle {
...
 }
 
 #[derive(Debug)]
 struct CredentialStore {
     entries: HashMap<Uuid, ArcCredentialEntry>,
+    association_token_index: HashMap<Uuid, Uuid>,
 }
 
 #[derive(Debug)]
 pub struct CredentialEntry {
```
CBenoit, line 157: Nit: `CredentialEntry`, `InjectionState`, `SessionKerberos`, and `build_session_kdc_config` are all `pub` but the only consumer is `rdp_proxy` in the same crate. Restrict to `pub(crate)` (or `pub(super)`) to keep refactor flexibility. Same for the new fields you've added on `CredentialEntry`.

devolutions-gateway/src/credential/tests.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    assert!(
+        store.get(cred_injection_id).is_none(),
+        "expired entry must not be returned by id"
+    );
+    assert!(
+        store.get_by_token(&token).is_none(),
+        "expired entry must not be returned by token"
+    );
+}
```
CBenoit: Test coverage on `CredentialStore` mechanics is good. Three branches in this PR are still uncovered and I think are reachable from in-process tests (no real RDP client needed):

- `kdc_proxy::enforce_realm_token_match` bypass under `disable_token_validation`.
- `kdc_proxy::process_credential_injection` realm-fallback when the envelope realm is missing.
- `rdp_proxy::send_in_process_kdc_request` parses the URL path into a UUID, looks up the entry, and dispatches.

Full end-to-end CredSSP testing would need an actual RDP client and isn't the goal here — that's a welcome future improvement. Please add the easy-to-write ones if time permits; otherwise a tracked follow-up PR is fine.

CBenoit (changes requested):
Thanks for tackling this! This is clearly moving in the right direction, and the test coverage you added on `CredentialStore` is appreciated. That said, I'd like us to be more conservative on three architectural points before this lands:

1. **The `credential` module should not own Kerberos.** `SessionKerberos`, `build_session_kdc_config`, `principal_for_realm`, `kerberos_salt`, `synthetic_realm`, `encrypt_with_kerberos`, and the `dst_hst` extraction inside `insert` all turn what used to be a protocol-neutral store into a Kerberos-aware one. They belong in a new `rdp_proxy::kerberos_session` (or similar) submodule, with `InjectionState` reduced to `{ mapping, target_hostname }` and the target hostname passed to `insert` as a typed parameter from preflight (which already has the validated token).

2. **The new `cred_injection_id` claim/field is avoidable.** The PR's own backwards-compat fallback (`get_by_token` keyed by JTI) demonstrates that the association-token JTI is already a sufficient identifier. I'd like us to take the minimal-surface variant: keep a `jet_cred_id` claim **only on `KdcTokenClaims`**, defined as "the JTI of the associated session token", i.e. it's a *reference* to another token's ID, not a reuse of one. With that, we drop:
   - `jet_cred_id` on `AssociationTokenClaims`
   - `cred_injection_id` request field on `ProvisionCredentialsParams`
   - `PreflightOutputKind::ProvisionedCredentials` (revert to `Ack`)
   - `InsertError::Conflict` and the whole "caller proposes an ID" threat model
   - `association_token_index` + `get_by_token`
   - All the OpenAPI / .NET client additions tied to those

3. **The `enable_unstable` gate was silently removed.** Whether proxy-based Kerberos injection is ready to be promoted out of the unstable gate is a separate decision that deserves its own PR and CHANGELOG entry. Please keep the gate around `build_credential_injection_server_kerberos_config` here. Most likely we’ll simply open a PR once this lands.

A few smaller items in line comments.

devolutions-gateway/src/rdp_proxy.rs
CBenoit: cc @TheBestTvarynka

devolutions-gateway/src/credential/mod.rs
```diff
@@ -4,40 +4,16 @@ mod crypto;
...
-#[derive(Debug)]
-pub enum InsertError {
-    /// The provided token is invalid (e.g., missing or malformed JTI).
-    ///
-    /// This is a client-side error: the caller supplied bad input.
-    InvalidToken(anyhow::Error),
-    /// An internal error occurred (e.g., encryption failure).
-    Internal(anyhow::Error),
-}
-
```
CBenoit, line 29: Why was this removed?

devolutions-gateway/src/credential/mod.rs
CBenoit, line 1: At a glance, it seems to me that this module should be left untouched, in favor of lazily deriving the necessary elements in the `kdc_proxy` module, when necessary. It’s possible to obtain the original token from the `jet_cred_id` claim in the KDC token (it’s in the `CredentialEntry` that we can retrieve from the credential store handle using this ID). It’s then possible to extract the `target_hostname`. This avoids all the eager logic of pre-extracting more stuff than actually necessary. Again, this module should stay protocol-agnostic for the most part, and the new `InjectionState` seems to be designed with RDP in mind. Since it does not seem necessary to store such extra information (we can retrieve that from the token lazily at the most relevant place), we should do so in order to keep the code clean. A side effect of that is that the `preflight` module does not need to change either. A side benefit is that it completely removes some invalid states, making the code overall more straightforward.

devolutions-gateway/src/rdp_proxy.rs
TheBestTvarynka: @CBenoit, currently, the `sspi-rs` does not have such an API. We need to implement it. I created an issue and described the situation there: https://github.com/Devolutions/sspi-rs/issues/664. Feel free to edit and add more context.

@irvingoujAtDevolution, you can replace `TODO(sspi-rs#NNN)` with `TODO(sspi-rs#664` in the code

irvingoujAtDevolution:
@CBenoit, I'm doing an refactoring, I'll pin you once it's all ready and tested

devolutions-gateway/src/credential/mod.rs
irvingoujAtDevolution: removd from credential/ 
refactored everything you mentioned here into credential_injection_kdc.rs

devolutions-gateway/src/token.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/credential/mod.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/api/preflight.rs
irvingoujAtDevolution: updated

devolutions-gateway/openapi/gateway-api.yaml
irvingoujAtDevolution: updated

devolutions-gateway/src/rdp_proxy.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/rdp_proxy.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/credential/mod.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/credential/mod.rs
irvingoujAtDevolution: outdated

irvingoujAtDevolution:
@CBenoit I refactored it a bit, now we have a proper in process kdc that collecting all the pieces together and unifies the logic of in process kdc and in process fake server. credential/mod.rs is now clean

utils/dotnet/Devolutions.Gateway.Utils/src/ProvisionCredentialsRequest.cs
CBenoit, line 1: This can be removed

CBenoit (approved):
LGTM!

PR #1741, opened 2026-04-07
feat(agent): transparent routing through agent tunnel

irvingoujAtDevolution:
## Summary

Transparent routing through QUIC agent tunnel (PR 2 of 4, stacked on #1738).

When a connection target matches an agent's advertised subnets or domains, the gateway automatically routes through the QUIC tunnel instead of connecting directly.

**Depends on:** #1738 (must merge first)

### Changes

- Routing pipeline: explicit agent_id → subnet match → domain suffix (longest wins) → direct
- Integrated into all proxy paths: RDP (clean path), SSH, VNC, ARD, KDC proxy
- `ServerTransport` enum (Tcp/Quic) in `rd_clean_path.rs` for RDP tunnel support
- 7 routing unit tests

### PR stack

1. Protocol + Tunnel Core (#1738)
2. **Transparent Routing** (this PR)
3. Auth + Webapp
4. Deployment + Installer

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1769, opened 2026-04-28
fix(agent): bind QUIC sockets dual-stack with IPv4 fallback

irvingoujAtDevolution:
## Summary

Split out from #1741 per maintainer review. Two related fixes:

- Both ends of the agent-tunnel QUIC handshake used to bind `0.0.0.0` (IPv4 only). On modern Windows/Linux, an agent's DNS resolution for the configured gateway endpoint frequently returns an IPv6 address first (default for `localhost`, common for any name with an AAAA record), and `quinn` refuses to send to that peer over a v4 socket. The tunnel never connects.

- `quinn::Endpoint::server` honors the OS default for `IPV6_V6ONLY`, which is `1` (v6-only) on Linux per RFC 3493. Binding `[::]` without intervention left the listener IPv6-only on Linux, defeating the dual-stack intent.

### What this PR does

- Bind both gateway listener and agent client to `[::]` instead of `0.0.0.0`.
- On the gateway listener, construct the UDP socket via `socket2`, explicitly clear `IPV6_V6ONLY`, then hand the socket to `Endpoint::new`. Linux now genuinely accepts both v4 and v6 agents.
- If the v6 bind fails outright (IPv6 disabled in the host or container — common with `disable_ipv6=1` on Docker), fall back to plain IPv4 with a `warn!` on both ends.

### Test plan

- [x] Smoke-tested locally on Windows: agent connects from `[::1]:*` to gateway on `[::]:4433`.
- [ ] Linux Docker container with `disable_ipv6=1` should now start with a warn and bind v4.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,6 +95,7 @@ dependencies = [
  "rustls-pki-types",
  "serde",
  "sha2 0.10.9",
+ "socket2 0.5.10",
  "time",
  "tokio 1.52.1",
  "tracing",
--- a/crates/agent-tunnel/Cargo.toml
+++ b/crates/agent-tunnel/Cargo.toml
@@ -40,6 +40,7 @@ time = { version = "0.3", default-features = false, features = ["std"] }
 
 # QUIC
 quinn = "0.11"
+socket2 = "0.5"
 
 [dev-dependencies]
 base64 = "0.22"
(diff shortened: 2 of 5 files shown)


CBenoit (approved):
LGTM

CBenoit:
Double check the Copilot comments and feel free to merge yourself

PR #1770, opened 2026-04-29
chore(dgw): remove redundant and unused dependencies

CBenoit:
Remove dependencies added by the agent-tunnel integration that were already provided transitively or not used directly by the gateway: agent-tunnel-proto, rcgen, x509-parser, rustls-pemfile, quinn, rand, ipnetwork, base64, rustls-pki-types, and rustls (accessed via tokio-rustls re-export).

Replace portpicker with UdpSocket::bind("127.0.0.1:0") in the KDC proxy UDP path — simpler and avoids the TOCTOU race inherent to picking a port before binding.

Remove rustls-pki-types from agent-tunnel, using rustls::pki_types directly instead. Remove redundant `use rustls* as _` imports from devolutions-agent/src/main.rs.


CBenoit:
cc @irvingoujAtDevolution

mamoreau-devolutions (approved):

PR #1771, opened 2026-04-29
refactor(agent): rewrite read_cert_chain with picky read_pem

irvingoujAtDevolution:
## Summary

Replaces the manual `find("-----BEGIN ")` + `find("-----END ")` loop in `read_cert_chain` with a `picky::pem::read_pem` loop. Same behaviour (multi-block PEM chain → `Vec<CertificateDer>`), but the previous slice-arithmetic approach was fragile around the END marker's trailing whitespace and could misparse PEM that ends with anything other than a bare `\n`.

## Why split out

This is a small, isolated quality fix that's not part of the routing PR (#1741). Easier to review and land on its own.

## Test plan
- [ ] CI green
- [ ] `cargo test -p testsuite --test integration_tests agent_tunnel::integration::quic_agent_tunnel_e2e` (this test exercises the path; relies on the agent-tunnel feature merging first)


CBenoit (approved):
LGTM

PR #1738, opened 2026-04-02
feat: initial implementation of QUIC agent tunnel

irvingoujAtDevolution:
## Summary

QUIC-based agent tunnel (PR 1 of 4). Agents in private networks connect outbound to Gateway via QUIC/mTLS, advertise reachable subnets and domains, and proxy TCP connections. Pure Rust (Quinn + rustls), zero C dependencies.

See [Technical Spec](https://github.com/Devolutions/devolutions-gateway/pull/1738#issuecomment-4194903683) for protocol details.

### PR stack

1. **Protocol + Tunnel Core** (this PR)
2. Transparent Routing
3. Auth + Webapp
4. Deployment + Installer

### Highlights

- Quinn QUIC transport with mTLS (private PKI)
- CSR-based enrollment (private key never leaves agent)
- Auto-reconnect with exponential backoff
- AD domain auto-detection
- Bounded deserialization, buffer limits, connection limits
- 32 + 15 tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1773, opened 2026-04-29
feat(agent): complete and improve the enrollment flow for Devolutions Agent

irvingoujAtDevolution:
## Summary

The enrollment endpoint used to mint one-time tokens and remember them in an in-memory `HashMap` on the gateway. Those tokens died on restart and only worked against the specific gateway process that minted them, so any deployment with more than one gateway behind a TCP load balancer (or a rolling restart) silently broke. This PR drops the mint endpoint and the `EnrollmentTokenStore` — the agent now presents a JWT signed by DVLS, which the gateway verifies statelessly against the provisioner public key. The scope is renamed `gateway.tunnel.enroll` → `gateway.agent.enroll` (with a matching `gateway.agent.read`) so names reflect what they govern, and a new `EnrollmentClaims` .NET type lets DVLS sign the JWT via `TokenUtils.Sign`. After this, DVLS becomes the sole token authority and enrollment no longer depends on which gateway process or which gateway node the agent reaches.


devolutions-gateway/src/token.rs
```diff
@@ -472,8 +472,10 @@ pub enum AccessScope {
     NetMonitorConfig,
     #[serde(rename = "gateway.net.monitor.drain")]
     NetMonitorDrain,
-    #[serde(rename = "gateway.tunnel.enroll")]
-    TunnelEnroll,
+    #[serde(rename = "gateway.agent.enroll")]
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Decided not to add the alias. Rationale: PR #1738 introduced `TunnelEnroll` recently, the only users of the wire string `gateway.tunnel.enroll` should be DVLS test/dev environments that haven't shipped to production. A clean rename gives us a loud signal — anything still emitting the old name will fail validation and we'll fix the source. An alias would silently keep both working forever, which works against the rationale for renaming in the first place.

devolutions-gateway/src/api/tunnel.rs
```diff
@@ -96,8 +97,12 @@ pub fn make_router<S>(state: DgwState) -> Router<S> {
 
 /// Enroll a new agent.
 ///
-/// Requires a Bearer token matching the configured enrollment secret
-/// or a valid one-time enrollment token from the store.
+/// Requires a Bearer token that is either:
+/// - a JWT signed by the configured provisioner key with `AgentEnroll` /
+///   `Wildcard` scope (issued by DVLS — the only authority for agent
+///   enrollment tokens), or
```
CBenoit: issue:  You can mention DVLS as an example of provisioner, but technically any provisioner can work (another example is Hub / PEM service).

devolutions-gateway/src/api/tunnel.rs
```diff
@@ -96,8 +97,12 @@ pub fn make_router<S>(state: DgwState) -> Router<S> {
...
+/// - the static `enrollment_secret` from the gateway configuration (admin
+///   bootstrap fallback for environments without DVLS).
```
CBenoit: question: Is it a debug option? I think it’s only useful for testing, and even then we can still generate tokens using tokengen.

suggestion: In a follow-up PR, consider dropping this option.

devolutions-gateway/src/extract.rs
```diff
@@ -408,7 +408,9 @@ where
 
 /// Grants read access to agent management endpoints.
 ///
-/// Accepts a scope token with `DiagnosticsRead`, `ConfigWrite`, or `Wildcard` scope.
+/// Accepts a scope token with `AgentRead`, `DiagnosticsRead`, `ConfigWrite`, or
+/// `Wildcard` scope. `DiagnosticsRead` / `ConfigWrite` are kept for
+/// back-compat with callers that predate the dedicated `AgentRead` scope.
```
CBenoit: issue: We don’t need back-compatibility yet, the product never shipped yet.

irvingoujAtDevolution: updated

devolutions-gateway/src/api/tunnel.rs
irvingoujAtDevolution: updated

devolutions-gateway/src/api/tunnel.rs
irvingoujAtDevolution: updated

CBenoit (approved):
LGTM

PR #1774, opened 2026-04-29
feat(agent): --advertise-domains CLI flag

irvingoujAtDevolution:
## Summary

Master already plumbs `tunnel_conf.advertise_domains` into the agent's `RouteAdvertise` control message and the gateway's domain-suffix routing table, but there was no way to set the list from the CLI — the field was config-file-only. This PR threads it through both CLI entry points: the `up` subcommand gains a `--advertise-domains` flag (parallel to `--advertise-routes`), and the legacy positional `enroll` command gains a 6th positional arg (`<gateway_url> <token> <agent_name> [subnets_csv] [domains_csv]`). An empty list from the caller preserves whatever is already in the config so an admin who manages domains via the config file does not lose them on re-enrollment from the CLI.


irvingoujAtDevolution:
Closing per discussion — --advertise-domains direction dropped.

PR #1775, opened 2026-04-29
feat(agent): implement certificate renewal

irvingoujAtDevolution:
## Summary

The `CertRenewalRequest` / `CertRenewalResponse` control messages were defined back in #1738 but both ends only had stubs (`"not supported in this build"`). This PR wires up the real handlers so an agent's certificate can be renewed without a full re-enrollment, before its 1-year cert expires and locks the agent out.

On the **gateway** side, the listener now signs the agent's renewal CSR with the existing `CaManager::sign_agent_csr` and returns the new cert + CA bundle. Identity is taken from the mTLS-authenticated peer cert, never from the CSR subject — only the public key gets re-bound. On the **agent** side, `run_single_connection` runs a one-shot renewal check right after the QUIC handshake: if the current cert is within 15 days of expiry, it sends a `CertRenewalRequest`, persists the new cert + CA, closes the connection, and surfaces a `ConnectionOutcome::CertRenewed` that tells the outer reconnect loop to re-handshake immediately with the new cert.

This is deliberately a per-reconnect check rather than a periodic in-session timer. With a 120s idle timeout, 15s keep-alive, and the variety of network blips / VPN drops / host sleep / gateway restarts that hit any real deployment, the connection comes back through this path far more often than once every 15 days — so the periodic timer would only fire in scenarios that don't exist in practice.

### DIFF
--- a/crates/agent-tunnel/src/listener.rs
+++ b/crates/agent-tunnel/src/listener.rs
@@ -111,6 +111,7 @@ pub struct AgentTunnelListener {
     endpoint: quinn::Endpoint,
     registry: Arc<AgentRegistry>,
     agent_connections: Arc<RwLock<HashMap<Uuid, quinn::Connection>>>,
+    ca_manager: Arc<CaManager>,
 }
 
 impl AgentTunnelListener {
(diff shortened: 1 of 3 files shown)


CBenoit (approved):
LGTM

PR #1778, opened 2026-05-08
feat(dgw): refuse JREC push when recording storage is full

thenextman:
The goal is to provide richer information to the client when session recording fails due to a storage issue on the Gateway. 

Current behaviour is that the web socket upgrade is accepted unconditionally and then fails mid-stream with a generic 1011 server error close ("forwarding_failure").

With these changes:

- Before web socket upgrade the existing recording health probe is run and returns `HTTP 507 Insufficient Storage` if:
    - The recording directory is not writeable
    - Free space is below the new `min_recording_storage_free_space` threshold (default `None`)
- Mid stream, if `io::copy` returns `ErrorKind::StorageFull` the web socket is closed with private code `4010`. Other failures still close with `1011`.
- Adds `CloseWebSocketHandle::app_close(code)` for sending private range close codes (4000-4999) and a `PushOutcome` enum so `ClientPush::run` can report distinctly without overloading `anyhow::Error`.

For existing clients the changes are safe. `min_recording_storage_free_space` defaults to `None` so pre-flight rejections only happen when the storage is already unwritable. Existing code would have failed slightly later during file open with a generic close. The mid-stream change is irrelevant to existing clients (Cadeau, web) which collapse any error to a generic one on that side.

⚠️ These changes were engineered with Claude and made "blind". The goal is to get this into a Gateway release and then make the corresponding Cadeau (client) improvements.

In follow up we willl update the OpenAPI document for the new 507 response and improve Cadeau and web-side error handling.

Issue: DGW-382


devolutions-gateway/src/config.rs
```diff
@@ -1687,6 +1687,13 @@ pub mod dto {
         #[serde(skip_serializing_if = "Option::is_none")]
         pub recording_path: Option<Utf8PathBuf>,
 
+        /// Minimum free space (in bytes) required on the recording storage volume before a
+        /// new JREC push stream is accepted. When set, the gateway returns HTTP 507
+        /// Insufficient Storage on `/jet/jrec/push/{id}` if free space is below this value.
+        /// Omit to disable the pre-flight check.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub min_recording_storage_free_space: Option<u64>,
```
CBenoit: suggestion: Also document this new option in the README.md file

question: Does "omit to disable the pre-flight check" means that the /jet/jrec/push/{id} route will stop returning HTTP 507 error? I think it’s easy to misunderstand this with the preflight endpoint.

devolutions-gateway/src/api/jrec.rs
```diff
@@ -65,6 +66,52 @@ async fn jrec_push(
...
+    // depends on the path being present.
+    if let Err(error) = tokio::fs::create_dir_all(&conf.recording_path).await {
+        warn!(
+            client = %source_addr,
+            %session_id,
+            %error,
+            "Refusing JREC push: failed to ensure recording storage directory"
+        );
+        return Err(HttpErrorBuilder::new(StatusCode::INSUFFICIENT_STORAGE).msg("recording storage is not accessible"));
+    }
```
CBenoit: suggestion: Instead of lazily creating the folder in multiple places, we could create it once on start

devolutions-gateway/src/api/jrec.rs
```diff
@@ -110,14 +157,30 @@ async fn handle_jrec_push(
...
+/// Application-specific WebSocket close codes for the `/jrec/push` endpoint.
+///
+/// Codes in 4000-4999 are reserved for private use per
+/// <https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code>. Clients should switch on
+/// these numeric codes; the textual close reason is informational and may change.
+#[repr(u16)]
+enum JrecPushCloseCode {
+    /// The recording storage volume is full and the stream cannot continue.
+    StorageFull = 4010,
+}
```
CBenoit: suggestion: I see no value into using an enum for that. Could be a u16 constant since the only uses look like `StorageFull as u16`. We don’t use the extra type for anything meaningful.

devolutions-gateway/src/api/jrec.rs
```diff
@@ -65,6 +66,52 @@ async fn jrec_push(
...
+        );
+        return Err(HttpErrorBuilder::new(StatusCode::INSUFFICIENT_STORAGE).msg("recording storage is not accessible"));
+    }
+
+    // The probe performs synchronous filesystem I/O; run it on the blocking pool so we
+    // don't stall a Tokio worker thread under load.
+    let probe_path = conf.recording_path.clone();
+    let storage = tokio::task::spawn_blocking(move || recording_storage_health(probe_path.as_std_path()))
+        .await
+        .map_err(HttpError::internal().with_msg("recording storage probe panicked").err())?;
```
CBenoit: question: Is it something that may take more than 200ms? It’s not like we’re reading the contents of a file

thenextman: Claude and copilot went back and forth on this, I had him put it back to the original sync call like in get_heartbeat.

devolutions-gateway/src/config.rs
```diff
@@ -1687,6 +1687,13 @@ pub mod dto {
         #[serde(skip_serializing_if = "Option::is_none")]
         pub recording_path: Option<Utf8PathBuf>,
 
+        /// Minimum free space (in bytes) on the recording storage volume below which the
+        /// gateway returns HTTP 507 Insufficient Storage on `/jet/jrec/push/{id}`. Omit to
+        /// skip this threshold check; the gateway can still return 507 independently when
+        /// the recording storage is not writable.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub min_recording_storage_free_space: Option<u64>,
```
CBenoit, line 1695: question: Also, is there a reason to disable this entirely by default? Do you think there is no sensible default value we could have instead? We can also change how the default configuration file is generated to include a default value there instead, and keep previous behavior for old configurations (similar to the StrictTls option).

CBenoit (approved):
LGTM! I added a question, but the code is looking okay to me

devolutions-gateway/src/config.rs
thenextman: @CBenoit Actually I think moving forward we should place a reasonable default here (1GB, something like that?). In current iteration I wanted to avoid any kind of breaking change on the client; until we have the clients updated (so user is actually told "not enough storage" when they try to record). There are 3 moving pieces to update and I'm not sure how quickly I get to one after the other.

What do you think?

CBenoit: That’s true! In that case, I think the current code is good as-is, and we can treat the new option with the conservative default as a way for us to test the code in  RDM / DVLS, and then place a reasonable default (after a few releases?).

devolutions-gateway/src/api/jrec.rs
thenextman: I made the change but now I've back it out and gone back to lazy. Honestly I'm not sure which is better; should we be able to create/mount the recording storage with Gateway running? Currently you can, and get safe heartbeat checks... I think some more consideration needed before changing this on a whim.

devolutions-gateway/src/config.rs
thenextman: Exactly! Thanks

PR #1783, opened 2026-05-14
ci: bump macOS runners to macos-15 and drop CommandLineTools removal

thenextman:
Github have [shipped](https://github.com/actions/runner-images/issues/14097) a broken Homebrew `rustup` in their latest runner images. This PR adds a workaround for now.

Other changes

- I note that the maos-14 runners will begin deprecation soon, so take the opportunity to move to macos-15
- The macOS build was running `sudo rm -rf /Library/Developer/CommandLineTools`. This seems to be an artifact of when cross-compiled arm64 from an Intel runner a long time ago. It should not be needed. When we switched to arm64 runners (and Intel became the cross-compile target) it was kept for cargo cult reasons.

### DIFF
--- a/.github/workflows/build-tools.yml
+++ b/.github/workflows/build-tools.yml
@@ -17,7 +17,7 @@ jobs:
           - platform: unknown-linux-gnu
             runner: ubuntu-22.04
           - platform: apple-darwin
-            runner: macos-14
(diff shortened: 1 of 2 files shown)


CBenoit (approved):
Thank you!

PR #1786, opened 2026-05-19
feat(dgw): add jet_cred_id KDC claim

CBenoit:
Introduces the `jet_cred_id` KDC claim and a typed `KdcDestination` wrapper at the token-claims layer, plumbed through the Gateway, `tokengen` and the `Devolutions.Gateway.Utils` NuGet package.

A KDC session token now has two mutually exclusive shapes:

* `{ krb_realm, krb_kdc }` — forward traffic to a real upstream KDC (existing behavior, no wire change).
* `{ jet_cred_id }` — serve KDC traffic locally using the credentials provisioned at session establishment, identified by the JTI of the access token that carries them.

The new `KdcDestination::Real | Inject` enum keeps `TargetAddr` semantically pure (only real network addresses, schemes `tcp`/`udp`) while making the dispatch decision explicit at the type level.

Dispatch of the `Inject` variant in `/jet/KdcProxy` is intentionally deferred to a follow-up; tokens of this shape are accepted and validated but currently rejected at request time with a clear error.

### DIFF
--- a/devolutions-gateway/src/api/kdc_proxy.rs
+++ b/devolutions-gateway/src/api/kdc_proxy.rs
@@ -13,7 +13,7 @@ use tokio::net::{TcpStream, UdpSocket};
 use crate::DgwState;
 use crate::http::{HttpError, HttpErrorBuilder};
 use crate::target_addr::TargetAddr;
-use crate::token::AccessTokenClaims;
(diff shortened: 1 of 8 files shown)


irvingoujAtDevolution (approved):
LGTM!

PR #1789, opened 2026-05-21
feat(agent-installer): add Agent Tunnel configuration dialog

irvingoujAtDevolution:
## Summary

Adds an optional **Agent Tunnel** wizard step to the Devolutions Agent MSI installer so admins can enrol the agent in a Gateway QUIC tunnel during install — UI or unattended.

Three text fields in the dialog: enrolment string, advertise subnets, advertise domains. The feature is opt-in (`isEnabled: false, allowChange: true`); the dialog is skipped when the feature isn't selected, and an empty enrolment string skips tunnel setup even if the feature is on, so the installer remains usable for non-tunnel deployments.

Smoke-tested end-to-end against a local DVLS + Gateway + Agent stack: MSI built clean, agent enrolled successfully, RDP TCP traffic routed through the agent tunnel to `IT-HELP-DC:3389`.

## MSI public properties (unattended install)

```
msiexec /i DevolutionsAgent.msi /qn ADDLOCAL="...,AgentTunnel" \
  AGENT_TUNNEL_ENROLLMENT_STRING="dgw-enroll:v1:<base64>" \
  AGENT_TUNNEL_ADVERTISE_SUBNETS="10.10.0.0/24" \
  AGENT_TUNNEL_ADVERTISE_DOMAINS="corp.example.com"
```

## Custom action

`EnrollAgentTunnel` is deferred + elevated + `Impersonate=false`, runs `Before StartServices`, gated by `Features.AGENT_TUNNEL_FEATURE.BeingInstall()`. It:

1. Validates the `dgw-enroll:v1:` prefix, strips whitespace, pads base64url, decodes JSON payload (`api_base_url`, `enrollment_token`, optional `name`).
2. Shells out to `devolutions-agent.exe enroll <url> <token> <name> [subnets]` with a 60s timeout; on timeout the child is killed.
3. Token is redacted (`***`) symmetrically across the command-line log and child stdout/stderr.
4. After enrolment, advertise domains are patched into `Tunnel.AdvertiseDomains` of `%ProgramData%\Devolutions\Agent\agent.json`. This matches the design that advertise domains live in config, not on the CLI (replaces the closed #1774).

## Depends on

- #1773 — DVLS-signed JWT enrolment flow (`/jet/tunnel/enroll`)
- #1775 — agent cert renewal (referenced by the bundled agent binary)

Both merged to master 2026-05-20.

## Test plan

- [x] Cold MSI build clean, no errors (8 preexisting `CNDL1138`/`CNDL1006` warnings from existing code, not introduced here)
- [x] Wizard order: Welcome → Features → AgentTunnel → InstallDir → VerifyReady
- [x] AgentTunnel dialog skipped when feature not selected
- [x] Empty enrolment string skips enrolment (no failure)
- [x] Invalid `dgw-enroll:v1:` prefix → validation error in dialog
- [x] Invalid base64 → validation error in dialog (caught client-side, not at CA time)
- [x] Smoke test against local DVLS-signed enrolment JWT — agent registered, cert issued
- [x] Smoke test routing: RDP TCP through agent tunnel to `IT-HELP-DC:3389` succeeds

## Review history

Pre-review and codex-review passes folded into the single commit. Notable points addressed:

- Robust JSON payload parsing (`JObject`, `Value<string>()`, `IsNullOrWhiteSpace` validation)
- Symmetric token redaction (cmdline + stdout + stderr)
- `WaitForExit(60_000)` return-value check with child kill on timeout
- Whitespace-stripped + padded base64url decoding on both dialog validation and CA
- ADDLOCAL CSV value trimming for admin-supplied lists with whitespace
- Dropped 2 stray `InstallDirDlg*` strings leaked into en-us
- Added 5 missing fr-fr translations


mamoreau-devolutions (approved):

package/AgentWindowsManaged/Actions/AgentActions.cs
```diff
@@ -279,6 +279,28 @@ internal static class AgentActions
...
+        new Id($"CA.{nameof(enrollAgentTunnel)}"),
+        CustomActions.EnrollAgentTunnel,
+        Return.check,
+        When.Before, Step.StartServices,
+        Features.AGENT_TUNNEL_FEATURE.BeingInstall(),
+        Sequence.InstallExecuteSequence)
+    {
+        Execute = Execute.deferred,
+        Impersonate = false,
+        // Deferred CAs only see properties bubbled through CustomActionData. The Set_<CA>_Props
```
thenextman, line 292: This is just documented Windows Installer, I'd remove this comment

package/AgentWindowsManaged/Actions/AgentActions.cs
```diff
@@ -279,6 +279,28 @@ internal static class AgentActions
...
+        // immediate action expands [PROP] for each entry below before the deferred CA runs.
+        UsesProperties = string.Join(";", new[]
```
thenextman, line 294: There's a helper function that provides error checking, use e.g.

`UsesProperties = UseProperties(new [] { AgentProperties.installId })`

package/AgentWindowsManaged/Actions/AgentActions.cs
```diff
@@ -279,6 +279,28 @@ internal static class AgentActions
...
+        {
+            AgentProperties.AgentTunnelEnrollmentString,
+            AgentProperties.AgentTunnelAgentName,
+            AgentProperties.AgentTunnelAdvertiseSubnets,
+            AgentProperties.AgentTunnelAdvertiseDomains,
+            AgentProperties.InstallDir,
+        }.Select(p => $"{p}=[{p}]")),
+    };
```
thenextman, line 302: I'm not exactly sure what is being passed in these properties, but are you certain it doesn't/won't contain a semi-colon? Semi-colons break `CustomActionData` and need special handling. This is a known Windows Installer quirk. If you're sure it's ok, that's fine; if you're not I can explain how to fix it.

package/AgentWindowsManaged/Actions/CustomActions.cs
```diff
@@ -318,6 +320,196 @@ public static ActionResult SetFeaturesToConfigure(Session session)
...
+            {
+                session.Log(msg);
+                using Record record = new(0) { FormatString = msg };
+                session.Message(InstallMessage.Error, record);
+                return ActionResult.Failure;
+            }
+
+            if (enrollmentString.Length == 0)
+            {
+                return Fail("An enrollment string is required. Paste the enrollment string provided by your gateway operator, or deselect the Agent Tunnel feature.");
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Not acceptable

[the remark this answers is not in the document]
irvingoujAtDevolution: Hi — quick correction on this one. We initially acted on this suggestion and switched the empty-string case to a no-op (skip), but we've since **reverted it**: an empty enrollment string keeps failing the install intentionally.

The "non-tunnel / unattended deployments" concern doesn't actually apply here. `EnrollAgentTunnel` only runs under the `AGENT_TUNNEL_FEATURE.BeingInstall()` condition, and the Agent Tunnel feature is **opt-in** (Level 2 — not installed by default). A deployment that doesn't select the tunnel feature never reaches this code, so it can't trigger the failure. The only way to hit the empty-string branch is to *explicitly* select the Agent Tunnel feature (`ADDLOCAL=...,F.Tunnel` or `ALL`) while omitting the enrollment string — which is a misconfiguration we'd rather surface than silently skip.

(The genuinely separate bug nearby — `ADDLOCAL=ALL` wrongly skipping the dialog — is fixed in #1804.)

PR #1781, opened 2026-05-12
feat(dgw): route KDC traffic through agent tunnel

irvingoujAtDevolution:
Closes [DGW-384](https://devolutions.atlassian.net/browse/DGW-384).

Routes KDC traffic through the agent tunnel for the two remaining paths after [#1741](https://github.com/Devolutions/devolutions-gateway/pull/1741):

- HTTP `/jet/KdcProxy` endpoint
- RDP CredSSP/NLA (`rdp_proxy.rs::send_network_request`)

`send_krb_message` gains `(agent_tunnel_handle, session_id: Uuid)`. RDP callers pass `claims.jet_aid` so agent-side logs correlate KDC sub-traffic with the parent RDP session; the HTTP handler mints a fresh UUID since its token has no parent association.

**Depends on #1741 — must merge first** (uses `agent_tunnel::routing::try_route`).

[DGW-384]: https://devolutions.atlassian.net/browse/DGW-384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1792, opened 2026-05-21
feat(dgw): route KDC traffic through agent tunnel

irvingoujAtDevolution:
Closes [DGW-384]. Replaces #1781 (reverted in #1791).

Routes KDC traffic through the agent tunnel for the two paths still missing after #1741:

- HTTP `/jet/KdcProxy`
- RDP CredSSP/NLA (`rdp_proxy.rs::send_network_request`)

## How

A new `KdcConnector` (`src/kdc_connector.rs`) bundles the three routing inputs — `session_id`, `explicit_agent_id`, `agent_tunnel_handle` — into one value and delegates every send to `agent_tunnel::routing::try_route`. Callers no longer pre-decide direct vs. tunnel; the routing pipeline decides, and its existing pin-without-tunnel enforcement is preserved.

CredSSP plumbing (`perform_credssp_as_*` → `resolve_*_generator` → `send_network_request`) now takes a single `&KdcConnector` instead of three primitives.

## Session correlation

- RDP CredSSP uses the parent association's `claims.jet_aid`.
- HTTP `/jet/KdcProxy` uses the KDC token's own `jti`. `KdcTokenClaims` now exposes `jti` via its serde helper, matching every other `*TokenClaims` type. `KdcToken` stays a tuple struct.

## Guards

- `explicit_agent_id` (from `jet_agent_id`) is forwarded to `try_route`, so a token-pinned session can never silently fall back to a different agent or to direct connect.
- UDP-via-agent rejection: when `try_route` selects an agent for a `udp://` target, the connector errors instead of silently downgrading (the tunnel only carries TCP today).
- 64 KiB cap on TCP-framed KDC reply length, with overflow-safe math.

## Tests

`kdc_connector::tests` covers the two cases that don't need a live `AgentTunnelHandle`: pin-without-tunnel (must error) and no-pin-no-tunnel (falls through to direct). The remaining cases (pin-with-missing-agent, no-match-falls-back, tunnel success, UDP-via-agent guard) need an integration-style fixture that stands up a real listener; left as a follow-up.

## No-op for external consumers

JWT/wire format unchanged. Agent-tunnel `ConnectRequest` wire protocol unchanged. DVLS-side: no action required.

Issue: DGW-384

[DGW-384]: https://devolutions.atlassian.net/browse/DGW-384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/devolutions-gateway/src/api/kdc_proxy.rs
+++ b/devolutions-gateway/src/api/kdc_proxy.rs
@@ -1,12 +1,8 @@
-use std::io;
-
 use axum::Router;
 use axum::extract::State;
-use axum::http::StatusCode;
 use axum::routing::post;
 use picky_krb::messages::KdcProxyMessage;
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
-use tokio::net::{TcpStream, UdpSocket};
+use uuid::Uuid;
 
 use crate::DgwState;
 use crate::credential_injection_kdc::{
@@ -14,7 +10,8 @@ use crate::credential_injection_kdc::{
     kdc_proxy_message_realm,
 };
 use crate::extract::KdcToken;
-use crate::http::{HttpError, HttpErrorBuilder};
+use crate::http::HttpError;
+use crate::kdc_connector::KdcConnector;
 use crate::target_addr::TargetAddr;
 use crate::token::{KdcDestination, KdcTokenClaims};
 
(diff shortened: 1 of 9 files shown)


CBenoit (approved):
LGTM! Address Copilot suggestions and feel free to merge!

PR #1793, opened 2026-05-22
fix(jetsocat): embed Windows icon and version metadata

mamoreau-devolutions:
Embeds the Devolutions Gateway icon and Windows version metadata into jetsocat.exe on Windows so the binary has branded iconography and standard file properties in Explorer.

image

this will make automatic app icon extraction easier for jetsocat


mamoreau-devolutions:
Implementation notes:
- Adds a Windows-only build script to jetsocat and compiles both a generated VERSIONINFO resource and a static icon resource.
- Reuses package/WindowsManaged/Resources/DevolutionsGateway.ico so jetsocat.exe gets the existing Devolutions Gateway icon without duplicating assets.
- Uses CARGO_PKG_VERSION to generate the file and product version metadata embedded into the executable.
- Adds embed-resource as a Windows-only build dependency and updates Cargo.lock accordingly.

Validation:
- cargo build -p jetsocat -q

thenextman (approved):

PR #1794, opened 2026-05-22
feat(dgw): decouple agent tunnel SAN from network endpoint

irvingoujAtDevolution:
## Summary

First of three PRs implementing the agent-tunnel identity refactor
described in `AGENT_TUNNEL_IDENTITY_DESIGN.md`. This PR is **gateway-only**.

Today the gateway's `conf.hostname` is overloaded as both the cryptographic
identity written into the agent-tunnel server cert SAN *and* the network
hostname returned to agents as their QUIC dial target. In any realistic
deployment a single gateway is reachable via multiple names (internal FQDN,
LAN IP literal, public DNS), so picking one and forcing every agent to use
it breaks agents whose network can't resolve that chosen name.

Scope (per spec PR 1):

- New `AgentTunnel.AdvertisedNames` config: `Vec<AdvertisedName>` with serde
  untagged string-or-object form (label is informational).
- Default to `[conf.hostname]` when absent so existing single-name
  deployments keep working with no admin changes.
- Multi-SAN server cert generation: regenerate `agent-tunnel-server-cert.pem`
  at boot when SAN set differs from config, reusing the existing keypair so
  SPKI pins remain stable for enrolled agents.
- Compute `EnrollResponse.quic_endpoint` from the validated `jet_gw_url.host`
  rather than `conf.hostname`. Add `quic_port` field for new agents (the
  legacy `quic_endpoint` is retained one release for compatibility).
- Reject enrollment with HTTP 400 + structured `{error, message, help}` when
  `jet_gw_url.host` is not in `AdvertisedNames`.
- Extend `/jet/diagnostics/configuration` with an `agent_tunnel` field so
  DVLS can introspect via the existing endpoint instead of a new route.

## Dependency

This PR is the base for #pr2 (agent) and #pr3 (installer).

## Spec

See `AGENT_TUNNEL_IDENTITY_DESIGN.md` (PR 1 section).

## Test plan

- [ ] `cargo test -p devolutions-gateway` passes
- [ ] SAN regen idempotence + host normalization unit tests
- [ ] Manual: configure two AdvertisedNames, enroll via name1, verify cert
      SAN contains both names, reject enrollment via name3

### DIFF
--- a/crates/agent-tunnel/src/cert.rs
+++ b/crates/agent-tunnel/src/cert.rs
@@ -250,34 +250,62 @@ impl CaManager {
 
     /// Ensure a server certificate exists for the QUIC listener (signed by our CA).
     ///
+    /// `advertised_names` is the authoritative list of names/IPs the agent tunnel
+    /// is reachable as. Each entry is added to the cert SAN — DNS literals are
+    /// inserted as `SanType::DnsName`, IP literals (parseable by [`std::net::IpAddr`])
+    /// as `SanType::IpAddress`. When the on-disk cert's SAN set differs from the
+    /// expected SAN set, the cert is regenerated reusing the existing keypair so
+    /// the SPKI pin captured at enrollment stays stable.
+    ///
     /// Returns `(cert_path, key_path)` on disk.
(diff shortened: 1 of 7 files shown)


irvingoujAtDevolution:
Closing — not authorized; will be reopened after explicit owner approval.

PR #1801, opened 2026-05-26
fix(agent-installer): backfill Agent Tunnel localization strings to JSON

irvingoujAtDevolution:
## Summary

The 11 Agent Tunnel localization strings added in #1789 were hand-edited directly into the generated `.wxl` files without corresponding entries in the JSON sources (`Strings_en-US.json` / `Strings_fr-FR.json`). The next `Strings.g.tt` regeneration would silently delete them — exactly the failure mode Richard flagged.

This backfills all 11 keys to both JSON sources and regenerates.

## Review note

- Real change: JSON sources (+11 each) and the 11 new `Strings.g.cs` constants.
- The `.wxl` diffs are **pure reordering** — the generator sorts by `Id` within each group. Id set vs master is +0/-0 (no keys added/removed, no text changed); those keys already existed in the `.wxl`.
- Regenerated via VS "Run Custom Tool" on `Strings.g.tt`.

No user-visible change.

Changelog: ignore


CBenoit (approved):
Thank you!

PR #1802, opened 2026-05-26
fix(agent): systematic failure recovery for agent tunnel enrollment

irvingoujAtDevolution:
## Summary

Agent tunnel enrollment spans two non-transactional write phases — the agent's `up` (Rust: client cert/key + the fixed-name `gateway-ca.pem` + `agent.json`) and the MSI custom action (advertise subnets/domains + rollback bookkeeping). A failure in either phase left the machine in a partial state: orphaned cert files, a clobbered `gateway-ca.pem`, or a half-written `agent.json` that even a rollback couldn't recover from.

This makes the whole enrollment **recoverable end to end** — every failure path leaves the machine exactly as it was before enroll.

## Agent (Rust)

- `persist_enrollment_response` is now transactional: load/validate the config **before** any write (a corrupt `agent.json` fails before touching disk), back up the fixed-name `gateway-ca.pem`, and roll back partial cert/CA writes on any failure.
- `save_config` creates its parent directory (fixes fresh standalone `agent.exe up` on a clean machine) and writes **atomically** (temp + rename) so a mid-write failure never truncates `agent.json`.

## Installer (C#)

- `EnrollAgentTunnel` snapshots the pre-enrollment `Tunnel` section and `gateway-ca.pem` into a per-install **rollback marker** (`%TEMP%\{installId}-tunnel-rollback.json`), written atomically. The marker is **required**: if it can't be recorded, the enrollment is undone inline and the CA fails.
- New marker-driven `RollbackEnrollAgentTunnel` (`Execute.rollback`) only cleans up / restores when *this* install recorded a marker — so it never touches pre-existing or partial state. It restores the original `Tunnel` section and `gateway-ca.pem`, and deletes the certs this install wrote.
- All `agent.json` writes go through an atomic temp-replace helper, so the rollback can always re-parse it.
- Drains stdout/stderr concurrently with `WaitForExit` (fixes a pipe-buffer deadlock that could kill a healthy `up`) and fails loudly when operator-supplied advertise subnets/domains can't be persisted (instead of silently dropping them).

## Failure-recovery matrix

| Failure point | Recovery |
|---|---|
| Empty / pre-write failure | No marker → rollback no-op; nothing was written |
| `up` fails mid-write (bad json / save error) | Rust self-rolls-back its partial writes → non-zero exit → no marker |
| `up` ok, marker write fails | Inline cleanup + CA fails |
| `up` ok, advertisements write fails | Marker present → rollback restores |
| Later MSI action fails after success | Marker present → rollback restores |

## Test

- `cargo check -p devolutions-agent` — clean.
- `dotnet build DevolutionsAgent.csproj -c Debug` — 0 errors (8 pre-existing WiX CNDL warnings), MSI builds.

Changelog: ignore

### DIFF
--- a/devolutions-agent/src/config.rs
+++ b/devolutions-agent/src/config.rs
@@ -237,7 +237,25 @@ impl ConfHandle {
 pub fn save_config(conf: &dto::ConfFile) -> anyhow::Result<()> {
     let conf_file_path = get_conf_file_path();
     let json = serde_json::to_string_pretty(conf).context("failed JSON serialization of configuration")?;
-    std::fs::write(&conf_file_path, json).with_context(|| format!("failed to write file at {conf_file_path}"))?;
+
+    // Ensure the parent directory exists — a fresh machine (standalone `agent.exe up`, no MSI) may
+    // not have the data directory yet, and a bare write would fail before any config is created.
+    if let Some(parent) = conf_file_path.parent().filter(|parent| !parent.as_str().is_empty()) {
+        std::fs::create_dir_all(parent)
+            .with_context(|| format!("failed to create configuration directory: {parent}"))?;
+    }
+
+    // Write atomically: serialize to a sibling temp file, then rename over the target. A failure
+    // mid-write therefore never truncates or corrupts an existing agent.json — the original stays
+    // intact until the rename atomically replaces it (std::fs::rename replaces on Windows too).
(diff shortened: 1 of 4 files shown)


CBenoit (approved):
LGTM, but address Copilot’s relevant comments before merging

mamoreau-devolutions (approved):

PR #1806, opened 2026-06-01
fix(agent): improve enrollment token validation

irvingoujAtDevolution:


devolutions-gateway/src/api/tunnel.rs
```diff
@@ -243,10 +148,10 @@ async fn get_agent(
 
 /// Delete (unregister) an agent by ID.
 async fn delete_agent(
+    _access: AgentManagementWriteAccess,
```
CBenoit, line 151: thought: We need a dedicated scope for deletion, please send a follow up PR

utils/dotnet/Devolutions.Gateway.Utils/src/EnrollmentClaims.cs
```diff
@@ -28,7 +24,7 @@ public class EnrollmentClaims : IGatewayClaims
     /// Optional agent display-name hint.
     ///
     /// The gateway never reads this claim — it only verifies the JWT
-    /// signature, scope, and expiry on <c>POST /jet/tunnel/enroll</c>; the
+    /// signature, content type, and expiry on <c>POST /jet/tunnel/enroll</c>; the
```
CBenoit, line 27: question: Why not let the Gateway read the claim instead of relying on EnrollRequest body? The information is not supposed to be tampered with

utils/dotnet/Devolutions.Gateway.Utils/src/EnrollmentClaims.cs
```diff
@@ -39,14 +35,20 @@ public class EnrollmentClaims : IGatewayClaims
     /// it here lets DVLS pre-fill the name the admin typed in the "Generate
     /// Enrollment String" dialog so the agent shows up under that name in
     /// the registry without an extra CLI flag.
+    ///
+    /// Name resolution order at install time is: explicit name (CLI
+    /// <c>--name</c> / installer dialog value) &gt; this <c>jet_agent_name</c>
+    /// claim &gt; the local computer name. The computer-name fallback is
+    /// applied by the Windows installer when neither an explicit name nor
+    /// this claim is present; if it is omitted, the agent CLI's own
+    /// <c>up</c> command instead requires <c>--name</c> to be passed.
```
CBenoit, line 44: suggestion: Let’s make the claim non-optional and simplify the whole logic here.

CBenoit (approved):
Approving and merging for velocity. I’m expecting follow up PRs to fix the remaining flagged issues.

irvingoujAtDevolution:
> Approving and merging for velocity. I’m expecting follow up PRs to fix the remaining flagged issues.

they will arrive soon!

PR #1807, opened 2026-06-01
refactor: clarify agent delete access scope

irvingoujAtDevolution:
Renames the agent management delete extractor so the Gateway delete-agent route clearly reflects that it requires \gateway.agent.delete\. Issue: #1806


CBenoit (approved):
LGTM

PR #1808, opened 2026-06-01
fix(agent): remove optional agent name from installer side

irvingoujAtDevolution:
The signed `jet_agent_name` enrollment claim is now authoritative and required, so the installer's optional Agent Name dialog field, the `--name`/`--gateway` CLI flags, and the request-body agent name are all removed. The agent and gateway both read the name (and gateway URL) from the JWT itself, and the gateway stamps the signed name into the client certificate CN and the agent registry. Validated end-to-end on Hyper-V (gateway on IT-HELP-GW, MSI UI install on Win10): the agent enrolls and the tunnel connects with the registered name sourced solely from the token.

### DIFF
--- a/devolutions-agent/src/main.rs
+++ b/devolutions-agent/src/main.rs
@@ -55,7 +55,6 @@ const START_FAILED_ERR_CODE: u32 = 2;
 struct UpCommand {
     gateway_url: String,
     enrollment_token: String,
-    agent_name: String,
     advertise_subnets: Vec<String>,
 }
 
@@ -149,8 +148,6 @@ fn parse_up_command_args(args: &[String]) -> Result<UpCommand> {
 
 fn parse_up_command_args_with_reader<R: BufRead>(args: &[String], mut stdin_reader: R) -> Result<UpCommand> {
     let mut gateway_url = None;
-    let mut enrollment_token = None;
-    let mut agent_name = None;
     let mut enrollment_string = None;
     let mut advertise_subnets = Vec::new();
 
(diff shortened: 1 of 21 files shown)


irvingoujAtDevolution:
Force pushed because rebased on master

utils/dotnet/Devolutions.Gateway.Utils/src/EnrollmentClaims.cs
```diff
@@ -21,33 +21,21 @@ public class EnrollmentClaims : IGatewayClaims
...
+    /// "Generate Enrollment String" dialog so the agent shows up under that
+    /// name in the registry.
     /// </summary>
     [JsonPropertyName("jet_agent_name")]
-    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
-    public string? JetAgentName { get; set; }
+    public string JetAgentName { get; set; }
 
-    public EnrollmentClaims(string jetGwUrl, string? jetAgentName = null)
+    public EnrollmentClaims(string jetGwUrl, string jetAgentName)
```
CBenoit, line 31: @copilot Update the PR to remove words such as "authoritative". It’s a prompt residual because the claim was originally framed as optional. We typically don’t state thing as being "authoritative" otherwise, every claims are authoritative by default. Also, we should not mention DVLS. DVLS is not the only consumer of this library, and this is an overfitted description. The generic term is "provisioner". It’s also talking in length about the agent’s CLI, but that’s not directly a token shape thing. Re-eveluate that globally.

devolutions-agent/src/main.rs
```diff
@@ -160,8 +157,6 @@ fn parse_up_command_args_with_reader<R: BufRead>(args: &[String], mut stdin_read
 
         match arg {
             "--gateway" => gateway_url = Some(parse_required_value(args, &mut index, "--gateway")?),
```
CBenoit, line 159: question: Isn’t that already part of the enrollment token?

devolutions-agent/src/main.rs
CBenoit, line 163: thought: I need to understand again why this is not the default behavior, and why a parameter is required in the CLI

devolutions-agent/src/main.rs
```diff
@@ -160,8 +157,6 @@ fn parse_up_command_args_with_reader<R: BufRead>(args: &[String], mut stdin_read
...
-            "--token" | "--enrollment-token" => enrollment_token = Some(parse_required_value(args, &mut index, arg)?),
-            "--name" | "--agent-name" => agent_name = Some(parse_required_value(args, &mut index, arg)?),
             "--enrollment-string" => enrollment_string = Some(parse_required_value(args, &mut index, arg)?),
```
CBenoit, line 160: note: We used to have a `--token`, `--enrollment-token` and `--enrollment-string`. I’m not sure why we had the distinction originally, but now it seems that `--token` would be the best, simplest argument, and we don’t need aliases.

CBenoit (approved):

PR #1815, opened 2026-06-03
fix(dgw,agent): backwards-compatibility for update manifest V1

vnikonov-devolutions:
Fixes issue with invalid parsing/serialziation of UpdateManifest V1 (update.json).

 [Initial PR](https://github.com/Devolutions/devolutions-gateway/pull/1726) introduced bug where we expected "Version" field for product instead of "TargetVersion". This PR fixes this invalid behavior. (Also, "Version" is still will be accepted on agent side as field alias, so gateway `v2026.2.1` will work correctly with newer agent


CBenoit (approved):
Thank you!

PR #1816, opened 2026-06-04
fix(dgw): TRP live-shadow sizing + large-line panic

irvingoujAtDevolution:
Two TRP-only shadow-streaming fixes (cast path unaffected):

- **Sizing:** emit a standard asciicast `{width, height}` header with the correct axis. The decoder read columns from the wrong half of the size-change payload and emitted non-standard `row`/`col`, so clients sized the terminal transposed.
- **Panic:** `AsyncReadChannel::poll_read` copied a whole decoded line into the caller's `ReadBuf` unchecked, panicking when a line exceeded it (e.g. an htop full-screen redraw > 8 KiB). Now copies what fits and buffers the rest across reads.

### DIFF
--- a/crates/terminal-streamer/src/trp_decoder.rs
+++ b/crates/terminal-streamer/src/trp_decoder.rs
@@ -146,12 +146,10 @@ async fn parse_trp_stream(
                 }
             }
             2 => {
-                // Terminal size change
-                if before_setup_cache.is_some() {
-                    // Size-change payload is little-endian [columns, rows].
-                    header.col = u16::from_le_bytes(event_payload[0..2].try_into()?);
-                    header.row = u16::from_le_bytes(event_payload[2..4].try_into()?);
-                } else {
+                // Terminal size change. Payload is little-endian [columns, rows].
+                header.col = u16::from_le_bytes(event_payload[0..2].try_into()?);
+                header.row = u16::from_le_bytes(event_payload[2..4].try_into()?);
+                if before_setup_cache.is_none() {
                     let event = AsciinemaEvent::Resize {
                         width: header.col,
                         height: header.row,
--- a/crates/terminal-streamer/src/trp_decoder.rs
+++ b/crates/terminal-streamer/src/trp_decoder.rs
@@ -146,12 +146,12 @@ async fn parse_trp_stream(
                 }
             }
             2 => {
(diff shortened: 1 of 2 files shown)


crates/terminal-streamer/src/trp_decoder.rs
```diff
@@ -118,11 +146,13 @@ async fn parse_trp_stream(
...
-                if before_setup_cache.is_some() {
-                    header.row = u16::from_le_bytes(event_payload[0..2].try_into()?);
-                    header.col = u16::from_le_bytes(event_payload[2..4].try_into()?);
-                } else {
+// Terminal size change. Payload is little-endian [columns, rows].
+if event_payload.len() < 4 {
+    anyhow::bail!("invalid terminal size change payload length (len={})", event_payload.len());
+}
+header.col = u16::from_le_bytes(event_payload[0..2].try_into()?);
+header.row = u16::from_le_bytes(event_payload[2..4].try_into()?);
```
CBenoit:
```suggestion
                // Terminal size change. Payload is little-endian [columns, rows].
                if event_payload.len() < 4 {
                    anyhow::bail!("invalid terminal size change payload length (len={})", event_payload.len());
                }
                header.col = u16::from_le_bytes(event_payload[0..2].try_into()?);
                header.row = u16::from_le_bytes(event_payload[2..4].try_into()?);
```

CBenoit (approved):
LGTM!

crates/terminal-streamer/src/trp_decoder.rs
```diff
@@ -118,11 +146,13 @@ async fn parse_trp_stream(
...
             }
             2 => {
-                // Terminal size change
-                if before_setup_cache.is_some() {
-                    header.row = u16::from_le_bytes(event_payload[0..2].try_into()?);
-                    header.col = u16::from_le_bytes(event_payload[2..4].try_into()?);
-                } else {
+                // Terminal size change. Payload is little-endian [columns, rows].
+                if event_payload.len() < 4 {
+                    anyhow::bail!("invalid terminal size change payload length (len={})", event_payload.len());
```
CBenoit:
```suggestion
                    anyhow::bail!(
                        "invalid terminal size change payload length (len={})",
                        event_payload.len()
                    );
```

PR #1820, opened 2026-06-08
fix(webapp): pin asciinema-player to 3.8.2

irvingoujAtDevolution:
asciinema-player **3.9.0** reworked the WebSocket driver to negotiate a subprotocol — it connects with `new WebSocket(url, ["v1.alis", "v2.asciicast", ...])` and selects its stream parser from the one the server echoes back ([v3.9.0 release notes](https://github.com/asciinema/asciinema-player/releases/tag/v3.9.0): *"refactored websocket driver … uses WS subprotocol negotiation"*).The gateway terminal-stream WebSocket negotiates **no** subprotocol, so on 3.9+ the player falls back to the `raw` handler, never parses the asciicast live stream, and shows "Stream ended". Our `^3.8.1` range had floated to 3.13.5.Pin to **3.8.2** (last release before 3.9.0) to restore live terminal playback.

Issue: DGW-394

### DIFF
--- a/webapp/apps/recording-player/package.json
+++ b/webapp/apps/recording-player/package.json
@@ -16,7 +16,7 @@
   "dependencies": {
     "@devolutions/multi-video-player": "workspace:*",
     "@devolutions/shadow-player": "workspace:*",
-    "asciinema-player": "^3.8.1",
+    "asciinema-player": "3.8.2",
     "i18next": "^24.2.1",
     "vite-plugin-static-copy": "^2.3.0"
   }
--- a/webapp/pnpm-lock.yaml
+++ b/webapp/pnpm-lock.yaml
@@ -177,8 +177,8 @@ importers:
         specifier: workspace:*
         version: link:../../packages/shadow-player
       asciinema-player:
-        specifier: ^3.8.1
-        version: 3.13.5
+        specifier: 3.8.2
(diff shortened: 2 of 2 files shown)


CBenoit (approved):
LGTM

CBenoit:
> The gateway terminal-stream WebSocket negotiates **no** subprotocol, so on 3.9+ the player falls back to the `raw` handler, never parses the asciicast live stream, and shows "Stream ended". Our `^3.8.1` range had floated to 3.13.5.

Should we also update the terminal-stream WebSocket to negotiate the subprotocol? So ultimately we can remove the pin.

CBenoit:
Also, if you have a Jira ticket, please link it! Thank you 🙂

irvingoujAtDevolution:
> > The gateway terminal-stream WebSocket negotiates **no** subprotocol, so on 3.9+ the player falls back to the `raw` handler, never parses the asciicast live stream, and shows "Stream ended". Our `^3.8.1` range had floated to 3.13.5.
> 
> Should we also update the terminal-stream WebSocket to negotiate the subprotocol? So ultimately we can remove the pin.

Maybe, there's a bit more to consider because we are looking for adding Avalonia based terminal streamer. for now we'll just pin it

PR #1822, opened 2026-06-10
fix(dgw): improve session-level tracing

irvingoujAtDevolution:
## What

Three small, behavior-preserving tracing/logging improvements that make Gateway session failures easier to triage from logs alone.

### 1. Session-id correlation for recording uploads
- The `/jet/jrec/push` span now carries `session_id`, and a `Recording push stream started` event is emitted on connect.
- Recording-upload log lines now correlate with the owning `/jet/rdp` session through a single `session_id` grep (the push span previously only had `client=<ip>`).

### 2. Distinguish a Gateway-initiated kill from a peer-driven end
- `Proxy::forward` now logs `Forwarding ended because the session was killed by the Gateway` when the session is terminated via the kill path (recording-policy violation, max duration, ...), instead of the same `Forwarding ended` line used for a natural close.
- Removes a recurring triage ambiguity where a policy kill was indistinguishable from a target-side reset.

### 3. Log the resolved target IP on RDCleanPath TLS handshake failure
- When the TLS handshake to the target RDP server fails, log the resolved `server_addr` alongside the target hostname and error.
- Lets us tell a wrong / split-horizon DNS resolution apart from a target-side reset without requiring a packet capture.

## Why
These came out of recent support swarms around browser / web-client RDP: recording-required sessions getting killed when the recording stream drops, and target-side TLS resets where the resolved IP was the missing fact. In several of these the existing logs were correct but required expert correlation across multiple lines, or didn't surface the one fact needed.

## Scope / risk
Tracing and logging only - no control-flow or behavioral change. `cargo check` and `cargo clippy -p devolutions-gateway` are clean.

## Out of scope (possible follow-ups)
- Heartbeat / health observability (gateway-side record of inbound heartbeats + a "last heartbeat received" surface).
- Guaranteed startup-failure diagnostics written to the Windows Event Log.

### DIFF
--- a/devolutions-gateway/src/api/jmux.rs
+++ b/devolutions-gateway/src/api/jmux.rs
@@ -64,8 +64,9 @@ async fn handle_socket(
         keep_alive_interval,
     );
 
+    let session_id = claims.jet_aid;
     let result = crate::jmux::handle(stream, claims, sessions, subscriber_tx, traffic_audit_handle)
-        .instrument(info_span!("jmux", client = %source_addr))
+        .instrument(info_span!("jmux", client = %source_addr, session_id = %session_id))
         .await;
 
     if let Err(error) = result {
--- a/devolutions-gateway/src/api/jmux.rs
+++ b/devolutions-gateway/src/api/jmux.rs
@@ -66,7 +66,7 @@ async fn handle_socket(
 
     let session_id = claims.jet_aid;
     let result = crate::jmux::handle(stream, claims, sessions, subscriber_tx, traffic_audit_handle)
-        .instrument(info_span!("jmux", client = %source_addr, session_id = %session_id))
+        .instrument(info_span!("jmux", client = %source_addr, %session_id))
         .await;
 
     if let Err(error) = result {
--- a/devolutions-gateway/src/api/jrec.rs
+++ b/devolutions-gateway/src/api/jrec.rs
@@ -148,7 +148,7 @@ async fn handle_jrec_push(
         .shutdown_signal(shutdown_signal)
         .build()
         .run()
-        .instrument(info_span!("jrec", client = %source_addr))
+        .instrument(info_span!("jrec", client = %source_addr, session_id = %session_id))
         .await;
 
     match result {
--- a/devolutions-gateway/src/api/jrec.rs
+++ b/devolutions-gateway/src/api/jrec.rs
@@ -148,7 +148,7 @@ async fn handle_jrec_push(
         .shutdown_signal(shutdown_signal)
         .build()
         .run()
-        .instrument(info_span!("jrec", client = %source_addr, session_id = %session_id))
+        .instrument(info_span!("jrec", client = %source_addr, %session_id))
         .await;
 
     match result {
--- a/devolutions-gateway/src/proxy.rs
+++ b/devolutions-gateway/src/proxy.rs
@@ -145,7 +145,14 @@ where
 
         match res {
             Ok(()) => {
-                info!(?outcome, "Forwarding ended");
+                match outcome {
+                    ForwardingOutcome::KilledByGateway => {
+                        info!(?outcome, "Forwarding ended: session was killed by the Gateway");
+                    }
+                    ForwardingOutcome::Completed => {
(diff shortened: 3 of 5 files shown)


devolutions-gateway/src/api/jmux.rs
```diff
@@ -64,8 +64,9 @@ async fn handle_socket(
         keep_alive_interval,
     );
 
+    let session_id = claims.jet_aid;
     let result = crate::jmux::handle(stream, claims, sessions, subscriber_tx, traffic_audit_handle)
-        .instrument(info_span!("jmux", client = %source_addr))
+        .instrument(info_span!("jmux", client = %source_addr, session_id = %session_id))
```
CBenoit:
```suggestion
        .instrument(info_span!("jmux", client = %source_addr, %session_id))
```

devolutions-gateway/src/api/jrec.rs
```diff
@@ -148,7 +148,7 @@ async fn handle_jrec_push(
         .shutdown_signal(shutdown_signal)
         .build()
         .run()
-        .instrument(info_span!("jrec", client = %source_addr))
+        .instrument(info_span!("jrec", client = %source_addr, session_id = %session_id))
```
CBenoit:
```suggestion
        .instrument(info_span!("jrec", client = %source_addr, %session_id))
```

devolutions-gateway/src/proxy.rs
```diff
@@ -131,7 +145,14 @@ where
...
             Ok(()) => {
-                info!("Forwarding ended");
+                match outcome {
+                    ForwardingOutcome::KilledByGateway => {
+                        info!(?outcome, "Forwarding ended: session was killed by the Gateway");
+                    }
+                    ForwardingOutcome::Completed => {
+                        info!(?outcome, "Forwarding ended");
+                    }
+                }
```
CBenoit: suggestion: This is enough:

```suggestion
                info!(?outcome, "Forwarding ended");
```

devolutions-gateway/src/recording.rs
```diff
@@ -99,6 +99,8 @@ where
             anyhow::bail!("inconsistent session ID (ID in token: {})", claims.jet_aid);
         }
 
+        info!(%session_id, ?file_type, "Recording push stream started");
+
```
CBenoit: question: Isn’t that already logged as part of the URL?

devolutions-gateway/src/rd_clean_path.rs
CBenoit, line 1: question: Do we really need the complex classification introduced in the diff of this file?

I doubt it because errors are already supposed to be logged as `error!` (and thus we don’t need the extra `warn!` I think?), and if not, we should find the place were the error is bubbled and ensure we properly handle the error instead of adding special handling everywhere.

The exact error that is being classified is also already part of the underlying error, and will be logged if we use a proper error reporter (in Devolutions Gateway, that tends to be `anyhow::Error` currently).

devolutions-gateway/src/proxy.rs
```diff
@@ -30,6 +30,20 @@ pub struct Proxy<A, B> {
...
+/// closing log line is not mistaken for a target-side reset during triage.
+#[derive(Debug)]
+enum ForwardingOutcome {
+    /// The bidirectional forwarding finished on its own (a peer closed the connection or
+    /// reached EOF).
+    Completed,
+    /// The Gateway terminated the session through the kill signal, e.g. because the
+    /// recording policy was violated or the max session duration was reached.
+    KilledByGateway,
+}
```
CBenoit: question: Again, I’m not sure we really need to make the distinction here. When the session is killed, there is a log showing that the /terminate endpoint was poked; or a log indicating that the lifetime of the session was reached (which is more specific that just "killed by Gateway". Am I missing something?

CBenoit (commented):
Thank you!

devolutions-gateway/src/rd_clean_path.rs
irvingoujAtDevolution: updated

CBenoit (approved):
LGTM

PR #1824, opened 2026-06-12
feat(web-recorder): add @devolutions/web-recorder session capture library

irvingoujAtDevolution:
Solves two problems:

1. **Unified recorder** — DVLS web and Hub web each maintain their own copy of the recorder (~500 lines, ~92% identical and starting to diverge). This package is the single source of truth so fixes land once for both.

2. **Canvas idle regression in newer Chromium** — the old `setInterval` + `globalAlpha=0` keepalive no longer marks the canvas dirty in Chromium 149+, producing black/empty recordings on static content. This fixes it with a `requestAnimationFrame` keepalive that alternates a real pixel value so the compositor always has a fresh frame.

Adds `WebMRecorder` (canvas → WebM over WS) and `AsciiCastV2Recorder` (terminal → asciicast v2 over WS). Promise-based API, no framework dependency. Already in the `publish-libraries.yml` npm matrix.


CBenoit (approved):
Good initiative 💯 

> Promise-based API, no framework dependency. Already in the `publish-libraries.yml` npm matrix.

I like this approach!

kristahouse (approved):
This is fantastic! 🔥

PR #1835, opened 2026-06-23
fix(agent,agent-installer): fail the install if the agent tunnel can't reach the gateway

irvingoujAtDevolution:
Split out of #1831.

## Problem: the installer reports success on *enrollment*, not on tunnel *connectivity*

**Symptom:** the MSI install shows the Agent Tunnel step as **success**, but the agent never appears online in the Gateway / DVLS agent list.

**Root cause:** enrollment and the tunnel use two different network paths. `EnrollAgentTunnel` runs `devolutions-agent up` → `enroll_agent()`, whose success criteria were only:

1. `POST https://<gw>:7171/jet/tunnel/enroll` (HTTPS **management** port, TCP) returns 2xx and issues the client cert, and
2. the cert/key/CA + the `Tunnel` section are persisted to `agent.json`.

The actual data path — the **QUIC tunnel over UDP (4433)** — is established later by the agent service and was **never probed at install time**. Enrollment is TCP/7171; the tunnel is UDP/4433. So when UDP 4433 is blocked (firewall/NAT), the install is green while the agent silently fails to connect (`Tunnel connection lost error=QUIC handshake: timed out`).

## Fix

After enrolling, `agent up` now performs a **one-shot QUIC + mTLS connectivity probe** to the gateway tunnel endpoint and **exits non-zero on failure**. `EnrollAgentTunnel` already checks `up`'s exit code, so a blocked UDP path now **fails the install and rolls back** — giving the operator actionable feedback (verify UDP 4433 / firewall) while they're still at the machine.

- **agent** (`tunnel.rs`, `main.rs`): `probe_connectivity` reuses the live connect path (the same `connect_to_gateway` the running service uses) for a single mTLS + QUIC handshake, bounded by a timeout, then drains the connection (`close` + bounded `wait_idle`) so the gateway unregisters the probe promptly. A completed handshake is sufficient proof the UDP path is open, so the standalone `probe-tunnel` subcommand and the heavier heartbeat round-trip probe were removed.
- **agent-installer** (`CustomActions.cs`): removed the in-CA subprocess probe (it lives in `up` now). On a failed `up`, both the timeout and non-zero-exit paths roll back a freshly-persisted enrollment via a guarded helper that cleans up **only when `up` actually wrote new certs** (uuid-named client cert path changed from the pre-`up` snapshot) — never the prior install's certs, and never when the pre-`up` snapshot couldn't be captured. Fails safe.

## Testing

Validated end-to-end on a lab agent VM against the live gateway:

- **UDP 4433 open** → probe succeeds, install completes.
- **UDP 4433 blocked** (firewall rule) → `up` exits non-zero → **install fails and rolls back**, no orphaned artifacts left behind.

Unit tests: `probe_fails_fast_when_tunnel_disabled`, `probe_times_out_when_gateway_unreachable`.

Reviewed via an iterative Claude + Codex review loop (converged clean).

## Known follow-ups (out of scope here)

- **Kill-mid-enroll window:** if `up` is hard-killed (the installer's 60s timeout) *after* writing the cert files but *before* `agent.json` is updated, the new files can orphan / a fixed-name `gateway-ca.pem` may be left overwritten. Pre-existing; the clean fix is a transactional `up` or an MSI cert-directory snapshot.
- **Gateway-side registry** unregisters by `agent_id` without a connection-identity check; the probe's bounded drain compensates on this path (it runs before the service starts), but adding an identity/generation check gateway-side would close the race for good.

### DIFF
--- a/devolutions-agent/src/main.rs
+++ b/devolutions-agent/src/main.rs
@@ -38,6 +38,7 @@ mod service;
 use std::env;
 use std::io::{self, BufRead};
 use std::sync::mpsc;
+use std::time::Duration;
 
 use anyhow::{Context as _, Result, bail};
 use ceviche::Service;
@@ -277,7 +278,12 @@ fn main() {
                         &command.enrollment_token,
                         command.advertise_subnets,
                     )
-                    .await
+                    .await?;
+
+                    // Enrollment only proves HTTPS/TCP; fail the install now if the QUIC/UDP tunnel
+                    // path is blocked, while the operator is still here to fix the firewall.
+                    let conf = ConfHandle::init().context("load agent configuration for connectivity probe")?;
+                    devolutions_agent::tunnel::probe_connectivity(&conf.get_conf().tunnel, Duration::from_secs(15)).await
                 });
 
                 if let Err(error) = result {
--- a/devolutions-agent/src/tunnel.rs
+++ b/devolutions-agent/src/tunnel.rs
@@ -138,8 +138,7 @@ impl Task for TunnelTask {
                     return Ok(());
                 }
                 Ok(ConnectionOutcome::CertRenewed) => {
-                    // Renewal is a successful "completion", not a failure — skip
-                    // the backoff and reconnect immediately with the new cert.
+                    // Renewal is a completion, not a failure.
                     info!("Certificate renewed; reconnecting with new cert immediately");
                     backoff.reset();
                     continue;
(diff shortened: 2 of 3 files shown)


irvingoujAtDevolution:
Closing in favor of #1837 — recreated on a fresh branch (`fix/agent-tunnel-connectivity-probe`) as a single clean commit, with no force-push in its history.

PR #1837, opened 2026-06-25
fix(agent,agent-installer): fail the install if the agent tunnel can't reach the gateway

irvingoujAtDevolution:
Enrollment only proves the HTTPS/TCP path, but the tunnel runs over QUIC/UDP (4433). When UDP is blocked, enrollment still succeeds, so the installer reported success while the agent never connected.

**Fix:** after enrolling, `agent up` performs a one-shot QUIC + mTLS handshake to the gateway and exits non-zero on failure — which `EnrollAgentTunnel` already turns into a failed, rolled-back install. The standalone `probe-tunnel` subcommand and the heartbeat round-trip are gone. The installer-side cleanup rolls back only a freshly-persisted enrollment, never a prior install's certs (guarded, fails safe).

Verified on a lab agent VM: UDP 4433 open → install succeeds; blocked → install fails and rolls back, no orphaned artifacts.

Out-of-scope follow-ups: the kill-mid-enroll window (pre-existing) and a gateway-side connection-identity check on `unregister`.

### DIFF
--- a/devolutions-agent/src/tunnel.rs
+++ b/devolutions-agent/src/tunnel.rs
@@ -451,8 +451,8 @@ async fn connect_to_gateway(
     Ok((endpoint, connection))
 }
 
-/// Confirm the QUIC/UDP path to the gateway is open by completing one mTLS+QUIC handshake, then
-/// draining the connection, bounded by `timeout`.
+/// Confirm the QUIC/UDP path to the gateway is open by completing one mTLS+QUIC handshake within
+/// `timeout`, then draining the connection (a best-effort teardown that adds up to ~3s).
 pub async fn probe_connectivity(tunnel_conf: &crate::config::TunnelConf, timeout: Duration) -> anyhow::Result<()> {
     if !tunnel_conf.enabled {
         bail!("agent tunnel is not enabled");
@@ -707,7 +707,7 @@ mod tests {
     #[tokio::test]
     async fn probe_times_out_when_gateway_unreachable() {
         // Throwaway PEMs so the pre-connect file reads succeed; nothing listens on the target
-        // port, so the handshake never completes and the probe must hit its own timeout.
+        // port, so the probe fails quickly — via its own timeout or an immediate connect error.
         let cert_key =
             rcgen::generate_simple_self_signed(vec!["localhost".to_owned()]).expect("generate self-signed cert");
         let dir = tempfile::tempdir().expect("temp dir");
--- a/devolutions-agent/src/tunnel.rs
+++ b/devolutions-agent/src/tunnel.rs
@@ -189,10 +189,22 @@ enum ConnectionOutcome {
     CertRenewed,
 }
 
-/// Build the route advertisement payload from the current tunnel configuration.
-fn route_advertisements(
-    tunnel_conf: &crate::config::TunnelConf,
-) -> anyhow::Result<(Vec<Ipv4Network>, Vec<agent_tunnel_proto::DomainAdvertisement>)> {
+/// Run a single QUIC tunnel connection lifetime: config → connect → event loop.
+///
+/// - `Ok(Shutdown)`: graceful shutdown, exit the task.
+/// - `Ok(CertRenewed)`: certificate renewed; caller should reconnect immediately.
+/// - `Err(_)`: connection lost or handshake failed — caller should retry with backoff.
+async fn run_single_connection(
+    conf_handle: &ConfHandle,
+    shutdown_signal: &mut ShutdownSignal,
+) -> anyhow::Result<ConnectionOutcome> {
+    let agent_conf = conf_handle.get_conf();
+    let tunnel_conf = &agent_conf.tunnel;
+
+    let cert_path = &tunnel_conf.client_cert_path;
+    let key_path = &tunnel_conf.client_key_path;
+    let ca_path = &tunnel_conf.gateway_ca_cert_path;
+
     let advertise_subnets: Vec<Ipv4Network> = tunnel_conf
         .advertise_subnets
         .iter()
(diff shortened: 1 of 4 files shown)


devolutions-agent/src/tunnel.rs
```diff
@@ -638,3 +665,70 @@ async fn run_session_proxy(advertise_subnets: Vec<Ipv4Network>, send: quinn::Sen
...
+            "unexpected error: {error:#}"
+        );
+    }
+
+    #[tokio::test]
+    async fn probe_times_out_when_gateway_unreachable() {
+        // Throwaway PEMs so the pre-connect file reads succeed; nothing listens on the target
+        // port, so the probe fails quickly — via its own timeout or an immediate connect error.
+        let cert_key =
+            rcgen::generate_simple_self_signed(vec!["localhost".to_owned()]).expect("generate self-signed cert");
```
CBenoit: issue: Use a pre-generated certificate / key pair instead of generating a new one on each run. It slows down the testsuite.

devolutions-agent/src/tunnel.rs
```diff
@@ -638,3 +665,70 @@ async fn run_session_proxy(advertise_subnets: Vec<Ipv4Network>, send: quinn::Sen
...
+        conf.gateway_endpoint = "127.0.0.1:1".to_owned();
+        conf.client_cert_path = Utf8PathBuf::from_path_buf(cert_path).expect("utf8 cert path");
+        conf.client_key_path = Utf8PathBuf::from_path_buf(key_path).expect("utf8 key path");
+        conf.gateway_ca_cert_path = Utf8PathBuf::from_path_buf(ca_path).expect("utf8 ca path");
+
+        let started = std::time::Instant::now();
+        let result = probe_connectivity(&conf, Duration::from_secs(2)).await;
+
+        assert!(result.is_err(), "probe must fail when the gateway is unreachable");
+        assert!(started.elapsed() < Duration::from_secs(15), "probe must fail fast");
```
CBenoit: 15 seconds is overly generous. For a unit test it should not exceed 1 second at most.

devolutions-agent/src/tunnel.rs
```diff
@@ -638,3 +665,70 @@ async fn run_session_proxy(advertise_subnets: Vec<Ipv4Network>, send: quinn::Sen
...
+            server_spki_sha256: None,
+        }
+    }
+
+    #[tokio::test]
+    async fn probe_fails_fast_when_tunnel_disabled() {
+        let mut conf = tunnel_conf_template();
+        conf.enabled = false;
+
+        let error = probe_connectivity(&conf, Duration::from_secs(5))
```
CBenoit: Same, a unit test should not run more than 1 sec.

package/AgentWindowsManaged/Actions/CustomActions.cs
```diff
@@ -540,6 +525,11 @@ ActionResult Fail(string msg)
                 if (process.ExitCode != 0)
                 {
                     string detail = !string.IsNullOrWhiteSpace(stderr) ? Redact(stderr).Trim() : $"exit code {process.ExitCode}";
+
+                    // `up` enrolls then probes, so a non-zero exit can leave a freshly-persisted
+                    // enrollment on disk with no marker yet; undo it (guarded against early failures).
+                    RollbackFailedEnrollment(session, agentJsonPath, originalTunnel, originalGatewayCaB64, originalStateCaptured);
```
CBenoit: suggestion: Add a separate command specifically for probing. The probing should be used as a diagnostic, and the enrollment action doing "enrollment + diagnostic probe" is probably not granular enough.

irvingoujAtDevolution: Updated — it runs probe independently now. `up` only enrolls, `probe` is its own command (usable standalone as a diagnostic), and the installer runs it as a separate step. c88ab705

devolutions-agent/src/tunnel.rs
```diff
@@ -138,8 +138,7 @@ impl Task for TunnelTask {
                     return Ok(());
                 }
                 Ok(ConnectionOutcome::CertRenewed) => {
-                    // Renewal is a successful "completion", not a failure — skip
-                    // the backoff and reconnect immediately with the new cert.
+                    // Renewal is a completion, not a failure.
```
CBenoit:
```suggestion
                    // Renewal is a successful outcome, not a failure, we release the backpressure (backoff.reset()).
```

devolutions-agent/src/tunnel.rs
```diff
@@ -260,23 +256,112 @@ async fn run_single_connection(
...
+    Ok(ConnectionOutcome::Shutdown)
+}
+
+/// Build the mTLS client config, resolve the gateway endpoint, and perform the
+/// QUIC handshake, returning the live endpoint and connection.
+async fn connect_to_gateway(
+    tunnel_conf: &crate::config::TunnelConf,
+) -> anyhow::Result<(quinn::Endpoint, quinn::Connection)> {
+    // Ensure rustls crypto provider is installed (ring).
+    let _ = rustls::crypto::ring::default_provider().install_default();
```
CBenoit, line 347:
```suggestion
    let _ = rustls::crypto::ring::default_provider().install_default();

```

devolutions-agent/src/tunnel.rs
```diff
@@ -363,84 +448,88 @@ async fn run_single_connection(
...
+}
 
-    info!(epoch, "Sent initial RouteAdvertise");
+/// Reachability probe: confirm the gateway's QUIC/UDP endpoint answers within `timeout`.
+///
+/// This deliberately does NOT complete a QUIC/mTLS handshake. A real handshake would register
+/// this agent's connection on the gateway (keyed by the enrolled `agent_id`) and, on close,
+/// unregister it — so a probe run against a machine whose service tunnel is live would evict the
+/// real connection. Instead we send a QUIC long-header packet carrying an unsupported version and
+/// wait for the gateway to reply with a Version Negotiation packet. That reply proves UDP/4433
```
CBenoit, line 461: thought: For a follow up, maybe we want an official "probe version", otherwise I fear we’ll be seeing "invalid versions" errors in the Gateway logs.

CBenoit (approved):

CBenoit:
I think the probe here is hard blocking the user from moving forward, is that accurate? My preference would be for opening a popup asking the user what to do: "Do you want to continue anyway?"

Could you follow up on that too, please?

PR #1838, opened 2026-06-26
fix(dgw): write boot.stacktrace file when running as a service

CBenoit:
When Devolutions Gateway failed to start as a Windows service — for example because of an invalid TLS certificate or PFX file loaded before logging is initialized — no diagnostic was produced: the failure surfaced only as an SCM "service started then stopped" with no log line and no boot.stacktrace file. The boot.stacktrace file is now also written on service-mode startup failures, mirroring the behavior already present in CLI/foreground mode, so the underlying error is captured even when the logging subsystem never came up.

Issue: DGW-401

---
_Generated by [Claude Code](https://claude.ai/code/session_01LEseg8XXfDxiaHWkQCJteW)_


CBenoit:
I reviewed Claude’s output, and it looks good to me.

philippelatulippe (approved):

PR #1843, opened 2026-06-30
feat(agent): harden PSU gRPC POC support

mamoreau-devolutions:
Hardens the PSU gRPC agent POC and adds optional AppToken support for authenticating the agent to PowerShell Universal. The POC container also avoids running as root and no longer disables TLS verification during Docker builds.

This keeps unauthenticated local development possible when no AppToken is configured, while allowing PSU deployments to pass an application token through agent configuration or the container helper.

Issue: N/A

### DIFF
--- a/devolutions-agent/Dockerfile.psu-grpc-poc
+++ b/devolutions-agent/Dockerfile.psu-grpc-poc
@@ -2,8 +2,7 @@ FROM rust:1.90-bookworm AS build
 
 WORKDIR /workspace
 
-ENV CARGO_NET_GIT_FETCH_WITH_CLI=true \
-    GIT_SSL_NO_VERIFY=true
+ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends build-essential clang cmake git libclang-dev pkg-config \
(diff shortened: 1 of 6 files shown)


mamoreau-devolutions:
Implementation notes:
- Adds optional `PsuGrpcAgent.AppToken` config and sends it as `authorization: Bearer <token>` gRPC metadata only when configured.
- Adds `PSU_APP_TOKEN` and `-AppToken` wiring for the POC container scripts.
- Reworks process stop handling so graceful stop closes stdin while preserving later kill control, and cleans registry state after process failures.
- Reports spawn failures with `StreamClosed` and `ProcessCompleted`.
- Drains stdout/stderr pump tasks before completion to avoid losing final output.
- Removes `GIT_SSL_NO_VERIFY=true` from the POC Docker build and runs the runtime container as a non-root user.
- Validated with `cargo check -p devolutions-agent`, `cargo test -p devolutions-agent --quiet`, `cargo clippy -p devolutions-agent --tests -- -D warnings`, Rust fmt, and script parser checks.

Remaining follow-up work:
- Enforce TLS/HTTPS for AppToken-bearing production connections.
- Add command/runtime allowlisting and process-tree cancellation.
- Tighten protocol semantics around stream kinds, structured status, heartbeat, and cancel response.

PR #1844, opened 2026-06-30
refactor(agent): reuse NOW execution for PSU gRPC

mamoreau-devolutions:
Reuses the existing Devolutions Session NOW_EXEC process backend for PSU gRPC remote execution on Windows while keeping the gRPC protocol unchanged. This aligns process launch, stdin/stdout handling, cancellation, and cleanup with the Agent's existing NOW behavior instead of maintaining a separate Windows execution path.

Keeps the non-Windows fallback on tokio process execution and hardens duplicate process/stream validation and failure cleanup around the gRPC adapter.

Issue: N/A


mamoreau-devolutions:
Implementation notes:
- Windows PSU gRPC StartProcess now adapts into the existing NOW/DVC WinApiProcessBuilder path from devolutions-session, using raw data encoding because the current gRPC protocol has no NOW encoding flags.
- StopProcess maps kill_process=true to NOW abort and kill_process=false to NOW cancel; stdin StreamData is forwarded through the DVC process input channel.
- The gRPC protocol still has no stdout/stderr stream kind, so stdout remains StreamData and stderr is surfaced as diagnostics instead of being silently dropped.
- Added duplicate process and stream ID handling before spawning process work, plus explicit working-directory validation and registry cleanup on spawn failure.
- Added opt-in kill-on-drop support to the shared WinApiProcess wrapper and enabled it only for the gRPC adapter so dropped gRPC process tasks do not orphan child processes; existing NOW/DVC sessions keep the previous default.
- Validation: cargo +nightly fmt --all; cargo check -p devolutions-agent -p devolutions-session --features devolutions-session/dvc; cargo test -p devolutions-agent psu_grpc_agent -- --nocapture; cargo clippy --workspace --tests -- -D warnings; cargo test --workspace --exclude devolutions-agent-updater -- --test-threads=1. Full cargo test --workspace requires elevation for the updater test binary in this environment.

PR #1848, opened 2026-07-08
feat(agent): support PSU gRPC app token secrets

adamdriscoll:
PSU gRPC agents need to authenticate using configured application tokens while keeping secret-backed tokens out of static configuration. This change resolves the configured gRPC `AppToken` before opening the agent stream so literal tokens continue to pass through and `$secret:<name>` references reuse the existing PSU PowerShell secret resolver.

The implementation keeps token handling in the agent transport layer and avoids server-side or per-method parsing in this repository because the ASP.NET Core gRPC server and `PSUPermissions` integration are not present here. The gRPC request still carries the resolved token through standard `authorization` metadata for the PSU server authentication middleware to consume.

Validation:
- `cargo +nightly fmt --all`
- `cargo test -p devolutions-agent psu_grpc_agent && cargo test -p devolutions-agent psu_event_hub::powershell_worker && cargo test -p devolutions-agent config`
- `cargo clippy --workspace --tests -- -D warnings`
- `git --no-pager diff --check`
- `cargo test -p devolutions-agent`

`cargo test --workspace` was attempted. The parallel run hit Windows paging-file/mmap errors; the single-job retry with `ProgramData=C:\ProgramData` progressed until `devolutions-agent-updater` required elevation (`os error 740`).


devolutions-agent/src/psu_grpc_agent/mod.rs
```diff
@@ -140,6 +141,7 @@ impl PsuGrpcAgent {
     }
 
     async fn run_single_connection(&self, shutdown_signal: &mut ShutdownSignal) -> anyhow::Result<()> {
+        let app_token = self.resolve_app_token().await?;
         let endpoint = Endpoint::from_shared(self.server_url.clone())?;
```
CBenoit, line 152: issue: I think app-token resolution runs on every reconnect, not once.
`run_single_connection` is called inside the retry loop in `run()` (in `mod.rs`).
For a `$secret:` token that means a fresh `PowerShellWorker::new` (temp .ps1 write) plus a pwsh subprocess spawn each cycle. "resolve ... before opening the agent stream" reads as a single pre-connect resolution, so this per-attempt behavior looks accidental. Contrast the event-hub path, which builds one shared worker and resolves once.

Probably another issue with this: because resolution sits inside the retried `run_single_connection`, a permanently misconfigured secret (missing vault, typo'd name…) is indistinguishable from a network failure: warn, back off, retry forever, spawning pwsh each cycle. The event-hub path instead logs and *skips*
the connection on resolution failure. For a single-connection agent "skip" would mean exiting the task, so retry may be the right call if the vault could come online later, but this deserves an explicit decision.

suggestion: Fix is small: resolve once in `run()` before the loop and pass the resolved token into
`run_single_connection`. Moving resolution out of the loop also lets a hard config error fail loudly once rather than looping quietly. If per-reconnect re-resolution is instead a deliberate secret-rotation feature, please signal the intent with inline comments.

thought: Also, if we want to distinguish which error are transients and which are fatal, we may consider returning something else than anyhow::Result. It’s generally a good idea when dealing with retry loops.

devolutions-agent/src/psu_event_hub/mod.rs
```diff
@@ -1,7 +1,7 @@
 pub(crate) mod compat;
 mod executor;
 mod models;
-mod powershell_worker;
+pub(crate) mod powershell_worker;
```
CBenoit: thought: PowerShellWorker now has two consumers but still lives inside one of them. The gRPC transport now depends on the SignalR transport's internals for what is really a shared PSU concern (both spawn pwsh, both parse `$secret:`). Before this PR the two transports were independent siblings. This is the "make X reach into a sibling for a shared capability" shape that's usually better solved by extracting the shared piece: relocate powershell_worker (+ app_token_secret_reference_name) into a neutral module (e.g. psu_powershell, or
a psu parent both transports sit under). Note the worker also returns psu_event_hub::models::WebsocketEventResponse, so a clean extraction should carry/rename that type to something transport-neutral ("Websocket" is already a misnomer for a generic worker response).

Not blocking; happy to submit a follow up PR if you would rather not deal with that here.

devolutions-agent/src/psu_grpc_agent/mod.rs
```diff
@@ -259,6 +258,25 @@ impl PsuGrpcAgent {
...
     }
 
+    async fn resolve_app_token(&self) -> anyhow::Result<Option<String>> {
+        let Some(app_token) = self.app_token.as_deref() else {
+            return Ok(None);
+        };
+
+        if app_token_secret_reference_name(app_token).is_none() {
+            return Ok(Some(app_token.to_owned()));
+        }
```
CBenoit, line 276: nitpick: This is a redundant $secret: check. `PsuGrpcAgent::resolve_app_token` checks `app_token_secret_reference_name` and `PowerShellWorker::resolve_app_token` checks it again.
Harmless, and this does buy skipping worker construction for literal tokens. Fine to keep; just flagging the double parse in case it’s unintended.

suggestion: A small inline comment would be appreciated if it’s intended 🙂

devolutions-agent/src/psu_powershell.rs
```diff
@@ -563,9 +563,9 @@ mod tests {
 
     #[test]
     fn secret_reference_name_is_case_insensitive() {
```
CBenoit: nitpick: Stale test name after the rename.

CBenoit (commented):
Thank you Adam!

devolutions-agent/src/psu_grpc_agent/mod.rs
adamdriscoll: Addressed in https://github.com/Devolutions/devolutions-gateway/pull/1848/commits/341fd1d2dd59ac812172050ceed1c92a03f4828b

devolutions-agent/src/psu_event_hub/mod.rs
adamdriscoll: Addressed in https://github.com/Devolutions/devolutions-gateway/pull/1848/commits/ed2673d3401b0d173a4697f6a1d5015b5a8b4ea4

devolutions-agent/src/psu_grpc_agent/mod.rs
adamdriscoll: Comment added: https://github.com/Devolutions/devolutions-gateway/pull/1848/commits/6bdda93914fbe1da8d61049a29234290bd5d8876

devolutions-agent/src/psu_powershell.rs
adamdriscoll: Renamed here: https://github.com/Devolutions/devolutions-gateway/pull/1848/commits/061d38f71f58f173508e24e4aa695272fc78c373

CBenoit (approved):
Thank you, LGTM!

PR #1851, opened 2026-07-13
feat(agent-installer): configure the PSU agent from the installer

CBenoit:
Add a configuration screen to the Devolutions Agent installer, shown when the PowerShell Universal Agent feature is selected. Operators can enter the PSU server URL and application token — either verbatim or as a $secret:<name> reference resolved from SecretManagement at runtime — plus an optional agent id and display name. These are written to agent.json so the agent starts successfully on first run.

The installer performs a best-effort reachability probe against the server URL when the screen is confirmed, surfacing connectivity problems early instead of after installation. A failed probe warns but does not block installation.

The same values can be supplied for a silent install through public MSI properties; a silent major upgrade preserves an existing configuration when no PSU properties are passed. On rollback the previous agent.json PSU section is restored, and the token never lingers in temporary files or verbose MSI logs.

image

image

image

image

image


devolutions-agent/src/config.rs
```diff
@@ -737,13 +712,13 @@ pub mod dto {
...
-
-        /// PowerShell Universal gRPC remoting agent.
-        #[serde(skip_serializing_if = "Option::is_none")]
-        pub psu_grpc_agent: Option<PsuGrpcAgentConf>,
+        /// PowerShell Universal remoting agent.
+        // The `PsuGrpcAgent` alias is accepted on read so existing gRPC installations keep working
+        // on upgrade; new configs are always serialized under the `Psu` key.
+        // Existing installations here are mostly developer setups, so we’ll simply remove the alias once the feature is
+        // stabilized.
+        #[serde(alias = "PsuGrpcAgent", skip_serializing_if = "Option::is_none")]
```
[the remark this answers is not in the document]
CBenoit: We accept this; ignore

package/AgentWindowsManaged/Dialogs/PsuDialog.cs
```diff
@@ -0,0 +1,125 @@
...
+public partial class PsuDialog : AgentDialog
+{
+    public PsuDialog()
+    {
+        InitializeComponent();
+        label1.MakeTransparentOn(banner);
+        label2.MakeTransparentOn(banner);
+    }
+
+    public override bool ToProperties()
```
thenextman, line 22: Are we missing the corresponding `FromProperties` override? The purpose is that if you go "Forward" through the wizard (and every dialog is converting its state `ToProperties`) what happens when we go "Back" through the wizard? The `FromProperties` is to deserialize the state back into the dialog, so the user selections are still there

package/AgentWindowsManaged/Dialogs/PsuDialog.cs
```diff
@@ -0,0 +1,125 @@
...
+    {
+        Runtime.Session[AgentProperties.psuServerUrl.Id] = serverUrl.Text.Trim();
+        Runtime.Session[AgentProperties.psuAppToken.Id] = appToken.Text.Trim();
+        Runtime.Session[AgentProperties.psuAppTokenIsSecretReference.Id] = secretNameRadio.Checked ? "true" : "false";
```
thenextman: In general throughout I see we're manually coalescing `bool` from the underlying string, but the properties should be typed (at least, they are on Gateway - forgive me if we never did the same thing on Agent, but I think we did...)

So you can do e.g.

```
new AgentProperties(this.Runtime.Session)
        {
            PsuAppTokenIsSecretReference = this.secretNameRadio.Checked
        };
````

And the same in `FromProperties`, grab a reference to the typed properties and use it

```
AgentProperties properties = new(this.Runtime.Session);
this.secretNameRadio.Checked = properties.PsuAppTokenIsSecretReference;
````

File: package/AgentWindowsManaged/Dialogs/PsuDialog.cs [hunk 4]
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        string serverUrlText = serverUrl.Text.Trim();
+
+        // Validate the URL shape here so the agent doesn't fail to parse it as a url::Url on its
+        // first start after installation. The same check runs in the ConfigurePsuAgent custom action
+        // for the silent-install path.
+        if (!CustomActions.IsValidPsuServerUrl(serverUrlText))
+        {
+            ShowValidationErrorString("The server URL must be an absolute http or https URL, for example http://localhost:5000.");
```
thenextman: Validation errors  and other message box content should be localized

thenextman (approved):

PR #1852, opened 2026-07-13
fix(agent): recover DVC initialization after session readiness

mamoreau-devolutions:
Improves Devolutions Agent reliability for RDP connections that use multi-step authentication or reconnect an existing session. The session host now waits for the appropriate Windows logon or unlock event before starting, and transient virtual-channel initialization failures recover automatically instead of ending the session immediately.

Issue: N/A

### DIFF
--- a/devolutions-agent/src/session_manager.rs
+++ b/devolutions-agent/src/session_manager.rs
@@ -32,10 +32,36 @@ enum SessionKind {
     Remote,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum SessionReadiness {
+    WaitingForLogon,
+    WaitingForUnlock,
+    Ready,
+}
+
+impl SessionReadiness {
+    fn is_satisfied_by(self, event: SessionReadinessEvent) -> bool {
+        matches!(
+            (self, event),
+            (Self::WaitingForLogon, SessionReadinessEvent::Logon)
+                | (
+                    Self::WaitingForUnlock,
+                    SessionReadinessEvent::Logon | SessionReadinessEvent::Unlock
+                )
+        )
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+enum SessionReadinessEvent {
+    Logon,
+    Unlock,
+}
+
 struct GatewaySession {
     session: Session,
     kind: SessionKind,
-    is_session_ready: bool,
+    readiness: SessionReadiness,
 }
 
 // NOTE: `ceviche::controller::Session` do not implement `Debug` for session.
@@ -44,17 +70,17 @@ impl Debug for GatewaySession {
         f.debug_struct("GatewaySession")
             .field("session", &self.session.id)
             .field("kind", &self.kind)
-            .field("is_session_ready", &self.is_session_ready)
+            .field("readiness", &self.readiness)
             .finish()
     }
 }
 
 impl GatewaySession {
-    fn new(session: Session, kind: SessionKind) -> Self {
+    fn new(session: Session, kind: SessionKind, readiness: SessionReadiness) -> Self {
         Self {
             session,
             kind,
-            is_session_ready: false,
+            readiness,
         }
     }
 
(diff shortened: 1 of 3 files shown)


devolutions-agent/src/session_manager.rs
```diff
@@ -68,13 +94,16 @@ impl GatewaySession {
...
+        self.readiness.is_satisfied_by(event)
     }
 
-    fn set_session_ready(&mut self, is_ready: bool) {
-        self.is_session_ready = is_ready;
+    fn set_session_ready(&mut self) {
+        self.readiness = SessionReadiness::Ready;
+    }
+
+    fn wait_for_logon(&mut self) {
```
vnikonov-devolutions: rename -> `set_waiting_for_logon`

devolutions-session/src/dvc/io.rs
```diff
@@ -1,56 +1,101 @@
...
 
 const DVC_CHANNEL_NAME: &str = "Devolutions::Now::Agent";
+const DVC_RETRY_DELAYS: [Duration; 5] = [
+    Duration::from_millis(250),
+    Duration::from_millis(500),
+    Duration::from_secs(1),
+    Duration::from_secs(2),
+    Duration::from_secs(4),
+];
+const WAIT_TIMEOUT: WAIT_EVENT = WAIT_EVENT(258);
```
vnikonov-devolutions: It is already present in winapi crate `winapi::shared::winerror::WAIT_TIMEOUT`

devolutions-session/src/dvc/io.rs
```diff
@@ -144,6 +189,104 @@ pub fn run_dvc_io(
     }
 }
 
+fn initialize_dvc_with_retry(stop_event: &Event) -> anyhow::Result<Option<DvcIoContext>> {
+    let mut attempt = 1usize;
+
+    loop {
```
vnikonov-devolutions, line 199: I think its missing exit loop condition, we need to limit retries, lets try 5 minutes retry period

vnikonov-devolutions (commented):
Looks good overall, a few minor comments

devolutions-session/src/dvc/io.rs
thenextman: The loop is bounded because it returns `None` after 5 retries. But the spirit of the comment (make the bound explicit) is fair.

For the retry budget, I'm not so sure. Remember we have also RDM with a default success window of 15 seconds. Yes we can bump it client-side but 5 minutes I think is too long. We're still betting on a transient hypothesis here so I'd leave this as-is, let the customer try it and see the results.

devolutions-session/src/dvc/io.rs
```diff
@@ -144,6 +189,104 @@ pub fn run_dvc_io(
...
+    let mut overlapped = OVERLAPPED {
+        hEvent: read_event.raw(),
+        ..Default::default()
+    };
+
+    // Prepare async read operation.
+    // SAFETY: Both `channel_file` and event passed to `overlapped` are valid during this call,
+    // therefore it is safe to call.
+    let read_result: Result<(), windows::core::Error> =
+        unsafe { ReadFile(*channel_file, Some(&mut pdu_chunk_buffer), None, Some(&mut overlapped)) };
```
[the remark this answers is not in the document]
mamoreau-devolutions: Fixed in 53e80391. DVC I/O state is now heap-allocated before the initial `ReadFile`, remains in place for the pending overlapped operation, and is accessed through the boxed context for the lifetime of the I/O loop.

devolutions-agent/src/session_manager.rs
```diff
@@ -169,10 +206,15 @@ impl Task for SessionManager {
                             // Terminate old session process if it is already running.
                             terminate_session_process(&id);
 
+                            let readiness = if session_has_logged_in_user(id.id)? {
+                                SessionReadiness::WaitingForUnlock
+                            } else {
+                                SessionReadiness::WaitingForLogon
+                            };
```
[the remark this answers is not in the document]
mamoreau-devolutions: Fixed in 53e80391. A failure to query the user token is now logged locally; the manager stays alive and waits for either the next logon or unlock notification.

devolutions-agent/src/session_manager.rs
mamoreau-devolutions: Renamed to `set_waiting_for_logon` in 53e80391.

devolutions-session/src/dvc/io.rs
mamoreau-devolutions: Replaced the local literal with the Windows API `WAIT_TIMEOUT` constant in 53e80391.

devolutions-session/src/dvc/io.rs
mamoreau-devolutions: Kept the existing finite five-retry schedule as discussed. The schedule is explicit in `DVC_RETRY_DELAYS` and exhausts after 7.75 seconds of delays.

thenextman (approved):

vnikonov-devolutions (approved):
LGTM

thenextman:
@CBenoit I tagged this as "release required" after merging, not sure if that makes a difference

Not sure the timeline for next Agent release but sooner would be better for this one...

CBenoit:
It actually does not really make a difference anymore since I'm now checking to make a release every Friday (if there is no blocker)... so that would be today 🙂

PR #1856, opened 2026-07-14
fix(dgw): target Gateway hostname when using the credential-injection fake KDC proxy

TheBestTvarynka:
Hi,

I fixed the target hostname for the fake KDC inside the KDC proxy. I tried to write the minimal possible solution.

The problem: during the credentials injection, the RDP client does the Kerberos authentication via a KDC proxy with a fake KDC. In that case, the target hostname (Kerberos service name for the TGS) must be a proxy hostname. But prior to this PR, the KDC proxy always configured the internal KDC to the real target machine hostname. This problem was introduced in #1768.

For example, my local configuration:

* Devolutions-Gateway hostname: `dgateway.tbt.com`.
* Target machine behind the proxy: `WIN-956CQOSSJTF.tbt.com`.

When I connect to the target machine using credentials injection, my FreeRDP command looks like this:

```ps1
.\sdl-freerdp.exe /v:dgateway.tbt.com:8181 /u:fake_user /d:603077ae-b66b-4256-ae48-a458db5b90b0.jet /p:fake_password /sec:nla /log-level:TRACE /sspi-module:"<sspi.dll path>" /pcb:<token>
```

During the FreeRDP's NLA stage, it should perform Kerberos auth against `dgateway.tbt.com` (instead of `WIN-956CQOSSJTF.tbt.com`).

I hope I explained the problem well 😄.

### DIFF
--- a/devolutions-gateway/src/credential_injection_kdc.rs
+++ b/devolutions-gateway/src/credential_injection_kdc.rs
@@ -55,12 +55,6 @@ pub(crate) enum CredentialInjectionKdcResolveError {
     ExpiredCredential { jti: Uuid },
     #[error("credential-injection state is not available for {jti}")]
     NonInjectionCredential { jti: Uuid },
-    #[error("association token for {jti} is not valid for credential injection")]
-    InvalidAssociationToken {
-        jti: Uuid,
-        #[source]
-        source: anyhow::Error,
-    },
     #[error("credential-injection KDC config could not be initialized for {jti}")]
     BuildKdcConfig {
         jti: Uuid,
@@ -444,19 +438,15 @@ fn random_32_bytes() -> Vec<u8> {
 /// through this service, so callers see one handle instead of coordinating a store and a registry.
 #[derive(Debug, Clone)]
 pub struct CredentialService {
+    hostname: String,
     credentials: CredentialStoreHandle,
     sessions: Arc<Mutex<HashMap<Uuid, Arc<CredentialInjectionKdcSession>>>>,
 }
 
-impl Default for CredentialService {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl CredentialService {
-    pub fn new() -> Self {
+    pub fn new(hostname: String) -> Self {
         Self {
+            hostname,
             credentials: CredentialStoreHandle::new(),
             sessions: Arc::new(Mutex::new(HashMap::new())),
         }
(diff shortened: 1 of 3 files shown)


TheBestTvarynka:
@CBenoit kind reminder 🙃

devolutions-gateway/src/credential_injection_kdc.rs
```diff
@@ -444,19 +438,15 @@ fn random_32_bytes() -> Vec<u8> {
 /// through this service, so callers see one handle instead of coordinating a store and a registry.
 #[derive(Debug, Clone)]
 pub struct CredentialService {
+    hostname: String,
```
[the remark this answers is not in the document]
CBenoit: This is something I would prefer we address before merging

devolutions-gateway/src/credential_injection_kdc.rs
```diff
@@ -444,19 +438,15 @@ fn random_32_bytes() -> Vec<u8> {
...
 
-impl Default for CredentialService {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl CredentialService {
-    pub fn new() -> Self {
+    pub fn new(hostname: String) -> Self {
```
CBenoit: issue: I know it’s probably not a real issue today because we don’t change the hostname dynamically typically; that being said ConfHandle supports hot-reloading, and here we retrieve a snapshot of the hostname based on what was configured at startup time. The better style would be to take the ConfHandle and to retrieve the hostname each time it is necessary, so we have an up to date value (`get_conf()` for each request). This is relevant because CredentialService is not short-lived.

CBenoit: question: Also a question I got while reviewing. This is a fine, minimal fix, but it’s worth questioning if it’s not better to use the hostname of the external URL of the listener being used for this request. This makes the fix a bit more involved, and may be a follow up, but it would be "more correct". If a client reaches the gateway through a listener whose external hostname differs from conf.hostname, the SPN mismatch fixed here could resurface.

TheBestTvarynka: fixed in https://github.com/Devolutions/devolutions-gateway/pull/1856/commits/948c516018346d7ef1c23337c16c6b34a7158656

devolutions-gateway/src/credential_injection_kdc.rs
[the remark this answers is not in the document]
TheBestTvarynka: updated comments in https://github.com/Devolutions/devolutions-gateway/pull/1856/commits/34e7cc6dfece2f443d8e5e0caa8d393069250075

CBenoit (approved):
LGTM, thank you!

PR #1857, opened 2026-07-15
feat(dgw): add Session Recording Log artifacts

kristahouse:
Adds server-side support for Session Recording Log recording artifacts.

Gateway can now accept JREC recording pushes with `fileType=slog`, persist the raw UTF-8 NDJSON stream as `.slog`, and serve `.slog` downloads with the `application/x-ndjson` content type. This establishes the Gateway storage contract needed by AD Session Recording Log producers and future viewers. Consumers derive the artifact type from the file extension in `recording.json`.

This PR only adds the Gateway recording storage contract. Token generation remains provided by the host system, such as DVLS. Gateway standalone webapp sandbox wiring will follow separately on top of this server contract.

Issue: DGW-403

### DIFF
--- a/devolutions-gateway/src/api/jrec.rs
+++ b/devolutions-gateway/src/api/jrec.rs
@@ -6,6 +6,7 @@ use std::time::Duration;
 use anyhow::Context as _;
 use axum::extract::ws::{CloseFrame, WebSocket};
 use axum::extract::{self, ConnectInfo, Query, State, WebSocketUpgrade};
+use axum::http::header::{CONTENT_TYPE, HeaderValue};
 use axum::response::Response;
 use axum::routing::{delete, get};
 use axum::{Json, Router};
@@ -511,10 +512,21 @@ where
         return Err(HttpError::not_found().msg("requested file does not exist"));
     }
 
-    let response = tower_http::services::ServeFile::new(path)
+    let mut response = tower_http::services::ServeFile::new(&path)
         .oneshot(request)
         .await
         .map_err(HttpError::internal().err())?;
+
+    let content_type = path
+        .extension()
+        .and_then(RecordingFileType::from_extension)
+        .and_then(RecordingFileType::content_type);
+
+    if let Some(content_type) = content_type {
+        response
+            .headers_mut()
+            .insert(CONTENT_TYPE, HeaderValue::from_static(content_type));
+    }
 
     Ok(response)
 }
--- a/devolutions-gateway/src/token.rs
+++ b/devolutions-gateway/src/token.rs
@@ -280,14 +280,20 @@ pub enum RecordingFileType {
     TRP,
     /// asciinema cast for terminal playback
     Asciicast,
+    /// Session Recording Log NDJSON stream
+    #[serde(rename = "slog")]
+    SessionRecordingLog,
 }
 
 impl RecordingFileType {
+    pub const SLOG_CONTENT_TYPE: &'static str = "application/x-ndjson";
+
     pub const fn format_name(self) -> &'static str {
         match self {
             RecordingFileType::WebM => "WebM",
             RecordingFileType::TRP => "TRP",
             RecordingFileType::Asciicast => "asciicast",
+            RecordingFileType::SessionRecordingLog => "slog",
         }
     }
 
@@ -296,6 +302,24 @@ impl RecordingFileType {
             RecordingFileType::WebM => "webm",
             RecordingFileType::TRP => "trp",
             RecordingFileType::Asciicast => "cast",
+            RecordingFileType::SessionRecordingLog => "slog",
+        }
+    }
+
+    pub fn from_extension(extension: &str) -> Option<Self> {
+        match extension {
+            "webm" => Some(RecordingFileType::WebM),
+            "trp" => Some(RecordingFileType::TRP),
+            "cast" => Some(RecordingFileType::Asciicast),
+            "slog" => Some(RecordingFileType::SessionRecordingLog),
+            _ => None,
+        }
+    }
+
+    pub const fn content_type(self) -> Option<&'static str> {
+        match self {
+            RecordingFileType::SessionRecordingLog => Some(Self::SLOG_CONTENT_TYPE),
+            RecordingFileType::WebM | RecordingFileType::TRP | RecordingFileType::Asciicast => None,
         }
     }
 }


kristahouse:
Implementation notes:
- Adds `RecordingFileType::SessionRecordingLog` with wire value `slog`, stored extension `.slog`, and `application/x-ndjson` pull content type.
- ~~Extends `recording.json` entries with `fileType`, while reading older manifests by inferring missing file types from extensions.~~
  - Removed after review; the file type is derivable from the `fileName` extension, so the manifest is unchanged and consumers derive the type locally.
- Keeps Gateway write behavior as raw byte passthrough; Gateway does not parse or semantically validate SLOG NDJSON.
- ~~Adds standalone webapp server support to issue JREC push tokens through `/jet/webapp/session-token` and exposes `WebApp.EnableSessionRecording` through config/health for the later sandbox UI PR.~~
  - Removed from this PR after review; standalone webapp token/config/health support belongs in the follow-up sandbox PR if needed.
- Follow-up PR will wire the Gateway standalone AD webapp toggle/package integration on top of this server contract.
- Follow-up refactor PR planned: use `RecordingFileType::from_extension` in `streaming.rs` and provide content types for the other recording file types.

Validation:
- `cargo test -p devolutions-gateway --lib`
- `cargo check -p devolutions-gateway`
- `cargo clippy -p devolutions-gateway --tests -- -D warnings`
- Manual smoke test against local Gateway: pushed `fileType=slog`, verified `recording-0.slog` and raw NDJSON contents.

Out of scope:
- `.slog` viewer/rendering.
- DVLS UI/search MVP.
- Gateway standalone webapp checkbox/package wiring.
- Media sidecar/concurrent artifact support.

kristahouse:
Scope update:
- Removed the standalone webapp JREC token/config/health pieces from this PR.
- This PR now contains only the Gateway recording storage contract for `fileType=slog` / `.slog`.
- Host systems such as DVLS remain responsible for minting JREC push tokens.
- Gateway standalone webapp sandbox token/UI wiring will be handled in a separate follow-up PR.

Current changed files:
- `devolutions-gateway/src/api/jrec.rs`
- `devolutions-gateway/src/recording.rs`
- `devolutions-gateway/src/token.rs`

Validation after reducing scope:
- `cargo test -p devolutions-gateway --lib session_recording_log`
- `cargo check -p devolutions-gateway`
- `cargo clippy -p devolutions-gateway --tests -- -D warnings`

kristahouse:
Post-reduction smoke test:
- Built the current PR branch release Gateway executable.
- Started an isolated local Gateway using a copied config on `http://localhost:7173` / `tcp://localhost:8183`.
- Generated a JREC push token externally with `tools/tokengen`.
- Pushed UTF-8 NDJSON to `/jet/jrec/push/{sessionId}?fileType=slog&token=<redacted>`.
- Verified Gateway wrote `recording-0.slog`.
- Verified `recording.json` contains `fileName: "recording-0.slog"` and `fileType: "slog"`.
- Verified raw NDJSON contents were preserved.

devolutions-gateway/src/api/webapp.rs
```diff
@@ -255,6 +255,10 @@ pub(crate) enum SessionTokenContentType {
         /// Default port is `88`.
         krb_kdc: TargetAddr,
     },
+    Jrec {
+        /// Unique ID for this recording session
+        session_id: Uuid,
```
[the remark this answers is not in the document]
kristahouse: Resolved by scope reduction: the `/jet/webapp/session-token` JREC variant was removed from this PR. This PR now only adds the core Gateway `fileType=slog` / `.slog` storage contract; standalone webapp token support will be handled in a separate follow-up PR if needed.

devolutions-gateway/src/api/webapp.rs
```diff
@@ -426,6 +430,20 @@ pub(crate) async fn sign_session_token(
             ContentType::NetScan,
             None,
         ),
+
+        SessionTokenContentType::Jrec { session_id } => (
+            serde_json::json!({
+                "jet_aid": session_id,
+                "jet_rop": RecordingOperation::Push,
+                "jet_reuse": 0,
```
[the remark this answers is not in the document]
kristahouse: Resolved by scope reduction: the `/jet/webapp/session-token` JREC variant was removed from this PR. This PR now only adds the core Gateway `fileType=slog` / `.slog` storage contract; standalone webapp token support will be handled in a separate follow-up PR if needed.

devolutions-gateway/src/recording.rs
```diff
@@ -44,11 +45,70 @@ struct JrecManifest {
...
+    type Error = anyhow::Error;
+
+    fn try_from(file_on_disk: JrecFileOnDisk) -> Result<Self, Self::Error> {
+        let file_type = match file_on_disk.file_type {
+            Some(file_type) => file_type,
+            None => file_on_disk
+                .file_name
+                .rsplit_once('.')
+                .and_then(|(_, extension)| RecordingFileType::from_extension(extension))
+                .with_context(|| format!("missing file type for recording file {}", file_on_disk.file_name))?,
```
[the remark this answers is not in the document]
kristahouse: Updated the context message to describe the actual failure: the manifest reader could not infer the file type from the recording file name/extension. This now covers both missing extensions and unrecognized extensions.

devolutions-gateway/src/api/jrec.rs
```diff
@@ -511,11 +512,22 @@ where
...
+    let mut response = tower_http::services::ServeFile::new(path)
         .oneshot(request)
         .await
         .map_err(HttpError::internal().err())?;
 
+    if let Some(content_type) = content_type {
+        response
+            .headers_mut()
+            .insert(CONTENT_TYPE, HeaderValue::from_static(content_type));
+    }
```
CBenoit, line 529: suggestion: This reads more naturally

```suggestion
    let mut response = tower_http::services::ServeFile::new(path)
.oneshot(request)
.await
.map_err(HttpError::internal().err())?;

    let content_type = path
        .extension()
        .and_then(RecordingFileType::from_extension)
        .and_then(RecordingFileType::content_type);

    if let Some(content_type) = content_type {
        response
            .headers_mut()
            .insert(CONTENT_TYPE, HeaderValue::from_static(content_type));
    }
```

devolutions-gateway/src/api/jrec.rs
```diff
@@ -511,10 +512,21 @@ where
...
     }
 
-    let response = tower_http::services::ServeFile::new(path)
-        .oneshot(request)
-        .await
-        .map_err(HttpError::internal().err())?;
+    let mut response = tower_http::services::ServeFile::new(path)
+.oneshot(request)
+.await
+.map_err(HttpError::internal().err())?;
```
CBenoit:
```suggestion
    let mut response = tower_http::services::ServeFile::new(path)
        .oneshot(request)
        .await
        .map_err(HttpError::internal().err())?;
```

devolutions-gateway/src/token.rs
```diff
@@ -296,6 +302,24 @@ impl RecordingFileType {
...
+
+    pub fn from_extension(extension: &str) -> Option<Self> {
+        match extension {
+            "webm" => Some(RecordingFileType::WebM),
+            "trp" => Some(RecordingFileType::TRP),
+            "cast" => Some(RecordingFileType::Asciicast),
+            "slog" => Some(RecordingFileType::SessionRecordingLog),
+            _ => None,
+        }
+    }
```
CBenoit, line 317: thought: There is already a similar pattern that could be refactored to use this new helper.

In `validate_streaming_file` in devolutions-gateway/src/streaming.rs:

```rust
// devolutions-gateway/src/streaming.rs
if !(path_extension == RecordingFileType::WebM.extension()
    || path_extension == RecordingFileType::Asciicast.extension()
    || path_extension == RecordingFileType::TRP.extension())
{
    anyhow::bail!("invalid file type");
}
```

The same file also branches again later on the same extensions to pick a `StreamingType`.

Now that we have a centralized `from_extension`, both of those places can be simplified to:

```rust
let file_type = RecordingFileType::from_extension(path_extension)
    .ok_or_else(|| anyhow::anyhow!("invalid file type"))?;
```

and then the `StreamingType` (and terminal input type) can be derived from the resulting `RecordingFileType` instead of comparing strings repeatedly.

For a follow up refactor PR.

devolutions-gateway/src/token.rs
```diff
@@ -1866,4 +1890,20 @@ mod tests {
...
+
+        let serialized =
+            serde_json::to_value(RecordingFileType::SessionRecordingLog).expect("file type should serialize");
+        assert_eq!(serialized, serde_json::json!("slog"));
+        assert_eq!(RecordingFileType::SessionRecordingLog.extension(), "slog");
+        assert_eq!(
+            RecordingFileType::SessionRecordingLog.content_type(),
+            Some(RecordingFileType::SLOG_CONTENT_TYPE)
+        );
+    }
```
CBenoit: nitpick: This test is very low value: it only repeats the definitions.
```suggestion
```

thought: A broader snapshot test harness for various tokens shapes could be useful. I think we already have something similar in the integration tests ("dvls something" module). Ideally: test at the boundaries, the test will be more useful if it can be used to detect an interop regression (i.e.: don’t `use` constants to regenerate tokens understood by Gateway _today_, provide a concrete token shape actually generated by DVLS today, and crucially, _"eventually yesterday"_ and that must be understood by Gateway _today and tomorrow_).

devolutions-gateway/src/recording.rs
CBenoit, line 1: issue: Looks more complex than it really needs to be.

rationale: Here we are duplicating the manifest structures. We effectively get two DTOs (in practice): a filesystem DTO where file_type is optional, and a HTTP API DTO where file_type is always present.

My understanding is that you plan on having consumer code on DVLS side that is relying on the file_type field to branch on the recorder (webm, slog, cast…).

- We don’t have to duplicate the definitions to achieve that, it’s possible to manually implement serde’s Serialize and Deserialize traits.
- I’m also unsure it’s really worth doing that at all: as observed above, we are deriving information already available: the file extension that is found in the `file_name` field. We don’t have to persist that separately in the manifests.

suggestion: Preferred: do not add `file_type`, derive it locally, and revert this file.
It’s typically a oneliner in any language:
```js
type FileType = 'slog' | 'webm' | 'cast' | 'trp';
const fileType = path.extname(fileName).slice(1) as FileType;
```
```c#
string fileType = Path.GetExtension(fileName).TrimStart('.');
```

Alternative: if the API requires file_type: we can consider retaining it, but simplify deserialization and validate consistency.

devolutions-gateway/src/token.rs
```diff
@@ -296,6 +302,24 @@ impl RecordingFileType {
...
+            _ => None,
+        }
+    }
+
+    pub const fn content_type(self) -> Option<&'static str> {
+        match self {
+            RecordingFileType::SessionRecordingLog => Some(Self::SLOG_CONTENT_TYPE),
+            RecordingFileType::WebM | RecordingFileType::TRP | RecordingFileType::Asciicast => None,
         }
     }
```
CBenoit, line 324: thought: As a follow up, we can, and arguably should, provide content types for the others too.

CBenoit (commented):
Thank you!

I’m sorry, I pushed some fixup commits via the GitHub UI hoping to merge immediately, but I think there is a bigger change to make before merging. Can you look at my comments and tell me how that sounds to you? 🙏

devolutions-gateway/src/token.rs
kristahouse: Agreed — good catch. `validate_streaming_file` (and the later `StreamingType` branching) can be simplified with `from_extension` and derive the streaming/input type from the resulting `RecordingFileType`. I'll keep that for a follow-up refactor PR as suggested.

devolutions-gateway/src/token.rs
kristahouse: Agreed — removed the test. And noted on the snapshot harness idea: a boundary-level interop test using concrete tokens as generated by DVLS (frozen fixtures rather than regenerated from today's constants) would catch real regressions. Good candidate for a follow-up alongside the existing integration test module.

devolutions-gateway/src/recording.rs
kristahouse: Agreed — went with the preferred option: `fileType` removed from the manifest and `recording.rs` fully reverted. Consumers will derive the type from the `fileName` extension as you showed. This also drops the duplicated on-disk DTOs and the backward-compat inference code. The PR now only touches `token.rs` (enum variant + helpers) and `jrec.rs` (content type on pull).

devolutions-gateway/src/token.rs
kristahouse: Agreed — worth doing for the others too (e.g. `video/webm` for WebM, `application/x-asciicast` for cast, and something suitable for TRP). Since it changes pull behavior for existing recording types, I'll bundle it with the `from_extension` refactor in the follow-up PR.

CBenoit (approved):
LGTM!

PR #1865, opened 2026-07-24
feat(dgw): provision the target-side KDC for credential injection

irvingoujAtDevolution:
WIP NOT READY


devolutions-gateway/src/provisioning.rs
```diff
@@ -0,0 +1,81 @@
...
+
+pub(crate) type ArcProvisioningEntry = ArcTokenKeyedEntry<ProvisioningEntry>;
+pub(crate) type WeakProvisioningEntry = Weak<TokenKeyedEntry<ProvisioningEntry>>;
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum InsertError {
+    #[error(transparent)]
+    InvalidToken(#[from] TokenKeyError),
+    #[error("credential encryption failed")]
+    Internal(#[source] anyhow::Error),
```
CBenoit: suggestion: Based on the error description, "Internal" is a bit too broad. Could be more specific as `CredentialEncryption`.

devolutions-gateway/src/provisioning.rs
```diff
@@ -0,0 +1,81 @@
...
+use devolutions_gateway_task::Task;
+use uuid::Uuid;
+
+use crate::credential::{AppCredentialMapping, CleartextAppCredentialMapping};
+use crate::target_connection_options::TargetConnectionOptions;
+use crate::token_keyed_store::{ArcTokenKeyedEntry, TokenKeyError, TokenKeyedEntry, TokenKeyedStore};
+
+#[derive(Debug)]
+pub(crate) struct ProvisioningEntry {
+    pub(crate) mapping: Option<AppCredentialMapping>,
```
CBenoit: naming: `mapping` sounds like it could be any other kind of mapping, maybe this should be renamed to be slightly more specific (e.g.: `cred_mapping`)

devolutions-gateway/src/credential_injection.rs
```diff
@@ -435,99 +438,100 @@ fn random_32_bytes() -> Vec<u8> {
...
+/// A cached Kerberos session tagged with the provisioning entry it was derived for.
+///
+/// The weak reference is how we notice re-provisioning without the two stores talking to each other:
+/// a fresh provisioning under the same JTI produces a new entry, so the cached session stops matching
+/// and gets re-derived on the next lookup.
+#[derive(Debug)]
+struct CachedSession {
+    entry: WeakProvisioningEntry,
+    session: Arc<CredentialInjectionKdcSession>,
 }
```
CBenoit, line 481: question: This logic sounds a bit weak to me. If some other part of the code ends up holding a "strong" Arc on the same provisioning entry, then this logic will fail at detecting the re-provisioning happened. Also, this raises another question: do we need to cache the provisioning entry, at all, if we can just lookup again? And also, what are we trying to achieve exactly? What is the use case for re-provisioning? What happens when this happens in the middle of credential injection? Is it really something we want to deal with? I appreciate hardened code, but I’m not really sure this one makes sense

devolutions-gateway/src/provisioning.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub(crate) connection_options: Option<TargetConnectionOptions>,
```
CBenoit: suggestion: split this into two token-keyed stores, one per provisioned payload.

Two independent `Option`s under one key is really two separate provisioning events sharing a JTI. Rather than encoding "the other one wasn't provided" as `None`, let's split the preflight operation into `provision-credentials` and `provision-connection-options`, backed by two `TokenKeyedStore`s keyed by the same JTI:

- `TokenKeyedStore<AppCredentialMapping>`: stays behind `ProvisioningStore`, which remains the encryption boundary.
- `TokenKeyedStore<TargetConnectionOptions>`: no crypto dependency at all, which actually strengthens the "master key never leaves this module" property claimed in the doc comment above.

Rationale:

- Distinct trust semantics. One half is master-key-encrypted secret material, the other is plaintext routing metadata. Separate operations let them be authorized, logged and audited differently.
- Protocol variance. `TargetConnectionOptions` will grow per-protocol shapes (SSH, VNC, …). A dedicated operation gives it its own schema without dragging credential fields along.
- Independent provisioning. DVLS may know one before the other; today the second call has to either re-send the first payload or clobber it with `None`.

Keep replace semantics: no merge API. Merging in place would require `Arc<Mutex<TokenKeyedEntry<V>>>` so entries could be mutated, on top of the store-level mutex. That turns every read into a lock+clone (or a guard whose lifetime has to be threaded through the CredSSP paths) and forces a merge contract onto `V`, which would undo the opacity the generic store just bought us. Immutable snapshots per entry are much friendlier to the async consumers.

Suggested surface: one insert entry point dispatching on operation kind, plus `get_credentials(jti)` / `get_connection_options(jti)` and a combined getter returning a unifying view. The RDP path wants both; the SSH/VNC paths will likely want only the options half. Two `cleanup_task()`s (or one sweeping both) is a trivial cost.

TTLs may diverge: that's fine. I initially thought we should keep the longer of the two, but a missing half is already a normal, handled outcome: `credential_injection_kerberos_configs` errors out (and per my other review comment, should arguably fall back to NTLM) when `krb_kdc` is absent. So an early-expired half degrades into a path that has to exist anyway. Silently extending one store's TTL to match the other would also mean retaining provisioned data longer than the API user asked for. We should document on the preflight API that both operations should carry the same `timeToLive`, and `warn!` with the JTI when a lookup finds one half but not the other, so a support case reads "connection options expired before credentials" instead of the generic `krb_kdc` context error.

devolutions-gateway/src/credential_injection.rs
CBenoit: With the two-store split and replace semantics below, I think we can drop entry-identity tracking entirely, see https://github.com/Devolutions/devolutions-gateway/pull/1865/changes#r3657581558

devolutions-gateway/src/provisioning.rs
irvingoujAtDevolution: It does feel like a bad design, let me update accordingly

devolutions-gateway/src/target_connection_options.rs
```diff
@@ -0,0 +1,108 @@
...
+        if let Some(krb_kdc) = &krb_kdc {
+            if !is_supported_krb_kdc_scheme(krb_kdc.scheme()) {
+                return Err(InvalidKdcAddr::UnsupportedScheme(krb_kdc.scheme().to_owned()));
+            }
+            if krb_kdc.host().is_empty() {
+                return Err(InvalidKdcAddr::MissingHost(krb_kdc.as_str().to_owned()));
+            }
+            // The target-side CredSSP leg turns this into a URL. Reject a value that won't parse here,
+            // so provisioning fails fast instead of at session start.
+            url::Url::try_from(krb_kdc).map_err(|_| InvalidKdcAddr::NotAUrl(krb_kdc.as_str().to_owned()))?;
```
CBenoit, line 25: question: Is it possible for a TargetAddr to "not be a URL"? (I think it already guarantees a shape like: `<scheme>://<host>:<port>`

devolutions-gateway/src/target_connection_options.rs
```diff
@@ -0,0 +1,108 @@
...
+/// How the Gateway's internal client should reach the target, provisioned alongside the credentials.
+///
+/// The KDC address is fully validated at construction — supported scheme, a host, and a parseable
+/// URL — so the rest of the code can trust `krb_kdc` without re-checking it or failing late when a
+/// session starts.
+#[derive(Debug, Clone, Deserialize)]
+#[serde(try_from = "RawTargetConnectionOptions")]
+pub(crate) struct TargetConnectionOptions {
+    krb_kdc: Option<TargetAddr>,
+}
```
CBenoit, line 12: praise: It’s robust, fail-fast design. The type has good invariants.

devolutions-gateway/src/token_keyed_store.rs
```diff
@@ -0,0 +1,164 @@
...
+    use super::*;
+
+    fn token(jti: Uuid) -> String {
+        let engine = base64::engine::general_purpose::URL_SAFE_NO_PAD;
+        let header = engine.encode(r#"{"alg":"RS256"}"#);
+        let payload = engine
+            .encode(serde_json::to_vec(&serde_json::json!({ "jti": jti })).expect("token payload should serialize"));
+        let signature = engine.encode(b"signature");
+        format!("{header}.{payload}.{signature}")
+    }
```
CBenoit: suggestion: You can hardcode this JWT into a constant instead. It’s smaller and more self-contained. You also need to hardcode the JTI, but I think it’s not a problem.

devolutions-gateway/src/service.rs
```diff
@@ -267,7 +267,8 @@ async fn spawn_tasks(conf_handle: ConfHandle) -> anyhow::Result<Tasks> {
             .await
             .context("failed to initialize traffic audit manager")?;
 
-    let credentials = devolutions_gateway::credential_injection_kdc::CredentialService::new();
+    let provisioning = devolutions_gateway::provisioning::ProvisioningStore::new();
+    let credential_injection = devolutions_gateway::credential_injection_kdc::CredentialInjectionKdcService::new();
```
CBenoit: suggestion: This connects to what I suggested above about having two TokenKeyedStore, but I think the provisioning facade could assemble everything together:

```rust
#[derive(Debug, Clone)]
pub struct Provisioning {
    credentials: TokenKeyedStore<AppCredentialMapping>,
    connection_options: TokenKeyedStore<TargetConnectionOptions>,
    kdc_sessions: CredentialInjectionKdcService,
}
```

With a few helper functions to access things in a consistent manner; we are still keeping concerns separated without needing extra threading.

irvingoujAtDevolution:
@CBenoit You were very right. I revisited the design — I couldn't stand the complexity the previous one introduced, so I did a big refactor to get it under control.

Now it's three layers:

1. Provisioning (`ProvisioningStore`) — holds the parsed provisioning data (`ProvisionedConnection`), protocol-agnostic, so it's good for future extension.
2. Credential injection (`CredentialInjection`) — built from the provisioning data.
3. Synthetic KDC registry (`SyntheticKdcRegistry`) — keyed by JTI.

We don't ad-hoc retrieve the KDC and credential info to build a Kerberos injection session anymore. We construct it once from the provisioning store and register it across handlers, so both the RDP path and the KDC-proxy path use the same session instead of each rebuilding its own.

And we don't mix credential injection and Kerberos the same ad-hoc way — Kerberos is officially an enum variant of `CredentialInjection` now (Kerberos vs NTLM), handled through the enum for better semantics.

Tested end to end and all working. I think this is a much cleaner design.

irvingoujAtDevolution:
Did the split — `provision-credentials` and `provision-connection-options` are separate preflight ops now, each backed by its own JTI-keyed store ([f804807a](https://github.com/Devolutions/devolutions-gateway/commit/f804807a)). Credentials keep the encryption boundary; connection options are plaintext routing metadata with no crypto dependency. Each half is provisioned, expires and is replaced on its own, and since preflight is batched the caller just sends both in one request. `get()` folds them back together for the RDP path — credentials required, connection options optional.

Synthetic KDC registry stays its own handle, like we agreed.

Tested end to end: updated the DVLS side to send the two ops, and RDP Kerberos credential injection works through the gateway against a lab DC.

irvingoujAtDevolution:
Superseded by #1895 (merged, shipped in v2026.3.0).

PR #1869, opened 2026-07-27
fix(agent): compare broker client executables by file identity

vnikonov-devolutions:
The package broker authenticated pipe clients by comparing executable path strings using a lossy, ASCII-case-insensitive comparison. This was fragile for a security-relevant identity check, as it could be tripped up by Unicode casing rules, invalid UTF-8, and other path representation edge cases.

The broker now opens both executables and compares their file identity directly (volume serial number + 128-bit file ID via `GetFileInformationByHandleEx`/`FILE_ID_INFO`), which is immune to these edge cases.

Issue: DGW-419

### DIFF
--- a/devolutions-agent/src/broker/auth.rs
+++ b/devolutions-agent/src/broker/auth.rs
@@ -8,6 +8,7 @@ use tokio::net::windows::named_pipe::NamedPipeServer;
 use tracing::{debug, warn};
 use win_api_wrappers::process::Process;
 use windows::Win32::Security::TOKEN_QUERY;
+use windows::Win32::Storage::FileSystem::FILE_ID_INFO;
 use windows::Win32::System::Threading::PROCESS_QUERY_LIMITED_INFORMATION;
 
 use crate::code_signing::validate_devolutions_authenticode_signature;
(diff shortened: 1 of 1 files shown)


vnikonov-devolutions:
Implementation notes:
- Replaces the `canonicalize` + `to_string_lossy().eq_ignore_ascii_case()` path comparison with a handle-based identity check, addressing https://github.com/Devolutions/devolutions-gateway/pull/1847#discussion_r3628014190 and https://github.com/Devolutions/devolutions-gateway/pull/1847#discussion_r3627971206.
- `file_id()` opens the file with `FILE_READ_ATTRIBUTES` access and full sharing, then queries `FileIdInfo` via `GetFileInformationByHandleEx`. Identity = volume serial number + 128-bit file ID, robust against Unicode casing, invalid UTF-8, 8.3 short names, symlinks/junctions, and subst drives.
- The redundant `canonicalize_for_comparison` helper and `same_windows_path` are removed.
- Tests: two new unit tests verify that different path spellings of the same file (casing, redundant `.` component) yield the same identity, and that distinct files yield different identities.

CBenoit (approved):
LGTM!

PR #1873, opened 2026-07-27
fix(agent): create broker temp script atomically with target SDDL

vnikonov-devolutions:
The package broker's temporary scripts are now created atomically with their final security descriptor, instead of being created with the directory-inherited ACL and tightened afterwards.
This removes the short window during which the script existed with inherited permissions, hardening script execution for users of the Devolutions Agent package broker.

Behavior is otherwise unchanged: SYSTEM, Administrators, and the owner keep full control, and interactive users retain the read access needed to execute the script.

Issue: DGW-410

### DIFF
--- a/crates/devolutions-agent-shared/src/temp_file.rs
+++ b/crates/devolutions-agent-shared/src/temp_file.rs
@@ -33,6 +33,32 @@ impl TmpFileGuard {
         Ok(Self { path })
     }
 
+    /// Creates the temporary file using a caller-provided creation function.
+    ///
+    /// The `make` function receives the candidate path and must create the file at that path
+    /// with create-new semantics, failing with `io::ErrorKind::AlreadyExists` if the file
+    /// already exists so that a new random name can be retried.
+    ///
+    /// This allows creating the file atomically with non-default properties, such as a specific
+    /// Windows security descriptor applied at creation time.
+    pub fn with_prefix_in_using<F>(prefix: &str, extension: &str, temp_dir: Option<&Path>, make: F) -> io::Result<Self>
+    where
+        F: FnMut(&Path) -> io::Result<fs::File>,
+    {
+        let suffix = format!(".{extension}");
+        let mut builder = tempfile::Builder::new();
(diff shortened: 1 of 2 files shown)


vnikonov-devolutions:
Implementation notes:

- The temp script file is now created via `CreateFileW` with `CREATE_NEW` and a `SECURITY_ATTRIBUTES` built from the same SDDL as before, so the target DACL is applied atomically at creation; `protect_temp_script`/`set_file_dacl` (post-creation `SetNamedSecurityInfoW`) are removed.
- Random-name generation and collision retry are still delegated to the `tempfile` crate via a new `TmpFileGuard::with_prefix_in_using` constructor (`tempfile::Builder::make_in`) that accepts a custom file-creation closure; the closure fails with `ErrorKind::AlreadyExists` on collision so `tempfile` retries with a new random name.
- The SDDL changed from `D:PAI(...)` to `D:P(...)`: the `AI` (auto-inherited) control flag was an artifact of the old post-creation flow and is meaningless for a protected DACL applied at birth. The ACE list is identical.

Testing:
- New unit test `temp_script_is_created_with_target_dacl` reads back the DACL of a freshly created script (`GetNamedSecurityInfoW` + SDDL round-trip) and asserts it exactly matches the target descriptor — proving the file is born with it.
- New `#[ignore]` integration test `temp_script_is_readable_by_user_token_in_system_mode` for real SYSTEM mode: it asserts the process runs as SYSTEM, creates the script in the broker's default temp dir, obtains the non-elevated user token of an active interactive session via `WTSQueryUserToken` (SeTcb), impersonates it, and reads the script. Run manually with:
  `psexec -s cargo test -p devolutions-agent temp_script_is_readable_by_user_token_in_system_mode -- --ignored`
  (Notably this also covers Win11's `C:\Windows\SystemTemp`, where the user cannot list the directory but can open the file by full path thanks to bypass-traverse-checking + the file's `FR;;;BU` ACE.)

mamoreau-devolutions (approved):

PR #1874, opened 2026-07-27
feat(agent): add Chocolatey package broker support

vnikonov-devolutions:
Adds Chocolatey as a Windows package broker manager so the Devolutions Agent can build and execute install, update, and uninstall operations through `choco.exe`.

Issue: DGW-424

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4714,9 +4714,9 @@ dependencies = [
 
 [[package]]
 name = "now-policy"
-version = "0.1.0"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9d87d7ba63d8fcbbae125f975c5e30a54ed4507027dd70d6daa0aa63003cac2e"
+checksum = "2cc04b547209a37ecd3993eb47773ec6fa38f8cac178479079f650555bbedbbb"
 dependencies = [
  "chrono",
  "schemars",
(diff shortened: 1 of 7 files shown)


vnikonov-devolutions:
Implementation notes:
- Adds a dedicated Chocolatey command builder for install, upgrade, and uninstall operations.
- Mirrors UniGetUI Chocolatey semantics where applicable: `install`, `upgrade`, `uninstall`, `-y`, `--no-progress`, `--source`, `--notsilent`, `--forcex86`, `--prerelease`, `--ignore-checksums --force`, and version pinning with `--allow-downgrade`.
- Rejects unsupported or security-sensitive options explicitly instead of silently ignoring them: package channels, user scope, ARM64/neutral architecture, architecture on uninstall, custom install locations, arbitrary custom parameters, no-upgrade, uninstall-previous, and install-only flags on uninstall.
- Extends Windows executor preparation so Chocolatey commands run through the protected temporary batch wrapper and resolve `cmd.exe` from System32.
- Resolves Chocolatey through `ChocolateyInstall`, defaulting to `%ProgramData%\chocolatey`, and currently accepts only the trusted system Chocolatey folder path; broader SID/ACL-based trust checks are intentionally left for follow-up.
- Adds focused unit coverage for Chocolatey command construction, rejection behavior, capabilities, and executor script preparation.

Testing:
- `cargo test -p devolutions-agent chocolatey -- --nocapture`
- `cargo test -p devolutions-agent broker::command_builder -- --nocapture`
- `cargo clippy -p devolutions-agent --tests -- -D warnings`

Protocol dependency:
- This draft depends on the shared `now-policy-api` crate publishing `ManagerName::Chocolatey`.
- Local validation used temporary root `[patch.crates-io]` entries pointing to `D:\now-proto\policies\rust\...`; those local patches are not committed to this PR.

mamoreau-devolutions (approved):

PR #1878, opened 2026-07-28
feat(agent): add pip package broker support

vnikonov-devolutions:
Adds Devolutions Agent package broker support for pip package operations, allowing approved Python packages to be installed, updated, and uninstalled through the broker using the released NOW policy v0.2 protocol models.

Pip support is limited to standard user-scope operations. Elevated/admin and unsupported pip options are rejected explicitly to preserve the broker security model.

Issue: DGW-428

### DIFF
--- a/devolutions-agent/src/broker/command_builder/mod.rs
+++ b/devolutions-agent/src/broker/command_builder/mod.rs
@@ -5,6 +5,7 @@
 
 pub mod chocolatey;
 pub mod dotnet;
+pub mod pip;
 pub mod powershell;
 pub mod scoop;
 pub mod vcpkg;
@@ -24,6 +25,7 @@ pub fn build_command(request: &PackageRequest) -> anyhow::Result<Vec<String>> {
         ManagerName::PowerShell => powershell::build_powershell5_command(request),
         ManagerName::PowerShell7 => powershell::build_powershell7_command(request),
         ManagerName::Chocolatey => chocolatey::build_chocolatey_command(request),
+        ManagerName::Pip => pip::build_pip_command(request),
         ManagerName::Scoop => scoop::build_scoop_command(request),
         ManagerName::Vcpkg => vcpkg::build_vcpkg_command(request),
         unsupported => bail!("package manager is not supported by the broker: {unsupported}"),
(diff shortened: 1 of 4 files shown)


vnikonov-devolutions:
Implementation notes:
- Mirrors UniGetUI pip operation semantics with `python.exe -m pip`: install uses `pip install <id>[==version]`, update uses `pip install --upgrade <id>[==version]`, and uninstall uses `pip uninstall <id>`.
- Adds noninteractive defaults (`--no-input`, `--no-color`, `--no-cache`) and `--yes` for uninstall; install/update support `--pre` and user scope via `--user`.
- Rejects elevated/admin and machine-scope pip operations rather than attempting to secure arbitrary Python discovery under an elevated token.
- Rejects source URLs/custom sources, non-neutral architectures, channels, uninstall version pinning, interactive mode, skip-hash-check, custom install locations, custom parameters, uninstall-previous, no-upgrade, and pre-release uninstall.
- Windows execution wraps pip in the existing protected UTF-8 batch-script path and resolves `python.exe` from the target user's PATH.
- Uses released `now-policy`, `now-policy-api`, and `now-policy-server-template` v0.2 crates; no local protocol patches or vendored protocol crates are used.

Validation:
- `cargo test -p devolutions-agent pip --lib`
- `cargo test -p devolutions-agent broker::command_builder --lib`
- `cargo clippy -p devolutions-agent --tests -- -D warnings`

vnikonov-devolutions:
Addressed Copilot review feedback:
- Added pip `--isolated` so user-controlled pip environment variables and config files cannot override broker-validated source/location behavior.
- Made `--user` unconditional for install/update because pip is advertised and allowed as user-scope only.
- Validates pip package IDs as plain Python distribution names and versions as constrained version components before constructing exact requirements.
- Avoids serializing pip arguments through `cmd.exe`; the Windows executor now resolves `python.exe` from the target user's PATH and invokes it directly.
- Added regression tests for package ID/requirement injection, unsafe version characters, and direct Python execution.

Validation:
- `cargo test -p devolutions-agent pip --lib`
- `cargo test -p devolutions-agent broker::command_builder --lib`
- `cargo clippy -p devolutions-agent --tests -- -D warnings`

mamoreau-devolutions (approved):

mamoreau-devolutions (approved):

CBenoit (approved):

PR #1886, opened 2026-07-29
fix(dgw,agent): make DNS route matching explicit

irvingoujAtDevolution:
Makes Agent Tunnel DNS routing explicit and predictable:
- `example.com` matches only `example.com`.
- `*.example.com` matches subdomains, but not the bare parent.
- Domain detection suggests the explicit wildcard route instead of silently widening a plain name.

This moves Agent Tunnel to protocol/ALPN v2 so mixed old/new semantics fail clearly. The installer remains one-way and never imports stale domains from `agent.json`.

Tests: full Rust workspace, real Agent/Gateway E2E, installer compile, localization audit, x64 MSI build, and lab Gateway/Agent deployment.


irvingoujAtDevolution:
Implementation notes:
- Explicit domains win; auto-detection only runs when none are configured.
- Disabled auto-detection stays quiet for subnet-only agents.
- Added regression coverage for both cases.

CBenoit (approved):

irvingoujAtDevolution:
Updated implementation notes:
- Fresh enrollment defaults to no domain routes and `AutoDetectDomain=false`.
- CLI and silent MSI support explicit domains; interactive MSI offers detection as an unchecked suggestion.
- Existing domains, Back/Next state, empty edits, and long/localized labels are handled explicitly.
- Tunnel CustomActionData is encoded before crossing into the deferred SYSTEM action.

irvingoujAtDevolution:
Follow-up:
- Installer no longer reads or merges domains from an existing `agent.json`.
- Legacy `*.example.com` config is normalized to `example.com` at the Agent-to-Gateway boundary.
- DVLS wildcard display is split into `Devolutions/RDM` branch `dvls/agent-domain-wildcard-display`.

irvingoujAtDevolution:
Final DNS route semantics:
- Plain names are exact-only.
- Only an explicit `*.` route includes subdomains.
- Exact routes outrank wildcard candidates.
- Auto-detection of `ad.it-help.ninja` suggests `*.ad.it-help.ninja`.
- Protocol and ALPN are now v2; Gateway and Agent must be upgraded together.

The obsolete DVLS display-only PR was closed because DVLS should show the literal API route.

irvingoujAtDevolution:
Review follow-up is in `69f6f188`:
- DNS route syntax is validated at wire decode, CLI enrollment, runtime config, and MSI boundaries.
- Rejects ambiguous forms such as `*.`, `*.*`, `foo.*`, and empty labels.
- Invalid auto-detected routes are ignored instead of breaking the control connection.
- Fixed the registry specificity doc, installer comment wording, log punctuation, and redundant comments.

Final MSI SHA-256: `78F57C8E7C9C349D6C81361AE8401B79EA58B3D15BB5091D7A67A45E0A4EEE09`.

irvingoujAtDevolution:
Real lab E2E passed on the final branch using two Agents and distinct marker servers:

- exact name → exact Agent (`FROM-DC`)
- child of exact name → no Agent route
- wildcard child → wildcard Agent (`FROM-AGENT`)
- wildcard parent → no Agent route
- exact + wildcard overlap → exact Agent wins
- unmatched name → no Agent route

Gateway could not resolve any `.test` name directly. Positive cases had both `Routing via agent tunnel` and `Proxy stream established`; negative cases had no Agent routing and failed direct DNS lookup. Lab configs, hosts, services, token-validation setting, and marker processes were restored afterward.

PR #1892, opened 2026-07-29
fix(agent): enforce client identity for user-scope execution when broker is not SYSTEM

vnikonov-devolutions:
Ensures package operations always run in the requesting user's own context.

In service (SYSTEM) mode the broker already launches operations with the client user's real session token, so user-scope installs land in the correct profile (`%LOCALAPPDATA%`, HKCU) and interactive installer UIs appear on the user's desktop rather than in session 0.

This change closes a gap in the non-SYSTEM execution path, which is selected at runtime whenever the broker process does not run as SYSTEM (e.g. the agent binary launched manually as a regular user instead of as the Windows service — in any build profile, release included): the executor ran commands under the broker process token regardless of which user connected to the pipe, so a user-scope install requested by a different user would silently land in the broker user's profile. Such requests are now rejected with a clear error.

The error reported when the target user has no active logon session is also clarified, since user-scope and interactive operations require the user to be logged on.

Issue: DGW-436

### DIFF
--- a/devolutions-agent/src/broker/executor/windows/mod.rs
+++ b/devolutions-agent/src/broker/executor/windows/mod.rs
@@ -168,7 +168,11 @@ fn execute_as_system(
     Ok(output)
 }
 
-/// Execute a command as the current user (development mode).
+/// Execute a command as the current user (non-SYSTEM mode).
+///
+/// This path is taken whenever the broker does not run as SYSTEM, regardless of build
+/// profile — e.g. when the agent binary is launched manually as a regular user instead
+/// of as the Windows service.
 ///
 /// Opens the current process token and uses the same `create_process_as_user`
 /// code path as SYSTEM mode, ensuring consistent behavior (environment, desktop, flags).
@@ -183,7 +187,7 @@ fn execute_as_current_user(
 ) -> anyhow::Result<ExecutionOutput> {
     info!(
         effective_user = %ctx.effective_user,
-        "Executing command as current user (dev mode)"
+        "Executing command as current user (non-SYSTEM mode)"
     );
 
     let token = Process::current_process()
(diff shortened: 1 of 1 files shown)


vnikonov-devolutions:
Implementation notes (audit findings for DGW-436):

**Verified as already correct (SYSTEM/service mode)** — ``execute_as_system`` in ``devolutions-agent/src/broker/executor/windows/mod.rs``:
- Obtains the client user's real session token via ``WTSQueryUserToken`` (matched by SID, not name), duplicates it as primary, and assigns the session ID before ``CreateProcessAsUserW``.
- Loads the environment block from that token, so ``%LOCALAPPDATA%``, ``%USERPROFILE%``, and HKCU resolve to the actual client user — per-user scope installs land in the right profile.
- Processes are created on ``WinSta0\Default`` in the user's session, so ``--interactive`` installer UIs are visible on the user's desktop, never stuck in session 0.
- No active logon session for the user → the operation fails with a clear error instead of installing into the SYSTEM profile (error message clarified in this PR).

**Gap fixed (non-SYSTEM mode)** — the execution path is selected at *runtime* by checking whether the process token is LocalSystem, not by a cargo feature; a release binary launched as a regular user takes the same path. There, ``execute_as_current_user`` ignored ``ctx.user_sid`` and ran everything under the broker process token. Since the pipe DACL grants BUILTIN\Users connect access, a client running as a different user would get its user-scope install written into the broker user's profile. The executor now compares the authenticated pipe client SID against the broker process token SID and rejects mismatches.

**Why not a capability flag**: ``ManagerCapability`` in ``now-policy-api`` 0.2 has no interactive/user-scope execution field to gate on, and interactive mode is already gated per-manager (builders reject it where unsupported) and per-policy (``AllowInteractive``). The remaining gap was an executor identity check, addressed here.

**Testing**: new unit test ``non_system_mode_rejects_client_user_different_from_broker_user`` (uses the Everyone SID, which never matches the test process user). Existing SYSTEM-mode behavior is covered by the ignored ``psexec -s`` test. ``cargo clippy --tests -D warnings`` clean.

CBenoit (commented):
No findings.

The non-SYSTEM executor now compares the authenticated pipe client's captured SID with the broker process token SID before running the plan, and the added Windows test verifies mismatched identities are rejected. The protocol/API pass found no specification-defined discrepancy.

CBenoit (approved):
LGTM!

PR #1894, opened 2026-07-29
chore(openapi): regenerate the spec and clients for the provisioned target options

irvingoujAtDevolution:
WIP NOT READY


irvingoujAtDevolution:
Folded into #1895 — the regeneration now lands in the same PR as the schema change that causes it, so every layer's spec and clients agree.

PR #1900, opened 2026-07-30
fix(dgw): reuse credential injection across reconnects

irvingoujAtDevolution:
Keep provisioned injection mappings alive for as long as Gateway still accepts the association token, so native RDP reconnects reuse the same JWT and get injected again. If injection was provisioned but the material is gone, checkout fails closed before we dial the target.

Generic RDP and RDCleanPath now share one CredSSP session path. Kerberos uses the token `dst_hst` for the client-facing TERMSRV SPN.

I wrote the lifecycle contract down in [src/credential/INTENT.md](https://github.com/Devolutions/devolutions-gateway/blob/85545af64a3f911d0a7e3d3d68ecb8a105a1121f/devolutions-gateway/src/credential/INTENT.md) (following [intent-files.md](https://github.com/Devolutions/devolutions-gateway/blob/master/docs/intent-files.md)) — that's what this PR implements.

### Why this design

Credentials live in two phases. Before the first connect, the preflight `time_to_live` bounds how long staged secrets wait in memory. After the first checkout, retention switches to the token acceptance deadline (`exp` + leeway). If Gateway accepts the token for a reconnect, the mapping has to still be there — the old one-shot `take_mapping` broke exactly that. `jet_reuse` starts at disconnect and the TTL starts at preflight, so neither can cap the other; the token deadline is the one clock every connection attempt shares.

### Where to look hardest

- [Generation-guarded KDC publication](https://github.com/Devolutions/devolutions-gateway/blob/85545af64a3f911d0a7e3d3d68ecb8a105a1121f/devolutions-gateway/src/credential_injection.rs#L667-L698): re-provisioning the same JTI bumps a generation, and an older in-flight lease can never overwrite a newer published KDC, so the newest provisioning wins deterministically.
- [Deadline-driven cleanup](https://github.com/Devolutions/devolutions-gateway/blob/85545af64a3f911d0a7e3d3d68ecb8a105a1121f/devolutions-gateway/src/credential_injection.rs#L774-L835): expired session KDC material is removed by a background task, not lazily on the next checkout, and secrets are zeroized on drop.
- [Release after CredSSP](https://github.com/Devolutions/devolutions-gateway/commit/85545af64a3f911d0a7e3d3d68ecb8a105a1121f): the per-connection mapping and KDC lease drop as soon as both CredSSP legs finish, so plain forwarding never holds secret material.

Tests cover staging expiry, reconnect reuse, generation replacement, and physical removal without a later checkout. Manual run: 22-case DVLS matrix with live NTLM and Kerberos CredSSP through Gateway. E2E hops are stacked on #1938.

Issue: DVLS-14697


devolutions-gateway/src/generic_client.rs
```diff
@@ -152,14 +155,18 @@ where
                 // The credential store is keyed on the association token's JTI, so a direct
                 // lookup by `claims.jti` is the primary path.
                 if is_rdp
-                    && let Some(entry) = credentials.get(claims.jti)
+                    && let Some(entry) = provisioning.take(claims.jti)
```
irvingoujAtDevolution: Create credential injection from provisioning.
The analogy, as well as the mindset I'd like to use here, is that the provisioning store is like a supermarket, where each handler — for example, GenericClient here — can purchase what he wants with a `jti` and start preparing the meal (credential injection here).

This grants better separation of concerns, as well as better extensibility.

devolutions-gateway/src/generic_client.rs
```diff
@@ -152,14 +155,18 @@ where
...
                     && entry.mapping.is_some()
                 {
                     anyhow::ensure!(token == entry.token, "token mismatch");
-                    let credential_injection_kdc = credentials.kdc_for(claims.jti)?;
+                    let kerberos_enabled = conf.debug.enable_unstable && conf.debug.kerberos_credential_injection;
+                    let credential_injection =
+                        CredentialInjection::from_provisioned(claims.jti, entry, kerberos_enabled)?
+                            .register_if_kerberos(&synthetic_kdc_registry);
```
irvingoujAtDevolution: Now the Chef 1, the generic client has done the first round of preparation, we pass it to the second chef, which is the kdc proxy hereOnce Chef 1 (GenericClient) finishes the first round of prep, the dish gets handed off to the second chef — the KDC proxy — to continue the process.

devolutions-gateway/src/api/kdc_proxy.rs
```diff
@@ -47,7 +46,9 @@ async fn kdc_proxy(
         KdcDestination::Inject { jti } => {
             enforce_credential_injection_enabled(jti, conf.debug.enable_unstable)?;
 
-            let kdc = credentials.kdc_for(jti).map_err(credential_injection_resolve_error)?;
+            let kdc = synthetic_kdc_registry
```
irvingoujAtDevolution, line 49: Now everything the KDC Proxy need is in the synthetic_kdc_registry, there's no need for KDC handler to read provisioning store anymore. 
More semantically clear

devolutions-gateway/src/generic_client.rs
irvingoujAtDevolution: Notice it's a `take` here, so provisioning won't be reused, one of the benifits this refactoring brings

devolutions-gateway/src/credential_injection.rs
```diff
@@ -0,0 +1,794 @@
...
+            .as_ref()
+            .and_then(|o| o.krb_kdc())
+            .cloned()
+            .ok_or_else(|| {
+                warn!(%jti, "Kerberos credential injection requires krb_kdc");
+                CredentialInjectionKdcResolveError::MissingKrbKdc { jti }
+            })?;
+
+        let proxy_username = app_credential_username(&mapping.proxy).to_owned();
+        let synthetic = CredentialInjectionKdc::new(jti, target_hostname, &proxy_username, &mapping.proxy)
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Not applying. Lab + product path (RDM/DVLS Host = target machine) requests `TERMSRV/<dst_hst>`. Using `conf.hostname` broke injection with SPN mismatch (client ticket `IT-HELP-DC...` vs acceptor `devolutions432`). Synthetic KDC SPN stays association-token `dst_hst`; #1856's gateway-hostname assumption does not match this client path.

devolutions-gateway/src/rd_clean_path.rs
```diff
@@ -541,13 +534,16 @@ pub async fn handle(
     // proxy-based credential injection mode. Otherwise, we continue the usual
     // clean path procedure. The credential store is keyed on the association token's JTI.
     if let Some(jti) = crate::token::extract_jti(token).ok()
-        && let Some(entry) = credentials.get(jti)
+        && let Some(entry) = provisioning.take(jti)
         && entry.mapping.is_some()
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Fixed: CleanPath now peeks with `has_mapping`, authorizes the token, then `take`s one-shot groceries.

devolutions-gateway/src/credential_injection.rs
```diff
@@ -0,0 +1,794 @@
...
+#[derive(Debug, Clone)]
+pub struct SyntheticKdcRegistry {
+    inner: Arc<Mutex<RegistryInner>>,
+}
+
+#[derive(Debug, Default)]
+struct RegistryInner {
+    live: HashMap<Uuid, PublishedSyntheticKdc>,
+    next_generation: HashMap<Uuid, u64>,
+}
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Fixed: registry uses a single monotonic `next_generation: u64` instead of a per-JTI map.

devolutions-gateway/src/rdp_proxy/credssp.rs
```diff
@@ -0,0 +1,473 @@
...
+
+        buf.clear();
+        result.to_buffer(&mut buf).context("write early user auth result")?;
+        let response = &buf[..result.buffer_len()];
+        framed.write_all(response).await.context("write_all")?;
+    }
+
+    return result;
+
+    #[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
irvingoujAtDevolution: Fixed: replaced `allow` with justified `#[expect(clippy::too_many_arguments, reason = ...)]`.

CBenoit (changes requested):
> [!WARNING]
> **Automated review. Human insight is required**, especially to resolve the conflicting SPN assumptions described below.

Requesting changes on this stacked delta (`chore/regenerate-openapi-provisioning` -> `refactor/rdp-proxy-credssp-extract`).

1. **Blocking - the existing SPN regression thread remains unresolved.** `credential_injection.rs:222,260,351,416` now derives the synthetic-KDC service hostname from the association token's `dst_hst`. The base branch deliberately used `conf.hostname` after merged fix #1856, and this PR deletes `service_kdc_for_uses_gateway_hostname_for_spn`, the regression test that pinned that behavior. The author has supplied contrary product-path evidence in the existing thread, so this needs a human protocol decision rather than another automated guess. Whichever hostname is correct must be explicit, tested with different Gateway and destination hostnames, and described as a behavior change if #1856 is intentionally being reversed.

2. **The new one-shot store changes observable retry behavior without documenting it.** `ProvisioningStore::take` (`provisioning.rs:143-197`) removes credentials and connection options before target connection, TLS, or CredSSP can succeed (`generic_client.rs:158-197`, `rd_clean_path.rs:490-580`). A transient failure consumes the data, and a retry inside the advertised TTL can silently fall through to non-injected forwarding. This conflicts with the current `time_to_live` wording in `openapi.rs:396-400` ("Minimum persistence duration") and is not a behavior-preserving `refactor:`. Either retain entries across failed attempts, or document the single-use contract in OpenAPI/PR history and make skipped injection explicit instead of silently falling back.

3. **The CredSSP enclosure is incomplete and leaves two orchestration implementations.** `credssp.rs:73-153` owns both CredSSP legs, Connect Confirm interception, leftover flushing, and forwarding, but `rd_clean_path.rs:551-634` open-codes the same sequence. `rdp_proxy/mod.rs:12-14` then re-exports three internals solely to sustain that duplicate path. Build `CredsspSession`/`PreparedCredssp` from RDCleanPath too and call `run`; that removes roughly 80 duplicated lines and makes the stated module boundary real.

4. **`generic_client` destructively reads entries it does not use.** At `generic_client.rs:157-160`, `take` runs before `entry.mapping.is_some()`, so an RDP connection consumes a token-only `provision-token` entry and discards it. Match the safer RDCleanPath shape: check `has_mapping` first, then `take` only for the injection path.

5. **Protocol gating was duplicated while its regression coverage was deleted.** The two-flag condition is repeated at `generic_client.rs:162` and `rd_clean_path.rs:495`, while the prior flag-matrix test was removed. The empty test module at `rdp_proxy/mod.rs:274-278` incorrectly claims protocol selection remains covered. Centralize the predicate, restore the opt-in matrix test, and remove the empty scaffolding.

6. **The kitchen metaphor makes the new boundaries harder to learn.** `credential_injection.rs:1-5,101,108,205,524-531` and `rd_clean_path.rs:668-669` require readers to translate "supermarket/groceries/chef/dish/pass window" back into `ProvisioningStore`, `ProvisioningEntry`, preparation, and live KDC registration. Use those technical terms directly; this module's purpose is separation of concerns, so it should not introduce a second vocabulary for them.

Non-blocking cleanup: `CredentialInjectionKdcResolveError` no longer has a discriminating consumer (`credential_injection.rs:55-72`), so its four variants and duplicated pre-return warnings should either regain a matching caller or collapse to contextual `anyhow` errors. Also capitalize the newly added log messages and terminate sentence comments per repository conventions.

> [!NOTE]
> LLM-assisted content (no human feedback).

irvingoujAtDevolution:
Implementation notes:

- `ProvisioningStore` owns pending token, credential, and connection-option material.
- `SyntheticKdcRegistry` contains only KDCs published by active Kerberos injection sessions.
- Synthetic KDC ticket/SPN handling and target-side Kerberos both use association `dst_hst`; `conf.hostname` remains Gateway identity for TLS/configuration uses.
- RDCleanPath now calls `CredsspSession::run` instead of maintaining a second CredSSP orchestration.
- Generic RDP checks `has_mapping` before one-shot checkout, preserving token-only provisioning rows.
- Added explicit Kerberos/NTLM selection coverage and dynamic-destination tests.
- Local DVLS matrix: 22 Playwright cases passed, including live NTLM and Kerberos CredSSP through Gateway.
- CI packaging fix retries Chocolatey and verifies `candle.exe`/`WIXSHARP_WIXDIR` after transient 503 failures.

> [!NOTE]

irvingoujAtDevolution:
Final review follow-up:

1. **SPN decision:** `dst_hst` is the explicit service-name contract for supported credential-injection clients. Synthetic KDC tickets, CredSSP acceptor properties, and the target Kerberos leg use the same dynamic destination. Gateway-addressed clients that derive `TERMSRV/<conf.hostname>` from the transport endpoint are explicitly outside this unstable feature contract.
2. **One-shot behavior:** checkout is atomic; a consumed-JTI tombstone remains until the original expiry, so reuse and concurrent losers fail closed instead of falling through to non-injected forwarding. Re-provisioning clears the tombstone.
3. **CredSSP enclosure:** RDCleanPath constructs `CredsspSession` / `PreparedCredssp` and calls the single `run` implementation.
4. **Token-only provisioning:** both entry points inspect `MappingStatus` before checkout; token-only rows are preserved.
5. **Protocol gating:** both flags are centralized and covered by the full 2x2 matrix, with target username selection tested separately.
6. **Terminology:** kitchen/supermarket metaphors are removed from source and docs.

Additional cleanup: removed the unused typed KDC resolve error, centralized checkout validation, kept generation-guarded KDC leases for explicit same-JTI re-provision overlap, and regenerated OpenAPI artifacts.

Validation: full workspace Clippy; workspace lib/bin/test targets plus colliding examples run separately; generated .NET client build; 22-case DVLS matrix, including live NTLM and Kerberos CredSSP over RDCleanPath.

> [!NOTE]

devolutions-gateway/src/generic_client.rs
irvingoujAtDevolution: Updated in current HEAD `29248b24`: `generic_client` now peeks `MappingStatus` and calls `CredentialInjection::checkout`; `ProvisioningStore::take_mapping` owns atomic token validation and consumption. The supermarket terminology was removed in favor of the actual technical types.

> [!NOTE]

devolutions-gateway/src/generic_client.rs
irvingoujAtDevolution: Updated in current HEAD `29248b24`: there is no chef-style handoff. Kerberos checkout publishes a session-scoped KDC in `SyntheticKdcRegistry`, and `/jet/KdcProxy` resolves only that active registry entry (`api/kdc_proxy.rs:49-51`).

> [!NOTE]

devolutions-gateway/src/generic_client.rs
irvingoujAtDevolution: Updated in current HEAD `29248b24`: one-shot checkout is explicit and fail-closed. `take_mapping` consumes atomically and records a tombstone; a reused JTI returns `Consumed` until expiry or re-provisioning instead of silently falling back.

> [!NOTE]

irvingoujAtDevolution:
@CBenoit, ready for review

> [!NOTE]

CBenoit:
(Reviewed and feedback sent on Slack)

devolutions-gateway/src/api/kdc_proxy.rs
irvingoujAtDevolution: Still accurate on current HEAD `eb4587a4`: `/jet/KdcProxy` looks up `synthetic_kdc_registry` only and never reads `ProvisioningStore`.

> [!NOTE]
> LLM-assisted content (no human feedback).

irvingoujAtDevolution:
@CBenoit, ready for review.

Credential injection now keeps the mapping until the JWT can still authenticate, so `jet_reuse` reconnects inject again instead of consuming on first checkout. Required-but-missing material fails closed. Manual RDM/DVLS Kerberos and NTLM launches against IT-HELP-DC succeeded on this HEAD.

> [!NOTE]
> LLM-assisted content (no human feedback).

irvingoujAtDevolution:
@CBenoit ready for review.

Credential injection now keeps mappings across token-accepted reconnects (get_mapping, fail-closed if required material is missing). Stacked E2E is #1938.

> [!NOTE]
> LLM-assisted content (no human feedback).

mamoreau-devolutions (approved):

irvingoujAtDevolution:
The clocks are separate on purpose. We stopped trying to make one duration cover all of them.

```text
preflight time_to_live  = unused mapping wait (staging)
jet_reuse / RDP 10s     = can this JWT be replayed
exp + 5 min             = hard cap after first successful inject
jet_ttl                 = already-running session (unchanged)
```

**Token reuse is only the token cache.** `jet_reuse` is the optional reconnect window after disconnect. RDP still has the built-in ~10s handshake allowance even when `jet_reuse` is off (mstsc opens several sockets):

```rust
match (disconnected_info, jet_reuse) {
    (Some(info), ReconnectionPolicy::Allowed { window_in_seconds }) => {
        // not killed, inside window since disconnect, reconnect count <= 10
    }
    _ => {
        if matches!(jet_ap, ApplicationProtocol::Known(Protocol::Rdp)) {
            // ~10 seconds since last token use, not disconnect
            if now > bucket.get().last_use_timestamp + RDP_MAX_REUSE_INTERVAL_SECS {
                return Err(TokenError::UnexpectedReplay { ... });
            }
        }
    }
}
```

at [token.rs](https://github.com/Devolutions/devolutions-gateway/blob/e8261332f78e9937a60838c2e616adccf71e52aa/devolutions-gateway/src/token.rs#L1124-L1157)

That layer never looks at the credential store. If reuse is refused, we never reach injection.

**Credentials have two phases.** Staging uses preflight `time_to_live` (default 15 min, max 2 h), capped to the token acceptance deadline (`exp + 5 min leeway`). After the first successful checkout, retention is that deadline — not `jet_reuse`, and not the original staging TTL:

```rust
let staging_expires = now + time_to_live;
let required_until = if mapping.is_some() {
    Some(crate::token::token_acceptance_deadline(exp))
} else {
    None
};
let expires_at = required_until.map_or(staging_expires, |deadline| staging_expires.min(deadline));
```

at [provisioning.rs](https://github.com/Devolutions/devolutions-gateway/blob/e8261332f78e9937a60838c2e616adccf71e52aa/devolutions-gateway/src/provisioning.rs#L118-L127)

```rust
pub(crate) fn token_acceptance_deadline(exp: i64) -> time::OffsetDateTime {
    let timestamp = exp.saturating_add(i64::from(LEEWAY_SECS));
    // ...
}
```

at [token.rs](https://github.com/Devolutions/devolutions-gateway/blob/e8261332f78e9937a60838c2e616adccf71e52aa/devolutions-gateway/src/token.rs#L1294-L1296)

```rust
// First successful lookup keeps the mapping until the token deadline.
if now >= deadline || now >= entry.expires_at {
    anyhow::bail!("credential-injection material for {jti} is missing or expired; re-provision to retry");
}
entry.expires_at = deadline;
```

at [provisioning.rs](https://github.com/Devolutions/devolutions-gateway/blob/e8261332f78e9937a60838c2e616adccf71e52aa/devolutions-gateway/src/provisioning.rs#L260-L264)

So: unused secrets die with staging TTL. Once injected, “Gateway accepted this association JWT” implies the mapping is still there. We do not compare `time_to_live` to `jet_reuse` — they start at different events.

**If the token is still acceptable and the mapping is gone, we fail closed** (no ordinary forward, no upstream dial):

```rust
MappingStatus::RequiredMissing => anyhow::bail!(...),
MappingStatus::Available => true,
MappingStatus::Absent => false,
// checkout before connect_upstream
```

at [generic_client.rs](https://github.com/Devolutions/devolutions-gateway/blob/e8261332f78e9937a60838c2e616adccf71e52aa/devolutions-gateway/src/generic_client.rs#L125-L170)

```rust
if now >= deadline {
    MappingStatus::Absent
} else if entry.mapping.is_some() && now < entry.expires_at {
    MappingStatus::Available
} else {
    MappingStatus::RequiredMissing
}
```

at [provisioning.rs](https://github.com/Devolutions/devolutions-gateway/blob/e8261332f78e9937a60838c2e616adccf71e52aa/devolutions-gateway/src/provisioning.rs#L196-L201)

**Kerberos KDC follows the same activated deadline**, interned per JTI + provisioning generation so reconnects keep the same synthetic realm:

```rust
/// Reuse the synthetic KDC for this provisioning generation until `expires_at`.
pub(crate) fn intern_session_kdc(...) -> Arc<CredentialInjectionKdc> {
    if existing.provision_generation == provision_generation {
        return Arc::clone(&existing.kdc);
    }
}
```

at [credential_injection.rs](https://github.com/Devolutions/devolutions-gateway/blob/e8261332f78e9937a60838c2e616adccf71e52aa/devolutions-gateway/src/credential_injection.rs#L666-L692)

We kept the preflight TTL field. It only covers the wait before first connect. After that, `exp` is the cap. A long `jet_reuse` still needs a token `exp` that covers it; Gateway will not keep credentials past the token.

Stacked hop tests are on #1938. Signed `jet_reuse` token-cache is not covered there (`disable_token_validation` + unsigned JWTs).

> [!NOTE]

irvingoujAtDevolution:
lifetime-timeline

irvingoujAtDevolution:
### Summary of Timing & Lifetime Parameters

| Attribute | Starts At | Deadline / Rule | Core Role |
| :--- | :--- | :--- | :--- |
| **`preflight.time_to_live`** (`preflight.ttl`) | Preflight API call | `min(now + ttl, association.exp + 5m)` *(promoted on first connect)* | **Staging wait time before initial connect.**<br>How long provisioned credentials stay in memory if no client connects. |
| **`association.exp`** | Fixed wall-clock timestamp (issued by DVLS) | `association.exp + 5m` (with leeway) | **Global absolute ceiling for token & credential lifetime.**<br>Hard deadline for credential retention in memory after first checkout; requests past this are rejected and credentials evicted. |
| **`association.jet_reuse`** | Client disconnection timestamp (`T_disc`) | `T_disc + jet_reuse` *(and reconnect count <= 10)* | **Relative sliding window for fast reconnects after disconnect.**<br>Mitigates token replay; restricts reconnects to a bounded window post-disconnect. |

---

### Key Takeaway & Interaction

- **`preflight.time_to_live`** governs how long credentials wait in staging before the first connection arrives.
- **`association.jet_reuse`** governs the allowed reconnect window once disconnected.
- **`association.exp`** is the hard absolute ceiling that cannot be exceeded by either staging or reconnect windows.

> [!NOTE]

CBenoit (approved):
Thank you for the exhaustive write-ups, LGTM!

PR #1905, opened 2026-08-04
feat(agent): support canceling package broker operations

vnikonov-devolutions:
Adds a `POST /v1/package-operations/cancel` endpoint to the Devolutions Agent package broker. Clients can now cancel an in-flight package operation: the broker terminates the spawned package-manager process and reports the operation as `Canceled`. Cancelation is idempotent, and canceling an already-finished operation simply returns its final status. Unknown operations, or operations owned by another client, are reported as not found.

This enables UniGetUI (and other broker clients) to give users a working cancel button for installs, updates, and uninstalls that run through the broker.

Depends on now-policy-api / now-policy-server-template 0.3 (cancel API, `Canceling`/`Canceled` statuses, API version 1.1).

> [!NOTE]
> Draft: the workspace `Cargo.toml` temporarily patches `now-policy-api` and `now-policy-server-template` to a local path. These patches must be removed once the 0.3 crates are published to crates.io.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1524,7 +1524,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ccc2776f0c61eca1ca32528f85548abd1a4be8fb53d1b21c013e4f18da1e7090"
 dependencies = [
  "data-encoding",
- "syn 2.0.118",
+ "syn 1.0.109",
 ]
 
 [[package]]
@@ -2593,8 +2593,8 @@ dependencies = [
  "libc",
  "log",
  "rustversion",
- "windows-link 0.2.1",
- "windows-result 0.4.1",
+ "windows-link 0.1.3",
+ "windows-result 0.3.4",
 ]
 
 [[package]]
@@ -3120,7 +3120,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.62.2",
+ "windows-core 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


vnikonov-devolutions:
Implementation notes:

- **New endpoint**: `POST /v1/package-operations/cancel` (`CancelRequest`/`CancelResponse` from now-policy-api 0.3). Auth mirrors the status endpoint: client context + Devolutions Authenticode signature validation, plus an owner-key check (`effective_user|client_executable_path`), so a client can only cancel its own operations (`NotFound` otherwise).
- **Cancel semantics**: async + idempotent. A non-terminal operation transitions to `Canceling` and its per-operation `CancellationToken` (stored in `OperationTracker`) is triggered; the response reports `Canceling`. Terminal operations return their final status unchanged. Repeated cancels are harmless.
- **Process termination**: the Windows executor's `create_process` wait was changed from a single 1-hour blocking wait to a 500 ms polling loop that checks the cancel token and the overall deadline. On cancel it terminates the process (exit code 1) and returns a typed `ExecutionCanceled` error; the background execution task downcasts it and marks the operation `Canceled` instead of `Failed`.
- **Token plumbing**: the cancel token only guards the pre-command and the main command; kill-before-operation and post-operation commands get a fresh token (post is best-effort cleanup and is skipped anyway when the main command errors out).
- **Race handling**: `mark_running` refuses to regress `Canceling`/`Canceled`; `mark_completed`/`mark_failed` never overwrite `Canceled`. If a process exits normally before the terminate lands, its real outcome wins.
- **Testing**: unit tests cover unknown-operation, running-operation (token fired + `Canceling`), terminal-operation (status preserved), and owner-mismatch paths, plus tracker state-transition guards. `cargo +nightly fmt`, `cargo clippy --tests -D warnings`, and `cargo test -p devolutions-agent` all pass.
- **Temporary**: `[patch.crates-io]` path entries for `now-policy-api`/`now-policy-server-template` point at a local now-proto worktree until 0.3 is published; remove before merging (marked with a TODO).

vnikonov-devolutions:
superseded by #1913

PR #1908, opened 2026-08-04
fix(agent): allow DVC open retries to complete before negotiation times out

thenextman:
When an RDP client is slow to accept the Devolutions Agent's virtual channel; for example when a third-party multi-factor authentication provider delays sign-in, or when reconnecting to an existing session; the agent could give up on the session before the connection had a chance to establish. The in-session agent then stayed unavailable for the rest of that RDP session, and features that depend on it, such as jumping through a host from Remote Desktop Manager, reported the agent as missing.

The agent now gives the channel its full retry budget instead of abandoning it partway through, so these connections succeed where they previously did not.

When the channel genuinely cannot be established, the agent now reports it as an error rather than shutting down as though nothing were wrong, and the logs identify which stage failed. Virtual channel failures are also logged with their numeric error code and timing, so a support log makes the cause clear regardless of the system's display language.

### DIFF
--- a/devolutions-session/Cargo.toml
+++ b/devolutions-session/Cargo.toml
@@ -45,6 +45,10 @@ optional = true
 version = "0.4.3"
 features = ["std"]
 
+[dev-dependencies]
+# `rt` and `test-util` are needed to drive the async DVC initialization tests with paused time.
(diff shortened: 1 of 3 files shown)


vnikonov-devolutions (approved):
LGTM!

PR #1931, opened 2026-08-12
fix(agent): private key permissions on Windows

ptrstr:
Private keys for the agent tunnel are now created with an explicit access control list limited to SYSTEM and Administrators, instead of inheriting the permissions of the Devolutions data directory. Previously these files were readable by any local user on the machine: the tunnel CA and server keys on a Gateway host, and the tunnel client key on an enrolled Agent.

Existing keys keep their current permissions. I'm not sure how widely deployed/used this feature was, so let me know if a migration is required.


irvingoujAtDevolution (approved):
LGTM

PR #1935, opened 2026-08-14
feat(video-streamer): normalize recording sessions

irvingoujAtDevolution:
Turns one recording session into one pull-driven stream of independent VP8 WebM segments.

It accepts append-only VP8 or VP9 clips and starts a new fixed-size segment when the source clip or resolution changes. Gateway wiring and player changes stay out of this PR.

Depends on #1934. The Cadeau image-size API is already merged in Devolutions/cadeau#84.

Checks:
- `cargo +nightly fmt --all -- --check`
- `cargo check -p video-streamer`
- `cargo clippy -p video-streamer --lib -- -D warnings`
- Tests were not run in this session.

Issue: N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -929,8 +929,7 @@ dependencies = [
 [[package]]
 name = "cadeau"
 version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


irvingoujAtDevolution:
Superseded by #1942.

PR #1955, opened 2026-08-26
perf(dgw): avoid cloning decoded frames

CBenoit:
Re-encoding cloned each decoded video frame after extracting it. Move the sole frame out of the collection to avoid that allocation and copy.


CBenoit:
Implementation notes:
- Moves the sole decoded frame out of the temporary vector after the non-lacing invariant is checked.
- Preserves frame bytes and parsing behavior while removing one full-frame allocation and copy per re-encoded frame.
- Adds SimpleBlock extraction coverage.

Validation:
- cargo +nightly fmt --all
- cargo clippy --workspace --tests -- -D warnings
- cargo test -p video-streamer

> [!NOTE]

irvingoujAtDevolution (approved):
LGTM

PR #1958, opened 2026-08-27
feat(dgw): persist agent tunnel authorization

irvingoujAtDevolution:
Persists accepted Agent identities and authorized public keys so deleting an Agent remains effective after Gateway restarts.

Gateway now admits tunnel connections only when the certificate Agent ID and SPKI match stored authorization. Deleting an Agent removes that authorization before closing its live tunnel connection, and management endpoints continue to list accepted Agents while offline.

Existing pre-stable Agents must be re-enrolled after upgrading.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


crates/agent-tunnel/src/cert.rs
```diff
@@ -231,8 +248,6 @@ impl CaManager {
             .signed_by(&csr_params.public_key, &ca_cert, &self.ca_key_pair)
             .context("sign agent certificate with CA")?;
 
-        info!(%agent_id, %agent_name, "Signed agent CSR and issued client certificate");
-
```
CBenoit, line 235: Is it intentional?

irvingoujAtDevolution: recovered

CBenoit (approved):
LGTM

PR #1964, opened 2026-08-30
fix(webapp): fix type error in VNC form when the app run in non-secure context

RRRadicalEdward:
I was able to hit this error when running the WebApp in a non-secure context.
image
This PR fixes it.


CBenoit (approved):
LGTM, thank you!

PR #1967, opened 2026-09-02
chore(openapi): bump the .NET and TypeScript client versions

irvingoujAtDevolution:
Bumps the gateway OpenAPI clients so the API changes from #1947 (agent tunnel out of debug) can be published:

- `Devolutions.Gateway.Client` 2026.2.4 -> 2026.9.2
- `@devolutions/gateway-client` 0.16.0 -> 0.17.0

Only version strings change. I re-ran `tools/generate-openapi/generate.ps1` and `openapi/generate_clients.ps1` (generator 7.9.0); the spec and every other generated file are identical to master.


CBenoit (approved):
LGTM!

PR #1970, opened 2026-09-02
ci: build npm libraries without the full webapp

irvingoujAtDevolution:
The npm library jobs installed the full webapp workspace, which pulled in the private `@devolutions/icons` package even though the published libraries do not use it. This made releases depend on a warm pnpm cache.

Install only each library and its workspace dependencies, remove the shared pnpm cache configuration, and stop immediately when installation fails.


CBenoit (approved):
LGTM!

Issue #1983, opened 2026-09-10
feat(agent): add Swedish Desktop Agent and installer localization


yeager:
Add Swedish (sv) satellite resources for the Desktop Agent and its Windows installer.

The agent already ships English neutral resources and French localizations. Swedish resources localize visible profile, service-status, privilege-elevation, error and installation/configuration messages for Swedish-speaking administrators.

Scope:
- `dotnet/DesktopAgent/Properties/Resources.sv.resx`
- `package/AgentWindowsManaged/Resources/DevolutionsAgent_sv-se.wxl`
- installer resource registration in `DevolutionsAgent.csproj`.

PR #1984, opened 2026-09-10
feat(agent): add Swedish Desktop Agent localization

yeager:
Adds Swedish text for the visible Desktop Agent profile, service-status, privilege-elevation and error messages.

This gives Swedish-speaking administrators a localized agent interface while retaining the existing resource keys and behavior.

Issue: #1983

### DIFF
--- a/dotnet/DesktopAgent/Properties/Resources.sv.resx
+++ b/dotnet/DesktopAgent/Properties/Resources.sv.resx
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <resheader name="resmimetype">
+    <value>text/microsoft-resx</value>
+  </resheader>
+  <resheader name="version">
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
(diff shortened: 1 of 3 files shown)


yeager:
Implementation notes:
- Adds `Resources.sv.resx` as a standard .NET satellite resource, following the existing French resource pattern.
- Contains the same 14 visible Desktop Agent keys as the neutral resource; no source strings, generated code, or runtime behavior change.
- XML syntax was validated with the available parser, key parity found 14/14 keys with no extras or empty values, and `git diff --check` passes.
- The local environment does not include the .NET SDK, so resource compilation is intentionally left to the repository CI.
- No breaking changes or migration steps.

yeager:
Installer localization update:
- Adds a Swedish `sv-se` WiX localization resource for the Agent installer and registers it as an embedded installer resource.
- The resource has 72/72 current English installer keys, including Agent Tunnel DNS/CIDR enrollment, PowerShell Universal endpoint/token configuration, UAC, installation, repair and removal flows.
- XML parsing, exact key parity, MSI variable/format-token parity, and `git diff --check` pass locally.
- The .NET/Wix build remains delegated to repository CI because the local environment has no .NET SDK.

PR #1995, opened 2026-09-18
fix(agent): start the session process only for remote sessions

thenextman:
Devolutions Agent starts a session host process inside user sessions to serve the Devolutions Agent virtual channel. Only a remote session can serve that channel, because it needs an RDP client on the other end to open it against. The agent recorded whether each session was a console or a remote session, and the code stated in a comment that the process is only started for remote ones.

It was never actually restricted that way. The decision to start the process considered only whether the session was ready - registered, and past the logon or unlock event it was waiting for - and never consulted the console-or-remote value it had recorded. Console sessions are registered when they connect and Windows reports a logon for them like any other session, so they satisfied every condition the check applied. Signing in at a machine's console therefore started a session host process there, which found no RDP client to open the channel against, failed immediately, and exited. It repeated on every local sign-in.

The session kind is now part of the decision, so the behaviour matches what was intended: console sessions are still tracked, so that a later upgrade to a remote session is picked up, but the process is only ever started for a remote session.

This also removes a class of misleading support logs. The session host writes its log per user rather than per session, so these console-session failures were indistinguishable from a genuine RDP connection failing to establish the
channel. To make the remaining failures readable, the session host now records the session it was started in.


CBenoit (approved):
LGTM

PR #1996, opened 2026-09-22
fix(dgw): identify subscriber requests

CBenoit:
Send a versioned Devolutions Gateway User-Agent header with every subscriber notification so endpoints can identify the caller.


CBenoit:
Implementation notes:

- Adds `User-Agent` directly to each subscriber request so the shared proxy-client cache cannot discard it.
- Uses `Devolutions-Gateway/<version>`, sourced from `CARGO_PKG_VERSION`.
- Adds a wire-level regression test that pre-populates the cache without default headers and bounds the network exchange with a timeout.
- Validated with workspace-wide formatting, Clippy, and tests.

> [!NOTE]

